PR #2, opened 2016-01-14
Use mathematical 'For All' character instead of 'Latin Capital Letter…

cogwheel:
… Turned A'

∀ has wider support than Ɐ, and is semantically correct for its use in jank.

Breaks all code using Ɐ


jeaye:
Thanks! The updated tests are much appreciated.

PR #43, opened 2023-07-31
Correction in DESIGN.md

DanteFalzone0:
Original said "damn", which means "to send to hell". I think you meant "dam". :)


jeaye:
technically correct

PR #54, opened 2024-01-16
feat(cli): allows to run path.to.ns/-main on classpath

Samy-33:


src/cpp/main.cpp
```diff
@@ -55,6 +55,37 @@ namespace jank
...
+    {
+      profile::timer timer{ "require clojure.core" };
+      rt_ctx.load_module("/clojure.core").expect_ok();
+    }
+
+    {
+      profile::timer timer{ "eval user code" };
+      rt_ctx.load_module("/" + opts.target_module).expect_ok();
+
+      auto main_var = rt_ctx.find_var(opts.target_module, "-main").unwrap_or(nullptr);
```
jeaye, line 69: I changed this to be:

```cpp
auto const main_var(rt_ctx.find_var(opts.target_module, "-main").unwrap_or(nullptr));
```

1. We always use direct initialization
2. Whatever can be const must be

src/cpp/main.cpp
```diff
@@ -55,6 +55,37 @@ namespace jank
...
+      if(main_var)
+      {
+        // TODO: handle case when `-main` accept no arg.
+        runtime::detail::native_transient_vector extra_args;
+        for(auto const &s : opts.extra_opts)
+        {
+          extra_args.push_back(make_box<runtime::obj::string>(s));
+        }
+        jank::runtime::apply_to(main_var->deref(),
```
jeaye, line 78: I removed the `jank::` prefix here, which isn't needed.

jeaye:
Excellent work! Thank you!

PR #60, opened 2024-03-08
feat(transient-vector): adds initial support for transient vector

Samy-33:


Samy-33:
I noticed that we don't have `get` declared in `transient_vector` and `call` in `persistent_vector`. Any reason for that?

In clojure
```clojure
([1 2 3] 5)
```
throws and `IndexOutOfBoundException`, we should have the same behaviour or we can prefer nil punning here? Any reason why clojure can't just return `nil` instead of throwing exception?
I prefer nil punning personally, but is it valuable enough to divert from clojure's behaviour?

Also, `vector` is not a 2-arity fn in clojure, but we have declared it in jank. It makes sense to me to have it, but again diverting from clojure's behaviour.

Let me know your thoughts on this, @jeaye

jeaye:
> I noticed that we don't have a `call` in `persistent_vector`. Any reason for that?
> 
> In clojure
> 
> ```clojure
> ([1 2 3] 5)
> ```
> 
> throws and `IndexOutOfBoundException`, we should have the same behaviour or we can prefer nil punning here? Any reason why clojure can't just return `nil` instead of throwing exception? I prefer nil punning personally, but is it valuable enough to divert from clojure's behaviour?
> 
> Also, `vector` is not a 2-arity fn in clojure, but we have declared it in jank. It makes sense to me to have it, but again diverting from clojure's behaviour.
> 
> Let me know your thoughts on this, @jeaye

Yeah, we just haven't implemented that behavior on `persistent_vector` yet. However, the way that it's done is a little tricky, since we actually are hard-coding this right now, in `dynamic_call`. It's a huge performance win to do this, compared to implementing `callable` for those types, since that will result in vtable ptrs and virtual dispatch. If you'd like to add support for `persistent_vector` in there, take a look at `callable.cpp:70`, as well as `evaluate.cpp:260`.

As a rule, if Clojure throws, jank throws. We're in the business of hitting parity, not changing semantics (even when we don't necessarily like those semantics).

include/cpp/jank/runtime/obj/persistent_vector.hpp
```diff
@@ -5,9 +5,19 @@
 
 namespace jank::runtime
 {
+  namespace obj
+  {
+    using persistent_vector = static_object<object_type::persistent_vector>;
```
jeaye, line 10: This will be a problem, since we're instantiating `static_object<object_type::persistent_vector>` before it's defined. It's defined in this file, starting at line 17. That's why this was provided after. I know you need the transient version before, so let's just have both.

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  native_box<obj::transient_vector::persistent_type> obj::transient_vector::to_persistent()
+  {
+    assert_active();
+    active = false;
+    return make_box<obj::persistent_vector>(std::move(data));
+  }
+
+  object_ptr obj::transient_vector::call(object_ptr const o) const
```
jeaye, line 71: You'll need to add in support for this in `callable.cpp:70` and `evaluate.cpp:260`.

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,118 @@
...
+    assert_active();
+    if(o->type == object_type::integer)
+    {
+      auto const i(expect_object<obj::integer>(o)->data);
+      if(i < 0 || data.size() <= static_cast<size_t>(i))
+      {
+        return obj::nil::nil_const();
+      }
+
+      return data.at(i);
```
jeaye, line 82: In C++, the general practice is that `foo.at(i)` performs index checking and throws, whereas `foo[i]` is direct memory access. For immer's transients, this holds true. So this will be more efficient to use `data[i]` here instead, since we've already done the bounds checking.

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,118 @@
+#include "jank/runtime/detail/object_util.hpp"
+#include "jank/runtime/erasure.hpp"
+#include "jank/type.hpp"
+#include <jank/runtime/util.hpp>
+#include <jank/runtime/obj/native_function_wrapper.hpp>
+#include <jank/runtime/obj/transient_vector.hpp>
+#include <stdexcept>
```
jeaye, line 7: I'd be surprised if this is needed, but, if it is, please put it first, separated from the jank includes with an empty line.

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,118 @@
+#include "jank/runtime/detail/object_util.hpp"
+#include "jank/runtime/erasure.hpp"
+#include "jank/type.hpp"
```
jeaye, line 3: Please use angle brackets <> for all jank includes. However, I don't think you actually need these includes, since they're in the pre-compiled header (PCH).

include/cpp/jank/runtime/obj/transient_vector.hpp
```diff
@@ -0,0 +1,60 @@
...
+    native_hash to_hash() const;
+
+    /* behavior::countable */
+    size_t count() const;
+
+    /* behavior::consable_in_place */
+    native_box<static_object> cons_in_place(object_ptr head);
+
+    /* behavior::persistentable */
+    native_box<persistent_type> to_persistent();
```
jeaye, line 41: This can be const. It could be that it wasn't in my original transient, which is my fault. But that's the benefit of code review. :grin:

jeaye:
Some comments for things to improve, but overall this is great! Be sure to test it all in `./build/jank repl`, since we don't yet have a test suite for these things.

src/cpp/jank/runtime/obj/transient_vector.cpp
Samy-33: It's the LSP, lol

include/cpp/jank/runtime/obj/transient_vector.hpp
Samy-33: As discussed, in the implementation of the method we are setting the `active = false`, so this should not be the case.

Samy-33:
Tests:
```
➜  jank git:(transient-vector) ✗ ./build/jank repl
Bottom of clojure.core
> (def a (transient [1 2 3]))
#'clojure.core/a
> (count a)
3
> (get a 1)
2
> (a 1)
2
> (get a -1)
nil
> (a -1)
Exception: Index out of bound; index = -1, count = 3
> (persistent! a)
[1 2 3]
> (count a)
libc++abi: terminating due to uncaught exception of type std::runtime_error: transient used after it's been made persistent
[1]    94093 abort      ./build/jank repl
➜  jank git:(transient-vector) ✗ git status

```

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,172 @@
...
+  {
+  }
+
+  obj::transient_vector::static_object(runtime::detail::native_persistent_vector const &d)
+    : data{ d.transient() }
+  {
+  }
+
+  obj::transient_vector::static_object(runtime::detail::native_transient_vector &&d)
+    : data{ d }
```
jeaye, line 22: Be sure to use `std::move` here. `data{ std::move(d) }`

src/cpp/jank/runtime/behavior/callable.cpp
```diff
@@ -1,3 +1,4 @@
+#include "jank/runtime/obj/persistent_vector.hpp"
```
jeaye, line 1: Check if you need this. If you do, please put it after the standard includes and use <>.

include/cpp/jank/runtime/obj/transient_vector.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
+
+#include <jank/runtime/object.hpp>
```
jeaye, line 3: I suspect you don't need this, since it's the earliest part of the runtime.

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,172 @@
+#include <jank/runtime/detail/object_util.hpp>
+#include <jank/runtime/erasure.hpp>
+#include <jank/runtime/obj/persistent_vector_sequence.hpp>
+#include <jank/runtime/util.hpp>
+#include <jank/runtime/obj/native_function_wrapper.hpp>
+#include <jank/runtime/obj/transient_vector.hpp>
+#include <jank/type.hpp>
```
jeaye, line 7: I would figure you don't need half of these, since they're included so early in the runtime. Do you?

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,172 @@
...
+      return data[i];
+    }
+    else
+    {
+      throw std::runtime_error{ fmt::format("key must be an integer; found {}",
+                                            runtime::detail::to_string(idx)) };
+    }
+  }
+
+  object_ptr obj::transient_vector::get(object_ptr idx) const
```
jeaye, line 93: `idx` can be `const`.

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,172 @@
...
+      return data[i];
+    }
+    else
+    {
+      throw std::runtime_error{ fmt::format("key must be an integer; found {}",
+                                            runtime::detail::to_string(idx)) };
+    }
+  }
+
+  object_ptr obj::transient_vector::get(object_ptr idx, object_ptr fallback) const
```
jeaye, line 113: `idx` and `fallback` can be `const`.

src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -0,0 +1,164 @@
+namespace jank::runtime
+{
+  obj::transient_vector::static_object(runtime::detail::native_persistent_vector &&d)
+    : data{ std::move(d).transient() }
```
Samy-33, line 4: This, because of LSP warning about `d` not getting moved.

PR #62, opened 2024-03-16
feat(transient): adds conj!, assoc!, dissoc! and pop! for transients

Samy-33:


src/jank/clojure/core.jank
```diff
@@ -512,6 +512,27 @@
           (recur ret (first ks) (next ks))
           ret)))))
 
+(def pop!)
```
Samy-33, line 515: @jeaye, what's the purpose of `(def pop!)` before an actual `(def pop! ...)`?

jeaye: That's only necessary to first declare `pop!`, in the case where it's used recursively. Please use `defn` for all of these fns, not `def` and `fn*`. Also, there's no need for the declaration ahead of time. `defn` does that automatically.

Also, please use `let`, rather than `let*`.

src/jank/clojure/core.jank
```diff
@@ -431,6 +431,108 @@
...
+     (native/raw "__value = visit_object
+                  (
+                    [&](auto const typed_t, auto const head) -> object_ptr
+                    {
+                      using T = typename decltype(typed_t)::value_type;
+
+                        if constexpr(behavior::persistentable<T>)
+                        {
+                          return typed_t->cons_in_place(head);
+                        }
```
jeaye, line 451: Single statement expressions which can fit on one line get wrapped up like `{ this; }`. Check out the `else` right below.

Also, the `if` here is indented futher than the `using`. Both should be 2 spaces into the `{`.

Formatting of inline C++ is currently unsolved. I'd love a clang-format extension which can work with this and keep uniformity with our C++ code base. Until then, best to be consistent with what I have.

src/jank/clojure/core.jank
```diff
@@ -431,6 +431,108 @@
...
+    ([coll]
+     (native/raw "__value = visit_object
+                  (
+                    [&](auto const typed_t) -> object_ptr
+                    {
+                      using T = typename decltype(typed_t)::value_type;
+
+                      if constexpr(std::same_as<T, obj::transient_vector>)
+                      {
+                        return typed_t->pop();
```
jeaye, line 527: All of the other in-place functions were named `foo_in_place`. We should do the same for `pop`.

src/jank/clojure/core.jank
```diff
@@ -431,6 +431,108 @@
...
+                   );
+                 "))))
+
+(def assoc!)
+(def assoc!
+  (fn* assoc!
+    ([coll k v]
+     (native/raw "__value = visit_object
+                  (
+                    [&](auto const typed_t, auto const key, auto const val) -> object_ptr
```
jeaye, line 466: If you're not capturing anything in this lambda, you can use `[]` instead of `[&]`. This will generate much more efficient code, since we won't need a closure. It'll turn into a standalone free fn instead, which can then be easily inlined.

src/jank/clojure/core.jank
```diff
@@ -431,6 +431,92 @@
...
 
+(defn conj!
+  ([]
+   (transient []))
+  ([coll]
+   coll)
+  ([coll x]
+   (native/raw "__value = visit_object
+                (
+                  [=](auto const typed_t, auto const head) -> object_ptr
```
jeaye, line 442: Looks like these were changed from `[&]` to `[=]`, but I think they can be `[]`. Did you hit an issue with that?

src/jank/clojure/core.jank
```diff
@@ -431,6 +431,92 @@
...
+                  {
+                    using T = typename decltype(typed_t)::value_type;
+
+                    if constexpr(behavior::persistentable<T>)
+                    { return typed_t->cons_in_place(head); }
+                    else
+                    { throw #{ (ex-info :not-persistentable {:o coll}) }#; }
```
jeaye, line 449: In the latest main branch, interpolation has been updated to `~{ foo }` instead of `#{ foo }#`. Merging that in will be necessary first, so that this inline C++ can work.

src/jank/clojure/core.jank
```diff
@@ -431,6 +431,92 @@
...
+   coll)
+  ([coll x]
+   (native/raw "__value = visit_object
+                (
+                  [=](auto const typed_t, auto const head) -> object_ptr
+                  {
+                    using T = typename decltype(typed_t)::value_type;
+
+                    if constexpr(behavior::persistentable<T>)
+                    { return typed_t->cons_in_place(head); }
```
jeaye, line 447: The behavior for `cons_in_place` is `behavior::consable_in_place`. The behavior for `persistent` is `behavior::persistentable`.

src/jank/clojure/core.jank
```diff
@@ -431,6 +431,92 @@
...
+                  {
+                    using T = typename decltype(typed_t)::value_type;
+
+                    if constexpr(std::same_as<T, obj::transient_vector>)
+                    { return typed_t->pop_in_place(); }
+                    else
+                    { throw #{ (ex-info :not-transient-vector {:o coll}) }#; }
+                  },
+                  #{ coll }#
+                );
```
jeaye, line 517: `visit_object` is very expensive, at compile-time, since it results in an instantiation of the lambda for every object type (nearly 30 right now). For the other transient fns, we're supporting multiple object types using behaviors, so we're not checking for transient maps, sets, or vectors specifically. However, for `pop!`, we really only care about transient vectors. When we use `visit_object` for this, only one instantiation of the lambda is useful and all of the others just throw an exception. It would be much more efficient, at compile-time, to instead just check for `~{ coll }->type == object_type::transient_vector`.

The difference between a dozen `visit_object` calls versus explicit type checks can actually be multiple full seconds of compilation time. Since we JIT compile code in Clojure, compile-time performance can also mean run-time performance.

Tangentially, this is also why I have `visit_seqable`, which is like `visit_object`, but only instantiates the lambda for known seqable types. This is a very common thing we need to do and cutting down the number of those lambda instantiations to the known seqable types can significantly speed up compilation.

src/jank/clojure/core.jank
```diff
@@ -431,6 +431,92 @@
...
+  ([]
+   (transient []))
+  ([coll]
+   coll)
+  ([coll x]
+   (native/raw "__value = visit_object
+                (
+                  [=](auto const typed_t, auto const head) -> object_ptr
+                  {
+                    using T = typename decltype(typed_t)::value_type;
```
jeaye, line 444: The pattern used is that `foo` becomes `typed_foo`. So, in this case, the object we're visiting is `coll` and the unerased (visited) object should be `typed_coll`. Using `T` for the type name, especially when there's only one, is just one of those C++ idioms that exists.

jeaye: Similarly, if the param is called `x` in the `defn`, let's call it `x` when we visit. Or call it something else, but let's keep that naming.

include/cpp/jank/runtime/behavior/associatively_writable.hpp
```diff
@@ -14,7 +14,7 @@ namespace jank::runtime::behavior
   template <typename T>
   concept associatively_writable_in_place = requires(T * const t) {
     {
-      t->assoc_in_place(object_ptr{}, object_ptr{})
+      t->assoc_in_place(object_ptr{}, object_ptr{}), t->dissoc_in_place(object_ptr{})
```
jeaye, line 17: This needs to have its own entry. Concepts here, with the `requires` expression, are saying `<expression> -> <type>`, meaning `<this expression> results in <some type>`. For an example of this, you can look at `seqable.hpp`.

include/cpp/jank/runtime/obj/transient_vector.hpp
```diff
@@ -23,6 +23,8 @@ namespace jank::runtime
       return ret;
     }
 
+    native_box<static_object> pop_in_place();
+
```
jeaye, line 27: This is a nit pick, but let's move this down below the other behaviors. Currently, the constructors go first, then the objectable behavior, followed by other behaviors, and then utilities.

src/jank/clojure/core.jank
Samy-33: Yes, the compiler's mad at me:
```cpp
➜  jank git:(transient-vector-fns) ✗ ./bin/compile
[1/1] Generating classes/core-libraries
input_line_94:18:29: error: variable 'call_574' cannot be implicitly captured in a lambda with no capture-default specified
                    { throw call_574; }
                            ^
input_line_94:9:243: note: 'call_574' declared here
        jank::profile::timer __timer{ "dissoc_BANG__558" };auto const map_575(jank::make_box<jank::runtime::obj::persistent_array_map>(jank::runtime::detail::in_place_unique{}, jank::make_array_box<object_ptr>(const_561, coll),2));auto const call_574(jank::runtime::dynamic_call(ex_info_559->deref(), const_560, map_575));object_ptr native_573{ obj::nil::nil_const() };{ object_ptr __value{ obj::nil::nil_const() };__value = visit_object
                                                                                                                                                                                                                                                  ^
input_line_94:11:19: note: lambda expression begins here
                  [](auto const typed_t, auto const key) -> object_ptr
                  ^
input_line_94:11:20: note: capture 'call_574' by value
                  [](auto const typed_t, auto const key) -> object_ptr
                   ^
                   call_574
```

jeaye: Ah, that's because you're throwing `coll` instead of `typed_coll`. Since you have the fully typed version as the first param to the lambda, we can use that rather than also capturing the erased version.

src/jank/clojure/core.jank
```diff
@@ -418,11 +418,91 @@
...
+   (native/raw "__value = visit_object
+                (
+                  [=](auto const typed_coll, auto const head) -> object_ptr
+                  {
+                    using T = typename decltype(typed_coll)::value_type;
+
+                    if constexpr(behavior::consable_in_place<T>)
+                    { return typed_coll->cons_in_place(head); }
+                    else
+                    { throw ~{ (ex-info :not-persistentable {:o coll}) }; }
```
jeaye, line 441: `:not-consable-in-place`

PR #64, opened 2024-05-02
Fix if-let and if-some by not using vectors as functions

madstap:
Since calling vectors as functions isn't implemented yet we need to explicitly use `get`


jeaye:
You're right. :man_facepalming: Thank you!

PR #67, opened 2024-05-03
Add transduce; xform arities to into, map, filter, keep ++

madstap:
Also completing, run!, comp

### DIFF
--- a/src/jank/clojure/core.jank
+++ b/src/jank/clojure/core.jank
@@ -882,6 +882,57 @@
                 ~{ coll }
                 );")))
 
+(defn completing
+  "Takes a reducing function f of 2 args and returns a fn suitable for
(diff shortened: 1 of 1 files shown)


jeaye:
Nice! Excellent timing, too. I was just about to pushed chunked seqs, which will cause conflicts with this. I'll merge yours first and deal with the conflicts myself.

Thank you! If you're looking to tackle some more and need ideas, hit me up in Slack. :slightly_smiling_face:

jeaye:
Just tested this locally and it looks great. I updated `keep` to have a doc string and chunked seq support, matching Clojure. Once you've signed the CLA, I'll get this in.

You didn't have to sign it for the last PR, since it was just a couple of lines.

PR #69, opened 2024-05-04
More transducers

madstap:


src/jank/clojure/core.jank
```diff
@@ -927,6 +927,25 @@
          ret (reduce f init coll)]
      (f ret))))
 
+;; TODO: private
+(defn preserving-reduced
```
jeaye, line 931: We can use `^:private` here, since jank supports meta hints now. It just won't be marked as private; I don't think I'll implement private vars in jank, since Clojure does a halfway job of them and ClojureScript doesn't bother.

src/cpp/jank/runtime/obj/volatile.cpp
```diff
@@ -38,10 +38,10 @@ namespace jank::runtime
     return val;
   }
 
-  obj::volatile_ptr obj::volatile_::reset(object_ptr const o)
+  object_ptr obj::volatile_::reset(object_ptr const o)
   {
     val = o;
     assert(val);
-    return this;
+    return val;
```
jeaye, line 45: Ah, good catch with this.

PR #71, opened 2024-05-05
Add more pure core fns

madstap:
juxt, partial, doto, map-indexed (- seq), keep-indexed (- seq), frequencies, reductions, distinct, distinct?, dedupe


jeaye:
Thanks!

PR #75, opened 2024-05-18
Fix Ubuntu version in README

wodin:


wodin:
I tried it out on Ubuntu 22.04 and found that the libstdc++6 package was too old. Then I tried it on Ubuntu 24.04 and it worked.

FYI, I ran it in docker and had to install the following packages:
* libatomic1
* libjemalloc2
* libreadline8

jeaye:
Hey! Thanks for the report. The errors you're getting are due to your C++ compiler not being new enough, looks like; it's not understanding C++20 concepts, constexpr destructors, and so on.

Let me spin up a 24.04 machine this weekend and I'll see what can be done. If you'd like to test more before that, just make sure you're using the latest Clang!

jeaye:
Hey. Just checking in. I've verified that jank builds properly on Ubuntu 24.04, using the commands in the build doc. 

I can't merge this PR, as it stands, since the line that was changed is regarding the pre-compiled jank binaries. Those are indeed still build on Ubuntu 22.04. If you'd like to update the Github workflow file to try building them on 24.04, that would complete the change. My guess it that it'll require some debugging work, though, so I'm going to close this for now.

If you'd like to push that forward, feel free to open this back up and update your changes. If you need more support for compiling jank on Ubuntu 24.04, let me know. I just started in a clean distrobox, ran the commands listed, and everything is working.

PR #78, opened 2024-06-28
feat: adds character

Samy-33:


jeaye (changes requested):

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -86,6 +86,83 @@ namespace jank::read::parse
...
+
+        size_t offset{};
+        for(auto const &ch : { 'a', '1', '`', ':', '#' })
+        {
+          auto const r(p.next());
+          CHECK(runtime::detail::equal(r.expect_ok().unwrap().ptr,
+                                       make_box<runtime::obj::character>(ch)));
+
+          CHECK(r.expect_ok().unwrap().start
+                == lex::token{ offset, 2, lex::token_kind::character, std::string(1, ch) });
```
jeaye, line 104: Please use `native_persistent_string` for these.

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -86,6 +86,83 @@ namespace jank::read::parse
       CHECK(r.expect_ok().unwrap().end == r.expect_ok().unwrap().start);
     }
 
+    TEST_CASE("Character")
+    {
+      SUBCASE("single")
```
jeaye, line 91: All of the existing subcases start with an uppercase. Please follow that!

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -86,6 +86,83 @@ namespace jank::read::parse
       CHECK(r.expect_ok().unwrap().end == r.expect_ok().unwrap().start);
     }
 
+    TEST_CASE("Character")
```
jeaye, line 89: What happens when we have `\\`?

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -86,6 +86,83 @@ namespace jank::read::parse
...
+                               std::make_pair<native_persistent_string, char8_t>("tab", '\t'),
+                               std::make_pair<native_persistent_string, char8_t>("`", '`'),
+                               std::make_pair<native_persistent_string, char8_t>("space", ' ') })
+        {
+          auto const r(p.next());
+          CHECK(runtime::detail::equal(r.expect_ok().unwrap().ptr,
+                                       make_box<runtime::obj::character>(ch.second)));
+
+          /* +1 for backslash */
+          auto const len = ch.first.size() + 1;
```
jeaye, line 156: Stylistically, all of the jank code always uses direct initialization.

```cpp
auto const len(ch.first.size() + 1);
```

compiler+runtime/include/cpp/jank/runtime/obj/character.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+  template <>
+  struct static_object<object_type::character> : gc
+  {
+    static constexpr native_bool pointer_free{ true };
+
+    static_object() = default;
+    static_object(static_object &&) = default;
+    static_object(static_object const &) = default;
+    static_object(native_persistent_string_view const &);
```
jeaye, line 15: Let's add an overload which takes `native_persistent_string const &` as well, please. Let's also add an overload which takes a `char`.

compiler+runtime/src/cpp/jank/runtime/obj/character.cpp
```diff
@@ -0,0 +1,38 @@
...
+  native_persistent_string obj::character::to_string() const
+  {
+    fmt::memory_buffer buff;
+    to_string(buff);
+    return native_persistent_string{ buff.data(), buff.size() };
+  }
+
+  native_hash obj::character::to_hash() const
+  {
+    return static_cast<native_hash>(util::character::get_char_from_repr(data).unwrap());
```
jeaye, line 36: This function is just used once, from within `character.cpp`. We should not have a whole header/source for it. It can be a `static` function inside this cpp file.

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -582,6 +582,92 @@ namespace jank::read::lex
...
+        CHECK(tokens
+              == make_tokens({
+                { 0, 2, token_kind::character,       "\\1"sv},
+                { 3, 8, token_kind::character, "\\newline"sv},
+                {11, 2, token_kind::character,       "\\'"sv},
+                {14, 2, token_kind::character,      "\\\\"sv}
+        }));
+      }
+
+      SUBCASE("character followed by a backticked keyword")
```
jeaye, line 658: Upper case this, please.

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -1,3 +1,4 @@
+#include "jank/type.hpp"
```
jeaye, line 1: This header is:

1. Not needed
2. Using quotes instead of angle brackets

Please change your editor to not automatically add includes! Each include slows down jank's build system. You only need an include if you try to use something and find it doesn't exist AND you can't forward declare it.

compiler+runtime/src/cpp/jank/runtime/obj/character.cpp
```diff
@@ -0,0 +1,38 @@
+#include <jank/runtime/obj/character.hpp>
+#include <jank/util/character.hpp>
+
+namespace jank::runtime
+{
+  obj::character::static_object(native_persistent_string_view const &d)
+    : data{ native_persistent_string(d) }
```
jeaye, line 7: `data` is a `native_persistent_string`, so there's no need to build another `native_persistent_string` here. What you're doing is:

1. Building a `native_persistent_string` from `d`
2. Building `data` from that temporary `native_persistent_string`

It's less efficient and it's more code than we need: `data{ d }`

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -456,9 +492,16 @@ namespace jank::read
               return err(std::move(e.unwrap()));
             }
 
-            /* Support auto-resolved qualified keywords. */
             auto const oc(peek());
-            if(oc.is_some() && oc.unwrap() == ':')
+            if(oc.is_none() || std::isspace(oc.unwrap()))
```
jeaye, line 496: Nice!

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -79,6 +79,14 @@ namespace jank::codegen
             }
             return "jank::native_integer";
           }
+        case jank::runtime::object_type::character:
+          {
+            if(boxed)
+            {
+              return "jank::runtime::obj::character_ptr";
+            }
+            return "jank::native_persistent_string";
```
jeaye, line 88: This is not correct. We're constructing a character, not a string.

compiler+runtime/include/cpp/jank/runtime/detail/object_util.hpp
```diff
@@ -37,6 +37,7 @@ namespace jank
     using persistent_string = static_object<object_type::persistent_string>;
     using persistent_list = static_object<object_type::persistent_list>;
     using symbol = static_object<object_type::symbol>;
+    using character = static_object<object_type::character>;
   }
 
   /* TODO: Constexpr these. */
```
jeaye, line 43: Let's add a `make_box` overload below which takes a `char` and builds a character.

compiler+runtime/include/cpp/jank/runtime/obj/character.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+    /* behavior::objectable */
+    native_bool equal(object const &) const;
+    native_persistent_string to_string() const;
+    void to_string(fmt::memory_buffer &buff) const;
+    native_hash to_hash() const;
+
+    object base{ object_type::character };
+    native_persistent_string data;
```
jeaye, line 24: Please add a comment describing that `data` holds the literal form of the character, as it's written. For example `"\\tab"`.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -186,6 +194,12 @@ namespace jank::codegen
                              typed_o->name);
             }
           }
+          else if constexpr(std::same_as<T, runtime::obj::character>)
+          {
+            fmt::format_to(inserter,
+                           R"(jank::make_box<jank::runtime::obj::character>({}))",
```
jeaye, line 200: I don't think this will work. Have you tried doing something like `(fn [] \space)` in a repl?

compiler+runtime/src/cpp/jank/runtime/obj/character.cpp
```diff
@@ -0,0 +1,82 @@
...
+    : data{ d }
+  {
+  }
+
+  obj::character::static_object(native_persistent_string const &d)
+    : data{ d }
+  {
+  }
+
+  obj::character::static_object(char ch)
```
jeaye, line 50: `ch` should be `const` here, since it's a value we're never changing.

compiler+runtime/src/cpp/jank/runtime/obj/character.cpp
```diff
@@ -0,0 +1,82 @@
...
+  void obj::character::to_string(fmt::memory_buffer &buff) const
+  {
+    fmt::format_to(std::back_inserter(buff), "{}", data);
+  }
+
+  native_persistent_string obj::character::to_string() const
+  {
+    fmt::memory_buffer buff;
+    to_string(buff);
+    return native_persistent_string{ buff.data(), buff.size() };
```
jeaye, line 75: This is copy pasta from other objects, but it's not needed. We already have the string representation of the character. Just return it!

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -196,9 +197,10 @@ namespace jank::codegen
           }
           else if constexpr(std::same_as<T, runtime::obj::character>)
           {
-            fmt::format_to(inserter,
-                           R"(jank::make_box<jank::runtime::obj::character>({}))",
-                           typed_o->data);
+            fmt::format_to(
+              inserter,
+              R"(jank::make_box<jank::runtime::obj::character>(jank::runtime::get_char_from_repr({}).unwrap()))",
```
jeaye, line 202: Why do we call `get_char_from_repr` here and not in the ctor for character? Again, think about the API design here. jank is a C++ library.

Do we want everyone who constructs a character from a string to also need to call `get_char_from_repr`? No, we just want them to give us the string and have us do the right thing.

jeaye:
Great job, Saket!

PR #80, opened 2024-07-11
feat: adds persistent_string_sequence

Samy-33:


compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,113 @@
...
+                      str->data.end(),
+                      seq->str->data.begin() + seq->index,
+                      seq->str->data.end());
+  }
+
+  void obj::persistent_string_sequence::to_string(fmt::memory_buffer &buff) const
+  {
+    return behavior::detail::to_string(
+      str->data.begin()
+        + static_cast<decltype(obj::persistent_string::data)::difference_type>(index),
```
jeaye, line 36: Why was this cast needed, rather than just doing `+ index` (like you did on line 26)?

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,113 @@
+#include <jank/runtime/obj/persistent_string_sequence.hpp>
+
+namespace jank::runtime
+{
+  obj::persistent_string_sequence::static_object(obj::persistent_string_ptr s)
```
jeaye, line 5: Please make all value params `const` unless you're mutating them.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,113 @@
...
+    : str{ s }
+  {
+    assert(s->count() > 0);
```
jeaye, line 8: clang-tidy will complain about this, since using `!s->empty()` is more idiomatic.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,113 @@
...
+  {
+    if(o.type != object_type::persistent_string_sequence)
+    {
+      return false;
+    }
+    auto const seq(expect_object<obj::persistent_string_sequence>(&o));
+    return std::equal(str->data.begin() + index,
+                      str->data.end(),
+                      seq->str->data.begin() + seq->index,
+                      seq->str->data.end());
```
jeaye, line 29: Hmmm, I see. I think I've led you astray here, since I didn't quite see what you needed. In my example of `persistent_string::equal`, it will check if the rhs is a string first, since a string will only every be equal to a string. Sequences are different though. Check it:

```
user=> (= "foo" '(\f \o \o))
false
user=> (= (seq "foo") '(\f \o \o))
true
```

So, we need to `visit_seqable` on rhs and walk through the sequences, comparing by using your new `equal(char, object_ptr)` fn. There's a smart way we can do this, but it will require some more explanation.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,113 @@
...
+      str->data.end(),
+      "(",
+      ')',
+      buff);
+    return { buff.data(), buff.size() };
+  }
+
+  native_hash obj::persistent_string_sequence::to_hash() const
+  {
+    return str->data.substr(index).to_hash();
```
jeaye, line 58: This is a good attempt. It's not how Clojure's hashes work though. Hashing isn't arbitrary, but rather very intentional, since it's used for light equality checking. Sequences are hashed based on whether or not they're ordered. Ordered sequences with the same contents have the same hash. Check it:

```clojure
user=> (hash '(\f \o \o))
-1629510009
user=> (hash "foo")
493551392
user=> (hash [\f \o \o])
-1629510009
```

jank has a bunch of hash functions in `hash.hpp` for this. We'll want to use the iteretor-based `ordered`, but that will require a `char` overload for `visit`. Check out the header to see what I mean. From there, you need to figure out how to hash chars.

```clojure
user=> (hash \a)
97
user=> (hash \A)
65
```

Looks like Clojure just uses the integer value.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+  /* behavior::seqable */
+  obj::persistent_string_sequence_ptr obj::persistent_string_sequence::seq()
+  {
+    return this;
+  }
+
+  obj::persistent_string_sequence_ptr obj::persistent_string_sequence::fresh_seq() const
+  {
+    return jank::make_box<obj::persistent_string_sequence>(str, index);
```
jeaye, line 75: No need for the `jank::` prefix here.

compiler+runtime/src/cpp/jank/runtime/detail/object_util.cpp
```diff
@@ -7,6 +7,11 @@ namespace jank::runtime::detail
     return visit_object([](auto const typed_o) { return typed_o->to_string(); }, o);
   }
 
+  void to_string(char const ch, fmt::memory_buffer &buff)
+  {
+    make_box(ch)->to_string(buff);
```
jeaye, line 12: There is no need to do an allocation here. If you need a `character` temporarily, just create one locally.

```cpp
obj::character{ ch }.to_string(buff);
```

compiler+runtime/include/cpp/jank/runtime/detail/type.hpp
```diff
@@ -12,6 +12,7 @@
 
 namespace jank::runtime::detail
 {
+  bool equal(char const lhs, object_ptr const rhs);
```
jeaye, line 15: Please use `native_bool`. We have this alias in case we ever change the bool type. We'll be able to change it through the entire compiler and runtime using one line.

compiler+runtime/src/cpp/jank/runtime/detail/object_util.cpp
```diff
@@ -30,6 +35,17 @@ namespace jank::runtime::detail
       o);
   }
 
+  bool equal(char const lhs, object_ptr const rhs)
```
jeaye, line 38: Please use `native_bool` here.

compiler+runtime/src/cpp/jank/runtime/detail/object_util.cpp
```diff
@@ -30,6 +35,17 @@ namespace jank::runtime::detail
...
+  {
+    if(!rhs || rhs->type != object_type::character)
+    {
+      return false;
+    }
+
+    auto const typed_rhs = expect_object<obj::character>(rhs);
+    return typed_rhs->to_hash() == static_cast<unsigned int>(lhs);
```
jeaye, line 46: Please use `native_hash` instead of `unsigned int`.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,93 @@
...
+    : str{ s }
+    , index{ i }
+  {
+    assert(!s->data.empty() && i < s->data.size());
+  }
+
+  /* behavior::objectable */
+  native_bool obj::persistent_string_sequence::equal(object const &o) const
+  {
+    return detail::equal(o, str->data.begin(), str->data.end());
```
jeaye, line 21: Bug here: missing `+ index`. This will mean that `(= (next (seq "foo")) (seq "foo"))`, which is incorrect.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,93 @@
...
+  native_persistent_string obj::persistent_string_sequence::to_string() const
+  {
+    fmt::memory_buffer buff;
+    behavior::detail::to_string(str->data.begin() + index, str->data.end(), "(", ')', buff);
+    return { buff.data(), buff.size() };
+  }
+
+  native_hash obj::persistent_string_sequence::to_hash() const
+  {
+    return hash::ordered(str->data.begin(), str->data.end());
```
jeaye, line 38: Bug here: missing `+ index`. This will mean `(= (hash (next (seq "foo"))) (hash (seq "foo")))`, which is incorrect.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -0,0 +1,93 @@
...
+  {
+    auto n(index);
+    ++n;
+
+    if(n == str->data.size())
+    {
+      return nullptr;
+    }
+
+    return jank::make_box<obj::persistent_string_sequence>(str, n);
```
jeaye, line 74: Please remove `jank::` here. It's not needed (though it used to be at one point, which is why you may see it in older code; this is copy pasta).

compiler+runtime/src/cpp/jank/runtime/detail/object_util.cpp
```diff
@@ -30,6 +35,17 @@ namespace jank::runtime::detail
...
+  }
+
   bool equal(object_ptr const lhs, object_ptr const rhs)
```
jeaye, line 49: You're likely just following the usage of this `bool`, rather than `native_bool`, but that one should be changed, too. My code doesn't get reviewed though, so sometimes these things slip though. :grin:

compiler+runtime/src/cpp/jank/runtime/obj/persistent_vector_sequence.cpp
```diff
@@ -18,7 +18,11 @@ namespace jank::runtime
   /* behavior::objectable */
   native_bool obj::persistent_vector_sequence::equal(object const &o) const
   {
-    return detail::equal(o, vec->data.begin(), vec->data.end());
+    return detail::equal(
```
jeaye, line 21: I spotted these and fixed them on my clang 19 branch, since they were leading to hard crashes. :facepalm: Also led to iterations on your PR. My own bugs. We can keep this and I'll just resolve the conflict when clang 19 merges.

jeaye:
Looks good to me! No other feedback for now. It's late, though, so rather than merging now I'll check out your branch tomorrow after work to test everything locally.

Thanks, Saket!

PR #93, opened 2024-09-28
feat(shell.nix): Edit shell.nix to use reproducible development environment

haruki7049:
The `/shell.nix` seemed to use `<nixpkgs>` when I checked. `<nixpkgs>` is dependent on the individual's environment, so it is not possible to build a reproducible development environment. So I rewrote `/shell.nix` and wrote `/flake.nix`. The reason why I wrote `/flake.nix` is not only to guarantee reproducibility by `/flake.lock`, but also to use useful features such as code formatter and code linter.

It is completely up to you whether you use the code formatter in `/flake.nix` or not, so you can remove it if you find it unnecessary. However, the previous way of writing `/shell.nix` obviously does not guarantee reproducibility of the development environment, so I created this PR.


jeaye:
Thanks for the PR! I'm interested in this but I'd like to understand it more. I run NixOS as my main desktop driver, but building jank with Nix has proven to be quite difficult, so all of my development actually happens in various distroboxes. Even on the rest of my NixOS machine, I haven't jumped on the flake train yet, though I hear about flakes so often these days.

The main benefit, I gather, here is that we're locking down specific package versions, rather than just fetching the latest unstable. However, that part alone is done by just your tarball fetching change to `shell.nix`, so what does the `flake.nix` and additional `flake.lock` provide on top of that?

If you're interested in contributing further, we could really benefit from getting jank building within Nix, which is also the first step of getting it into nixpkgs. Last I tried, jank's vcpkg dependencies were an issue.

haruki7049:
I said that:

> The reason why I wrote /flake.nix is not only to guarantee reproducibility by /flake.lock, but also to use useful features such as code formatter and code linter.

The code formatter and code linter can be used by adding `flake.nix`, so I added.

The below code block is the `flake.nix` in my personal project. This file serves some functions, code formatter by [treefmt-nix](https://github.com/numtide/treefmt-nix), and `devShells.default` such as `shell.nix`.

```nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
    flake-parts.url = "github:hercules-ci/flake-parts";
    treefmt-nix.url = "github:numtide/treefmt-nix";
  };

  outputs =
    inputs:
    inputs.flake-parts.lib.mkFlake { inherit inputs; } {
      systems = [
        "x86_64-linux"
      ];

      imports = [
        inputs.treefmt-nix.flakeModule
      ];

      perSystem =
        { pkgs, lib, ... }:
        {
          treefmt = {
            projectRootFile = "flake.nix";
            programs.nixfmt.enable = true;
            programs.actionlint.enable = true;
          };

          devShells.default = pkgs.mkShell {
            packages = [
              # CLASP
              pkgs.clasp-common-lisp

              # LSP
              pkgs.nil
            ];

            shellHook = ''
              export PS1="\n[nix-shell:\w]$ "
            '';
          };
        };
    };
}
```

I am very interested in building Jank-lang by Nix, but there is the same problem about Nix with vcpkg...

haruki7049:
By the way, Why did you select `vcpkg` package manager?

jeaye:
> By the way, Why did you select `vcpkg` package manager?

C++ lacks a standard package manager, but I wanted something which resembled Cargo or NPM. I tried a few different options, early in jank's development, and vcpkg was the only one which worked (somewhat) reliably in a cross-platform manner. At that point, and maybe still now, Nix on macOS was a big hassle. Windows is a tertiary concern, but I'd like to not shut it out entirely.

I do see jank moving away from vcpkg. From the packages we're bringing in, most are either in a typical distro's package manager or are header-only, so we can just use a submodule: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/vcpkg.json

If pulling vcpkg out of jank is of interest to you, so we can get it building properly with Nix, you have my blessing. :heart:

haruki7049:
I understand. I'm going to remove flakes from this pull request for now and create an issue to think about regarding building with Nix.

haruki7049:
I created #94 to set up `/compiler+runtime/default.nix`

PR #95, opened 2024-10-04
Fix compiling and linking config for macOS

### LINKED ISSUE #86: Reported build failure with double-conversion and openssl on macOS
A user in Slack reported macOS build failures for jank, after running `./bin/compile`:

```
jank/compiler+runtime/third-party/folly/folly/Conv.h:124:10: fatal error: 'double-conversion/double-conversion.h' file not found
jank/compiler+runtime/src/cpp/jank/util/sha256.cpp:3:10: fatal error: 'openssl/sha.h' file not found
```

The user had `double-conversion` and `openssl` installed via homebrew. I suspect that we need to find these packages via cmake. This wasn't a problem before, since we used vcpkg to build folly, but now that we build it as part of jank's compilation, we'll need to meet its dependencies.

I suspect this hasn't been an issue on Linux, since those headers are standard places. Homebrew puts them somewhere non-standard.

This should be a reasonably straight-forward CMake change for anyone on macOS.

avitkauskas:
Fixes #86 as discussed there.

I built clang and jank from scratch with the updated CMakeLists.txt and following the udpated instructions for macOS without any problems now.

### DIFF
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -62,6 +62,7 @@ endif()
 
 set(
   jank_aot_compiler_flags
+  $<$<PLATFORM_ID:Darwin>:-I/opt/homebrew/include>
   -Wall -Wextra -Wpedantic
   -Wfloat-equal -Wuninitialized -Wswitch-enum -Wnon-virtual-dtor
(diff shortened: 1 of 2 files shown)


jeaye:
Great work! Thanks for testing this on a clean clone, too. :heart:

PR #101, opened 2024-10-06
Add support for escape character \r

jianlingzhong:


jeaye:
Perfect! First PR merged. Next step will be getting those other escape sequences going. :muscle:

PR #102, opened 2024-10-06
Add support for all remaining simple escape sequences

### LINKED ISSUE #97: Support all simple string escape sequences
Currently, we support only `\n` and `\t` properly. We have partial support for `\r`, too. Let's support all of the simple escape sequences listed here:  https://en.cppreference.com/w/cpp/language/escape

- [ ] Add lexing support for each escape sequence (in lex.cpp)
- [ ] Add escape/unescape support for each escape sequence (in escape.hpp/cpp)
- [ ] Update the lexing tests to include all supported sequences
- [ ] Update the parse tests to include all supported sequences

jianlingzhong:
This closes #97.

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -566,6 +566,13 @@ namespace jank::read
                   case 't':
                   case 'r':
                   case '"':
+                  case '\'':
+                  case '?':
+                  case '\\':
+                  case 'a':
+                  case 'b':
+                  case 'f':
+                  case 'v':
                     break;
                   default:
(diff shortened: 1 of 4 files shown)


jeaye:
So, there's some missing work here. Check out the steps in the ticket for details: https://github.com/jank-lang/jank/issues/97

We've done the first and third, as part of this PR, but the second and fourth are needed in order for this to fully work. Right now, we will lex these escape sequences, but they won't be parsed properly. To see what I mean, try this:

```
$ jank repl
> "\v"
```

jianlingzhong:
@jeaye updated. could you take another look?

jeaye:
Looks great!

PR #103, opened 2024-10-09
test(flake.nix): Init treefmt-nix and flake.nix

haruki7049:
PR for combining declarations of which formatter to use in the same file, so [treefmt](https://github.com/numtide/treefmt) may be used instead of [treefmt-nix](https://github.com/numtide/treefmt-nix).

## How to use formatter
```bash
# To format all this project's file
nix fmt

# To check all file's format in this project
nix flake check
```

## How to add formatter
Just add an attribute in flake.nix:
```nix
      perSystem =
        { pkgs, ... }:
        {
          treefmt = {
            projectRootFile = "flake.nix";
            programs.shellcheck.enable = true;
            programs.nixfmt.enable = true;

            # Just add formatter, declared in https://github.com/numtide/treefmt-nix#supported-programs
            #programs.clang-format.enable = true;
          };
```


jeaye:
Hi @haruki7049! Thanks for this PR. jank currently requires devs to format their C++ with `clang-format`, but we don't have any other project-wide formatting. I'm not convinced we want this right now, especially tied to Nix. The vast majority of devs will not be using Nix.

I know you put a lot of work into this, which I very much appreciate. For now, I think we should close this; if we want project-wide formatting, in the future, this is where we should start.

To be sure that the work you're doing is going to be merged, assuming there's no ticket for it, you can always check with me. :bow:

PR #106, opened 2024-10-10
feat(runtime-objs): to_code_string

Samy-33:
image

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/obj/detail/base_persistent_map.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/detail/base_persistent_map.hpp
@@ -47,7 +47,7 @@ namespace jank::runtime::obj::detail
     static void to_string_impl(typename V::const_iterator const &begin,
                                typename V::const_iterator const &end,
                                fmt::memory_buffer &buff,
-                               native_bool to_code = false)
+                               native_bool const to_code)
     {
       auto inserter(std::back_inserter(buff));
       inserter = '{';
@@ -86,7 +86,8 @@ namespace jank::runtime::obj::detail
     {
       to_string_impl(static_cast<parent_type const *>(this)->data.begin(),
                      static_cast<parent_type const *>(this)->data.end(),
-                     buff);
+                     buff,
+                     false);
     }
 
     native_persistent_string to_string() const
@@ -94,7 +95,8 @@ namespace jank::runtime::obj::detail
       fmt::memory_buffer buff;
       to_string_impl(static_cast<parent_type const *>(this)->data.begin(),
                      static_cast<parent_type const *>(this)->data.end(),
-                     buff);
+                     buff,
+                     false);
       return native_persistent_string{ buff.data(), buff.size() };
     }
 
--- a/compiler+runtime/include/cpp/jank/runtime/obj/detail/base_persistent_map.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/detail/base_persistent_map.hpp
@@ -46,16 +46,32 @@ namespace jank::runtime::obj::detail
 
     static void to_string_impl(typename V::const_iterator const &begin,
                                typename V::const_iterator const &end,
-                               fmt::memory_buffer &buff)
+                               fmt::memory_buffer &buff,
+                               native_bool to_code = false)
     {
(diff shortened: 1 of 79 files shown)


compiler+runtime/include/cpp/jank/runtime/obj/detail/base_persistent_map.hpp
```diff
@@ -46,16 +46,32 @@ namespace jank::runtime::obj::detail
 
     static void to_string_impl(typename V::const_iterator const &begin,
                                typename V::const_iterator const &end,
-                               fmt::memory_buffer &buff)
+                               fmt::memory_buffer &buff,
+                               native_bool to_code = false)
```
jeaye, line 50: Please make this `const`.

jeaye: Also, please don't use default arguments. We can update all of the calls to this. As soon as there's one default arg, adding any other args requires them to be default args, too.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -565,7 +565,10 @@ namespace jank::read
                   case 'n':
                   case 't':
                   case 'r':
+                  case 'f':
+                  case 'b':
                   case '"':
+                  case '\\':
```
jeaye, line 571: Why were these added?

compiler+runtime/src/cpp/jank/runtime/obj/chunked_cons.cpp
```diff
@@ -186,6 +186,11 @@ namespace jank::runtime
     return runtime::to_string(seq());
   }
 
+  native_persistent_string obj::chunked_cons::to_code_string() const
+  {
+    return to_string();
```
jeaye, line 191: I don't think this is correct. For sequences, we'll need to use `to_code_string` on all of the elements. Should be able to see this with:

```clojure
user=> (cons "foo bar" '("spam"))
("foo bar" "spam")
```

compiler+runtime/src/cpp/jank/runtime/obj/cons.cpp
```diff
@@ -102,6 +102,11 @@ namespace jank::runtime
     return runtime::to_string(seq());
   }
 
+  native_persistent_string obj::cons::to_code_string() const
+  {
+    return to_string();
```
jeaye, line 107: Same concern with `chunked_cons`. This is a sequence. We need to use `runtime::to_code_string(seq());`.

compiler+runtime/src/cpp/jank/runtime/obj/iterator.cpp
```diff
@@ -90,6 +90,11 @@ namespace jank::runtime
     return runtime::to_string(seq());
   }
 
+  native_persistent_string obj::iterator::to_code_string()
+  {
+    return to_string();
```
jeaye, line 95: Same thing with `cons`. We need to use `runtime::to_code_string(seq());`.

compiler+runtime/src/cpp/jank/runtime/obj/range.cpp
```diff
@@ -234,6 +234,11 @@ namespace jank::runtime
     return runtime::to_string(seq());
   }
 
+  native_persistent_string obj::range::to_code_string()
+  {
+    return to_string();
```
jeaye, line 239: I would still do `return runtime::to_code_string(seq());` here, even though we know all `range` elements are numbers. If we end up adding different `to_code_string` implementations for numbers, we want that to reflect when use ranges of those numbers.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -185,7 +185,9 @@
 
 ;; Strings.
 ; TODO: Proper version.
```
jeaye, line 187: We can remove this comment now.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -185,7 +185,9 @@
...
-(def pr-str)
+(def pr-str
+  (fn* pr-str* [o]
```
jeaye, line 189: The fn name should also be `pr-str`.

PR #107, opened 2024-10-12
optimise: stores char data as bytes instead of their code representation

Samy-33:


compiler+runtime/src/cpp/jank/runtime/obj/character.cpp
```diff
@@ -81,23 +47,21 @@ namespace jank::runtime
...
   native_persistent_string obj::character::to_code_string() const
   {
-    return fmt::format("\\{}", data);
+    return get_literal_from_char(data[0]);
   }
 
   native_hash obj::character::to_hash() const
   {
-    return hash::visit(get_char_from_literal(data).unwrap());
+    return hash::visit(data[0]);
```
jeaye, line 65: We want to hash the whole string. This will be important for multi-byte chars (i.e. unicode support).

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -134,19 +136,37 @@ namespace jank::read::parse
...
         }
       }
+
+      SUBCASE("Invalid character literal")
+      {
+        lex::processor lp{ R"(\ne\apple\backspace)" };
+        processor p{ lp.begin(), lp.end() };
+
+        // First two lex tokens are invalid characters i.e. \ne and \apple
+        for(int _ignored = 1; _ignored <= 2; _ignored++)
```
jeaye, line 160: Two stylistic points.

1. We always use `/* */` comments in jank.
2. This for loop is more idiomatically written like this:

```cpp
for(size_t i{}; i < 2; ++i)
```

In particular, a few things change:

1. We use `size_t` because we're counting up from 0
2. We use direct initialization, as we do in modern C++
3. We zero-base our indices/loop counters
4. We name the counter `i`, even if it's ignored
5. We use `++i` instead of `i++`; while there's no practical difference here, the postfix increment has serious performance implications in other scenarios and we want to only use it when we actually want postfix; here, we actually want prefix increment

Samy-33: Could you point me to a resource for why `i++` would have perf implications?

jeaye: The `++` operator in C++ is overloadable. We do this regularly for custom iterator types. For more complex iterator types, copying the iterator may be very slow. `i++` requires copying `i` so it can be returned and then incrementing the actual `i`. So, imagine an iterator which holds a vector and an index in it. By doing `i++`, you need to copy the iterator, which means deep copying the entire vector. If you use `++i`, the iterator is updated in place.

As I said, in this situation, the compiler will treat `i++` to mean `++i` because clang knows that people like to write `i++` when they mean `++i`. But C++ is all about building good habits so that you don't get bitten. By using `++i` as a default, and only `i++` when you actually want a copy of the value *before* it was incremented, you won't be surprised down the road.

jeaye: Since you asked for a reference:

* https://stackoverflow.com/questions/5223950/stl-iterators-prefix-increment-faster
* https://stackoverflow.com/questions/41765835/why-use-the-prefix-increment-form-for-iterators

jeaye:
Nice work, Saket!

PR #109, opened 2024-10-12
feat(real): lex scientific notation

Samy-33:


jeaye:
This looks great. I think we need a more rigorous testing suite for it, though. There are paths in the code which aren't currently being tested, like `1ee1`, `1Efoo`, etc. As I mentioned in the ticket, the main work for this task is going to be the negative test suite. Let's try to permutate all of the ways this could go wrong.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -377,30 +380,66 @@ namespace jank::read
...
                   return err(error{ token_start, pos, "invalid number" });
                 }
                 contains_dot = true;
               }
+              else if(c == 'e' || c == 'E')
+              {
+                if(is_scientific || !contains_leading_digit)
+                {
+                  ++pos;
+                  return err(error{ token_start, pos, "invalid number" });
```
jeaye, line 395: What's a better error we can return here?

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -377,30 +380,66 @@ namespace jank::read
...
+                }
+                is_scientific = true;
+                expecting_exponent = true;
+              }
+              else if(c == '+' || c == '-')
+              {
+                if(found_exponent_sign || !is_scientific || !expecting_exponent)
+                {
+                  ++pos;
+                  return err(error{ token_start, pos, "invalid number" });
```
jeaye, line 405: What's a better error we can return here?

PR #110, opened 2024-10-15
feat(chars): support unicode and octal char literals

Samy-33:
image

### DIFF
--- a/compiler+runtime/include/cpp/jank/read/parse.hpp
+++ b/compiler+runtime/include/cpp/jank/read/parse.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <codecvt>
+
 #include <jank/result.hpp>
 #include <jank/option.hpp>
 #include <jank/read/lex.hpp>
@@ -12,33 +14,75 @@ namespace jank::runtime
 /* TODO: Rename file to processor. */
 namespace jank::read::parse
 {
-  static option<char> get_char_from_literal(native_persistent_string const &sv)
+  static option<native_persistent_string>
+  parse_character_in_base(native_persistent_string_view const &char_literal, int const base)
   {
-    if(sv.size() == 2)
+    try
     {
-      return sv[1];
+      size_t chars_processed{};
+      auto const codepoint(
+        std::stol(native_persistent_string{ char_literal.data() + 2, char_literal.size() - 2 },
+                  &chars_processed,
+                  base));
+
+      /* Some characters that weren't processed at all.
+       * An example would be when trying to parse `12a` in octal base,
+       * `stol` will only parse `12` and will ignore `a`, returning `2`
+       * in `chars_processed`.
+       * We return none in that case, to represent to the downstream callers
+       * the enability to parse the char literal.
+       *
+       * Refer: https://en.cppreference.com/w/cpp/string/basic_string/stol
+       */
+      if(chars_processed != char_literal.size() - 2)
+      {
+        return none;
+      }
+
+      std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> converter;
+      return native_persistent_string{ converter.to_bytes(codepoint) };
     }
-    else if(sv == R"(\newline)")
+    catch(std::range_error const)
+    {
+      return none;
+    }
+    catch(std::invalid_argument const)
+    {
+      return none;
+    }
+    catch(std::out_of_range const)
+    {
+      return none;
+    }
+  }
+
+  static option<char> get_char_from_literal(native_persistent_string const &s)
+  {
+    if(s.size() == 2)
+    {
+      return s[1];
+    }
+    else if(s == R"(\newline)")
     {
       return '\n';
     }
-    else if(sv == R"(\space)")
+    else if(s == R"(\space)")
     {
       return ' ';
     }
-    else if(sv == R"(\tab)")
+    else if(s == R"(\tab)")
     {
       return '\t';
     }
-    else if(sv == R"(\backspace)")
+    else if(s == R"(\backspace)")
     {
       return '\b';
     }
-    else if(sv == R"(\formfeed)")
+    else if(s == R"(\formfeed)")
     {
       return '\f';
     }
-    else if(sv == R"(\return)")
+    else if(s == R"(\return)")
     {
       return '\r';
     }
--- a/compiler+runtime/include/cpp/jank/read/parse.hpp
+++ b/compiler+runtime/include/cpp/jank/read/parse.hpp
@@ -14,81 +14,10 @@ namespace jank::runtime
 /* TODO: Rename file to processor. */
 namespace jank::read::parse
 {
-  static option<native_persistent_string>
-  parse_character_in_base(native_persistent_string_view const &char_literal, int const base)
-  {
-    try
-    {
-      size_t chars_processed{};
-      auto const codepoint(
-        std::stol(native_persistent_string{ char_literal.data() + 2, char_literal.size() - 2 },
-                  &chars_processed,
-                  base));
+  result<option<native_persistent_string>, native_persistent_string>
+  parse_character_in_base(native_persistent_string const &char_literal, int const base);
 
-      /* Some characters that weren't processed at all.
-       * An example would be when trying to parse `12a` in octal base,
-       * `stol` will only parse `12` and will ignore `a`, returning `2`
-       * in `chars_processed`.
-       * We return none in that case, to represent to the downstream callers
-       * the enability to parse the char literal.
-       *
-       * Refer: https://en.cppreference.com/w/cpp/string/basic_string/stol
-       */
-      if(chars_processed != char_literal.size() - 2)
-      {
-        return none;
-      }
-
-      std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> converter;
-      return native_persistent_string{ converter.to_bytes(codepoint) };
-    }
-    catch(std::range_error const)
-    {
-      return none;
-    }
-    catch(std::invalid_argument const)
-    {
-      return none;
-    }
-    catch(std::out_of_range const)
-    {
-      return none;
-    }
-  }
-
-  static option<char> get_char_from_literal(native_persistent_string const &s)
-  {
-    if(s.size() == 2)
-    {
-      return s[1];
-    }
-    else if(s == R"(\newline)")
-    {
-      return '\n';
-    }
-    else if(s == R"(\space)")
-    {
-      return ' ';
-    }
-    else if(s == R"(\tab)")
-    {
-      return '\t';
-    }
-    else if(s == R"(\backspace)")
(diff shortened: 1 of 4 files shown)


compiler+runtime/include/cpp/jank/read/parse.hpp
```diff
@@ -12,33 +14,75 @@ namespace jank::runtime
...
+      return none;
+    }
+    catch(std::invalid_argument const)
+    {
+      return none;
+    }
+    catch(std::out_of_range const)
+    {
+      return none;
+    }
```
jeaye, line 56: Just one `catch(std::exception const &)` will do. Always catch exceptions by reference.

compiler+runtime/include/cpp/jank/read/parse.hpp
```diff
@@ -12,33 +14,75 @@ namespace jank::runtime
...
+        std::stol(native_persistent_string{ char_literal.data() + 2, char_literal.size() - 2 },
+                  &chars_processed,
+                  base));
+
+      /* Some characters that weren't processed at all.
+       * An example would be when trying to parse `12a` in octal base,
+       * `stol` will only parse `12` and will ignore `a`, returning `2`
+       * in `chars_processed`.
+       * We return none in that case, to represent to the downstream callers
+       * the enability to parse the char literal.
```
jeaye, line 33: Please update this comment to convey that we're checking for this case, rather than this being the case. When I read "Some characters that weren't processed at all", I think there's a problem. Perhaps: "We check for characters that were unable to be processed."

Also: `s/enability/inability/`

Overall, I like this helpful comment, though. :+1:

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -348,6 +348,41 @@ namespace jank::read::parse
 
     if(character.is_none())
     {
+      /* Unicode */
```
jeaye, line 351: This is hexadecimal Unicode. The other branch is octal Unicode.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -348,6 +348,41 @@ namespace jank::read::parse
...
+      if(sv[0] == '\\' && sv[1] == 'u')
+      {
+        /* Should always be of length 6 eg. \uaBc5 */
```
jeaye, line 354: This is how Clojure handles it, but I don't think we should have this limitation. For example, `\u1f920` is needed for 🤠. Same goes for octal. We should read however much is there, I think, as long as it's one codepoint.

compiler+runtime/include/cpp/jank/read/parse.hpp
```diff
@@ -12,33 +14,75 @@ namespace jank::runtime
...
+       *
+       * Refer: https://en.cppreference.com/w/cpp/string/basic_string/stol
+       */
+      if(chars_processed != char_literal.size() - 2)
+      {
+        return none;
+      }
+
+      std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> converter;
```
jeaye, line 42: The `converter` has a member fn on it to check how many codepoints were consumed. We could check that this is `1` and fail otherwise. Also, we may be getting into an area here where `result` is more useful than `option`.

compiler+runtime/include/cpp/jank/read/parse.hpp
```diff
@@ -12,33 +14,75 @@ namespace jank::runtime
 /* TODO: Rename file to processor. */
 namespace jank::read::parse
 {
-  static option<char> get_char_from_literal(native_persistent_string const &sv)
+  static option<native_persistent_string>
+  parse_character_in_base(native_persistent_string_view const &char_literal, int const base)
```
jeaye, line 18: I think that it's odd to keep the `\o` or `\u` prefix on this when parsing. We should strip it out using `substr` before calling.

compiler+runtime/include/cpp/jank/read/parse.hpp
Samy-33: `catch(std::exception const &)` will catch all exceptions. Even the ones thrown not because of wrong syntax.

Why is that better in that case?

compiler+runtime/include/cpp/jank/read/parse.hpp
Samy-33: Is it because we also want to return the cause of the error?

If so, what's the purpose of returning cause of the error? Is it to show a better error message?

jeaye: Yes and yes!

compiler+runtime/include/cpp/jank/read/parse.hpp
jeaye: Let's be not use a string_view here, though. I'm thinking of removing them from the supported jank types entirely, since they're altogether too difficult to use. For example, if we kept the string_view as the arg and we changed our call sites to use `substr`, we'd have undefined behavior since `substr` returns a temporary string, which we then borrow with the string_view and that doesn't keep it alive. Much better to just use a ref to const persistent string.

compiler+runtime/include/cpp/jank/read/parse.hpp
jeaye: We ant all exceptions caught here, since we're encoding error handling in the return type. Having an `option` or a `result` as a return type AND having to deal with exceptions is the worst of both worlds, in terms of API design.

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -244,10 +246,10 @@ namespace jank::read::parse
 
         SUBCASE("Invalid length")
         {
-          lex::processor lp{ R"(\o1 \o23 \o4567 \o0123454)" };
+          lex::processor lp{ R"(\o12345677 \o23007673323)" };
```
jeaye, line 249: What makes these invalid length?

compiler+runtime/include/cpp/jank/read/parse.hpp
```diff
@@ -12,33 +14,74 @@ namespace jank::runtime
 /* TODO: Rename file to processor. */
 namespace jank::read::parse
 {
-  static option<char> get_char_from_literal(native_persistent_string const &sv)
+  static result<option<native_persistent_string>, native_persistent_string>
```
jeaye, line 17: Both of these fns should be defined in the `.cpp` file, not the `.hpp` file. They're not templates. By having them be `static`, we're actually putting a separate copy of this fn into every single translation unit. We'll have over 80 copies of this fn in the jank program this way; they'll all be identical.

compiler+runtime/test/cpp/jank/read/parse.cpp
jeaye: Fair enough.

compiler+runtime/include/cpp/jank/read/parse.hpp
jeaye: Based on your questions, this needs to be explained in a few parts.

## Translation units
A translation unit (TU), in C++, is everything which compiles down to an object file (`.o`). Generally, this ties 1:1 with `.cpp` files. Each TU is compiled separately, based on the code within the `.cpp` file and all the headers included. If multiple TUs include the same header, they do all the same duplicate work to parse it.

## Linkage
Within a TU, symbols can have different types of linkage. This is what makes the symbols within the TU available to other TUs when they're linked together into a program. By default, symbols have external linkage, meaning the symbols can be accessed by other TUs when everything is linked. Furthermore, with external linkage, there may be only one definition per symbol. This is the one definition rule (ODR).

However, putting `static` on a free fn or var gives it internal linkage, which allows multiple definitions with the same name, since they don't compete in the global symbol space. Many TUs may a definition of a symbol, even differing definitions, as long as that symbol has internal linkage.

## Inline
There is another way to get around the ODR, for functions, which is to use `inline`. A function with external linkage can be marked `inline` and it will still be allowed to have multiple definitions, though still only one per TU. This allows multiple TUs to include a header which defines an inline fn without linker errors.

Function templates are implicitly inline and are generally defined in headers to be included into many TUs.

## Static
Note that the `static` keyword, in C++, has a few different meanings. Only one of them implies internal linkage. All of the meanings are noted here: https://en.cppreference.com/w/cpp/keyword/static

## Wrap
So, by having your function defined in a header and marked `static`, every TU which includes it will get its own copy, with internal linkage. This is a recipe for code bloat in the resulting binary. We have two options:

1. Mark the function `inline`, rather than `static`
2. Define the fn in a single TU and make sure it has external linkage (no `static`) and then just forward declare it in a header which is included by other TUs

Which one to choose? Well, we don't want the first one, since defining too many inline fns in headers slows down compilation. Remember, every TU which includes that header needs to process it. This takes time. In this case, it's in a common header, too. So our choice is number 2.

How does that work? Well, every TU which wants to call this fn, but which doesn't define it, will reference the symbol, but it will remain unresolved in that TU's generated `.o` file. This is ok. Once all of the `.o` files are linked together into our `jank` binary, all symbols are resolved. As long as we link in the TU which has that fn, and the fn has external linkage, we're golden. If not, we'll get linker errors.

You'll see this is the pattern that we use already in jank. For example:

* Declaration: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/include/cpp/jank/runtime/core/seq.hpp#L172
* Definition: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/runtime/core/seq.cpp#L14

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -15,6 +15,81 @@ namespace jank::read::parse
 {
   using namespace jank::runtime;
 
+  result<option<native_persistent_string>, native_persistent_string>
```
jeaye, line 18: `result<option<T>, E>` implies that `ok(none)` is expected, but we're never returning that. Why do we have the `option<T>` in there at all?

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -15,6 +15,81 @@ namespace jank::read::parse
...
+       * Refer: https://en.cppreference.com/w/cpp/string/basic_string/stol
+       */
+      if(chars_processed != char_literal.size())
+      {
+        return err("Invalid unicode digit");
+      }
+
+      std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> converter;
+
+      auto const converted(native_persistent_string{ converter.to_bytes(codepoint) });
```
jeaye, line 44: Here we're using `auto` and the `type`. If we need to specify the type, just specify the type.

```cpp
native_persistent_string const converted{ converter.to_bytes(codepoint) };
```

jeaye:
Great work!

PR #111, opened 2024-10-16
Compile and run on Mac OS 15.1 (Sequoia)

hadils:
### Summary
Compile and test jank on a MacOS system

### Context
jank would not compile on Mac OS 15.1 (Sequoia).

### Changes Made
* Moved /opt/homebrew/include as the last preprocessor dependency
*  Built libunwind.a and libc++abi.a in llvm-install subdirectory
* Added libraries to link dependencies

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ CMakeUserPresets.json
 
 # Rust
 rust/jank
+/compiler+runtime/llvm-build/
+/compiler+runtime/llvm-install/
+/compiler+runtime/llvm/
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -62,7 +62,6 @@ endif()
 
 set(
   jank_aot_compiler_flags
-  $<$<PLATFORM_ID:Darwin>:-I/opt/homebrew/include>
   -Wall -Wextra -Wpedantic
   -Wfloat-equal -Wuninitialized -Wswitch-enum -Wnon-virtual-dtor
   -Wold-style-cast -Wno-gnu-case-range
@@ -82,6 +81,7 @@ set(
   -w
 )
 
+
 # TODO: Remove FMT_HEADER_ONLY once the extern template Clang bug is fixed.
 set(
   jank_common_compiler_flags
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -40,3 +40,6 @@ CMakeUserPresets.json
 
 # Rust
 rust/jank
+/compiler+runtime/llvm-build/
+/compiler+runtime/llvm-install/
+/compiler+runtime/llvm/
```
jeaye, line 45: These dirs should be in `build`. I suspect that you didn't `cd build` before running `./bin/build-clang`, which is what lead to this. However, the entire `build` dir is currently ignored, so there should be no change needed here.

compiler+runtime/CMakeLists.txt
```diff
@@ -98,6 +98,8 @@ set(
   jank_linker_flags
   $<$<PLATFORM_ID:Darwin>:-L/opt/homebrew/lib>
   #-stdlib=libc++ -lc++abi
+  ${CMAKE_SOURCE_DIR}/llvm-install/usr/local/lib/libunwind.a
+  ${CMAKE_SOURCE_DIR}/llvm-install/usr/local/lib/libc++abi.a
```
jeaye, line 102: Since we're including the libs in `jank_lib` below, we shouldn't also need them here. Will you please confirm that?

compiler+runtime/bin/build-clang
```diff
@@ -40,31 +40,29 @@ srcdir="${PWD}"
...
-        -DLLVM_TARGETS_TO_BUILD="host" \
-        -DLLVM_ENABLE_EH=OFF \
-        -DLLVM_ENABLE_RTTI=OFF \
-        -DLLVM_INCLUDE_BENCHMARKS=OFF \
-        -DLLVM_INCLUDE_EXAMPLES=OFF \
-        -DLLVM_INCLUDE_TESTS=OFF \
-        -DLLVM_ENABLE_ZSTD=OFF \
-        -G "Unix Makefiles" \
-        ../llvm/llvm
+    -DLLVM_ENABLE_RUNTIMES=all \
```
jeaye, line 54: Your change only affected this one line. Please don't re-format the whole file along with that one line change.

compiler+runtime/CMakeLists.txt
```diff
@@ -247,6 +249,8 @@ target_link_libraries(
   clangInterpreter
   Boost::boost
   $<$<PLATFORM_ID:Darwin>:crypto>
+  ${CMAKE_SOURCE_DIR}/llvm-install/usr/local/lib/libunwind.a
+  ${CMAKE_SOURCE_DIR}/llvm-install/usr/local/lib/libc++abi.a
```
jeaye, line 253: This should be based on the `CMAKE_BINARY_DIR`, since clang/llvm should've been built within the `build` dir (according to the docs).

jeaye:
Closing this, since it lead to #112. Thanks so much for sending this in! It was very helpful.

PR #116, opened 2024-10-21
Add support for ratio

jianlingzhong:
Add support for ratio

Currently, ratio (of two integers) will be converted to a real number during parsing.

Lexing and parsing should work:

```
❯ build/jank repl   
Bottom of clojure.core
> 4/5
0.8
> 4/5/6
Read error (0 - 3): invalid number
> 4//5
Read error (0 - 4): invalid number
> 4/5/6/7
Read error (0 - 3): invalid number
> 4.4/5
Read error (0 - 3): invalid number
> 5/5.5
Read error (0 - 5): invalid number
> 5/7
0.7142857142857143
> (+ 4 4/5)
4.8
> (+ 4 -4/5)
3.2
> (+ 4 4/-5)
3.2
> (+ 4 -4/-5)
4.8
> (+ -4/5 -4/-5)
0
```


compiler+runtime/include/cpp/jank/read/lex.hpp
```diff
@@ -142,6 +153,7 @@ namespace jank::read::lex
     size_t pos{};
     /* Whether or not the previous token requires a space after it. */
     native_bool require_space{};
+    native_bool found_slash_after_number{}; // True if seeing a '/' following a number.
```
jeaye, line 156: Please put comments above the line and use `/* */` style comments, like the comment on line 154.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -154,6 +154,22 @@ namespace jank::read
     {
     }
 
+    token::token(size_t const p, size_t const s, token_kind const k, ratio d)
```
jeaye, line 157: Please make `d` const, if you're not mutating it.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -154,6 +154,22 @@ namespace jank::read
...
+      : pos{ p }
+      , size{ s }
+      , kind{ k }
+      , data{ d }
+    {
+    }
+
+    native_bool ratio::operator==(ratio const &rhs) const {
+      return this->numerator == rhs.numerator && this->denominator == rhs.denominator;
```
jeaye, line 166: Note that `this->` is unnecessary here.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -266,7 +287,8 @@ namespace jank::read
       }
 
       auto const token_start(pos);
-      switch(file[token_start])
+      auto a = file[token_start];
```
jeaye, line 290: 1. Please use direct initialization. i.e. `auto a(...);`
2. Please make this const, if it's not being mutated
3. I'm not sure what `a` stands for; if it's just a lazy name, please it something more clear like `current_char`
4. I don't actually see this local being used anywhere else; is it needed?

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1166,6 +1168,15 @@ namespace jank::read::parse
                                token };
   }
 
+  processor::object_result processor::parse_ratio() {
+    auto const token(token_current->expect_ok());
+    ++token_current;
+    const lex::ratio ratio_data = boost::get<lex::ratio>(token.data);
```
jeaye, line 1174: Please use `auto const &ratio_data(boost::get<lex::ratio>(token.data));`.

1. Uses direct initialization
2. Uses a ref to const, to avoid copying
3. Uses auto, since the type is clear from the right hand side

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -377,30 +402,88 @@ namespace jank::read
...
+              else if(c == '/') {
+                require_space = false;
+                ++pos;
+                if(found_exponent_sign || is_scientific || expecting_exponent || contains_dot
+                  || found_slash_after_number) {
+                  return err(error{ token_start, pos, "invalid number" });
+                }
+                found_slash_after_number = true;
+                ++pos; // skip the '/' char and look for the denominator number.
+                const result<token, error> denominator = next();
```
jeaye, line 440: Please use `auto const denominator(next());` to use direct initialization and avoid spelling out an obvious type.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -377,30 +402,88 @@ namespace jank::read
...
+                  const token& denominator_token = denominator.expect_ok();
+                  found_slash_after_number = false;
+                  return ok(token(token_start,
+                                 pos - token_start,
+                                 token_kind::ratio, {
+                                   .numerator = std::strtoll(file.data() + token_start, nullptr, 10),
+                                   .denominator = boost::get<native_integer>(denominator_token.data)}));
+
+                }
+                return err(error{ token_start, pos, "invalid number" });
```
jeaye, line 451: What's a more helpful error message we can put here? We know that we're expecting an integer denominator and we didn't get one.

compiler+runtime/include/cpp/jank/read/lex.hpp
```diff
@@ -47,6 +47,15 @@ namespace jank::read::lex
     /* Has string data. */
     escaped_string,
     eof,
+    /* Has two integer data. */
+    ratio,
```
jeaye, line 51: Please put this *before* `eof`; the intention is that it's the last one. Would seem odd to have it in the middle of our token types. Since it's numeric, I think it makes sense to go right after `real`.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -377,30 +402,88 @@ namespace jank::read
...
                   return err(error{ token_start, pos, "invalid number" });
                 }
                 contains_dot = true;
               }
+              else if(c == 'e' || c == 'E')
+              {
+                if(is_scientific || !contains_leading_digit)
+                {
+                  ++pos;
+                  return err(error{ token_start, pos, "invalid number" });
```
jeaye, line 417: What's a better error message we can surface here?

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -377,30 +402,88 @@ namespace jank::read
...
+                }
+                is_scientific = true;
+                expecting_exponent = true;
+              }
+              else if(c == '+' || c == '-')
+              {
+                if(found_exponent_sign || !is_scientific || !expecting_exponent)
+                {
+                  ++pos;
+                  return err(error{ token_start, pos, "invalid number" });
```
jeaye, line 427: What's a better error message we can surface here?

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -377,30 +402,88 @@ namespace jank::read
...
+                }
+                found_exponent_sign = true;
+              }
+              else if(c == '/') {
+                require_space = false;
+                ++pos;
+                if(found_exponent_sign || is_scientific || expecting_exponent || contains_dot
+                  || found_slash_after_number) {
+                  return err(error{ token_start, pos, "invalid number" });
```
jeaye, line 436: What's a better error message we can surface here?

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -374,7 +374,54 @@ namespace jank::read::lex
...
+      SUBCASE("Arithmetic") {
+        processor p{ "(+ 4/5 5)" };
+        native_vector<result<token, error>> tokens(p.begin(), p.end());
+        CHECK(tokens == make_tokens({
+          {token(0, 1, token_kind::open_paren)},
+          {token(1, 1, token_kind::symbol, "+")},
+          {token(3, 3, token_kind::ratio, {.numerator = 4, .denominator = 5})},
+          {token(7, 1, token_kind::integer, 5ll)},
+          {token(8, 1, token_kind::close_paren)}}));
+     }
```
jeaye, line 423: No need for this whole subcase. That's not what we're testing here.

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -63,6 +63,23 @@ namespace jank::read::parse
       CHECK(r.expect_ok().unwrap().end == r.expect_ok().unwrap().start);
     }
 
+    TEST_CASE("Ratio") {
+      SUBCASE("Single Ratio") {
```
jeaye, line 67: Let's add a subcase for `1/0`, which should fail. I just tested this in a Clojure repl and it fails. We should detect this while parsing, not while lexing.

```
user=> 1/0
Syntax error reading source at (REPL:27:0).
Divide by zero
```

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: sorry this is for debugging purpose. forgot to remove it. it's removed now.

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: changed it to 'invalid ratio'. it could be more informative by separating all the if cases, if needed.

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: i changed it to "invalid ratio: expecting an integer denominator".

PR #117, opened 2024-10-21
Add support for ratio

jianlingzhong:
Add support for ratio

Currently, ratio (of two integers) will be converted to a real number during parsing.

Lexing and parsing should work:

```
❯ build/jank repl   
Bottom of clojure.core
> 4/5
0.8
> 4/5/6
Read error (0 - 3): invalid number
> 4//5
Read error (0 - 4): invalid number
> 4/5/6/7
Read error (0 - 3): invalid number
> 4.4/5
Read error (0 - 3): invalid number
> 5/5.5
Read error (0 - 5): invalid number
> 5/7
0.7142857142857143
> (+ 4 4/5)
4.8
> (+ 4 -4/5)
3.2
> (+ 4 4/-5)
3.2
> (+ 4 -4/-5)
4.8
> (+ -4/5 -4/-5)
0
> 1/0
Read error (0 - 0): Divide by zero
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -154,6 +154,24 @@ namespace jank::read
     {
     }
 
+    token::token(size_t const p, size_t const s, token_kind const k, ratio const d)
+      : pos{ p }
+      , size{ s }
+      , kind{ k }
+      , data{ d }
+    {
+    }
+
+    native_bool ratio::operator==(ratio const &rhs) const
+    {
+      return numerator == rhs.numerator && denominator == rhs.denominator;
+    }
+
+    native_bool ratio::operator!=(ratio const &rhs) const
+    {
+      return !(*this == rhs);
+    }
+
     native_bool token::no_data::operator==(no_data const &) const
     {
       return true;
@@ -184,6 +202,11 @@ namespace jank::read
     std::ostream &operator<<(std::ostream &os, token::no_data const &)
     {
       return os << "<no data>";
+    }
+
+    std::ostream &operator<<(std::ostream &os, ratio const &r)
+    {
+      return os << r.numerator << "/" << r.denominator;
     }
 
     processor::processor(native_persistent_string_view const &f)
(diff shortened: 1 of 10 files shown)


compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -406,6 +427,30 @@ namespace jank::read
...
+                if(found_exponent_sign || is_scientific || expecting_exponent || contains_dot
+                  || found_slash_after_number) {
+                  return err(error{ token_start, pos, "invalid ratio" });
+                }
+                found_slash_after_number = true;
+                /* skip the '/' char and look for the denominator number. */
+                ++pos;
+                auto const denominator(next());
+                if (denominator.is_ok() && denominator.expect_ok().kind == token_kind::integer) {
+                  const token& denominator_token = denominator.expect_ok();
```
jeaye, line 443: Please use `auto const &denominator_token(denominator.expect_ok());`

1. Prefer auto
2. Prefer direct initialization
3. const always goes to the right

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -374,7 +374,61 @@ namespace jank::read::lex
         }));
       }
     }
-
+    TEST_CASE("Ratio")
+    {
+      SUBCASE("Successes")
```
jeaye, line 379: Please split this subcase up into named subcases, like all of our other tests. This allows you to avoid the `p`, `p1`, `p2` problem, too.

jianlingzhong: splitted

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: done

compiler+runtime/test/cpp/jank/read/lex.cpp
jianlingzhong: split the test for the "success" cases as well.

PR #123, opened 2024-11-03
Add support for delay

jianlingzhong:
Still a draft. Needs more work:

```
❯ build/jank repl
Bottom of clojure.core
> (def my-delay (delay (println "did some work")))
did some work
#'clojure.core/my-delay
> (force my-delay)
Unknown exception thrown
```

```
❯ build/jank repl
Bottom of clojure.core
> (def my-delay (delay (+ 4 5)))
#'clojure.core/my-delay
> (println my-delay)
delay@0x1170c0d80
nil
> (force my-delay)
Unknown exception thrown
```

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/obj/delay.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/delay.hpp
@@ -23,8 +23,9 @@ namespace jank::runtime
     static object_ptr force(object_ptr const &d);
 
     object base{ object_type::delay };
-    object_ptr val{};
+    mutable object_ptr val{};
     object_ptr fn{};
+    mutable std::atomic<object_ptr> error{};
     std::unique_ptr<std::exception_ptr> delay_exception_ptr
       = std::make_unique<std::exception_ptr>(nullptr);
   };
--- a/compiler+runtime/include/cpp/jank/runtime/obj/delay.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/delay.hpp
@@ -18,16 +18,13 @@ namespace jank::runtime
     native_hash to_hash() const;
 
     /* behavior::derefable */
-    object_ptr deref() const;
-
-    static object_ptr force(object_ptr const &d);
+    object_ptr deref();
 
     object base{ object_type::delay };
-    mutable object_ptr val{};
+    object_ptr val{};
     object_ptr fn{};
-    mutable std::atomic<object_ptr> error{};
-    std::unique_ptr<std::exception_ptr> delay_exception_ptr
-      = std::make_unique<std::exception_ptr>(nullptr);
+    object_ptr error{};
+    std::mutex mutex;
   };
 
   namespace obj
--- a/compiler+runtime/include/cpp/jank/runtime/obj/delay.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/delay.hpp
@@ -1,0 +1,37 @@
+#pragma once
+
+namespace jank::runtime
+{
+  template <>
+  struct static_object<object_type::delay> : gc
+  {
+    static constexpr native_bool pointer_free{ false };
+
+    static_object() = default;
+    static_object(object_ptr fn);
+
+    /* behavior::object_like */
+    native_bool equal(object const &) const;
+    native_persistent_string to_string() const;
+    void to_string(fmt::memory_buffer &buff) const;
+    native_persistent_string to_code_string() const;
+    native_hash to_hash() const;
+
+    /* behavior::derefable */
+    object_ptr deref() const;
+
+    static object_ptr force(object_ptr const &d);
+
+    object base{ object_type::delay };
+    object_ptr val{};
+    object_ptr fn{};
+    std::unique_ptr<std::exception_ptr> delay_exception_ptr
+      = std::make_unique<std::exception_ptr>(nullptr);
+  };
+
+  namespace obj
+  {
+    using delay = static_object<object_type::delay>;
+    using delay_ptr = native_box<delay>;
+  }
+}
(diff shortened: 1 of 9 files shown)


compiler+runtime/src/cpp/jank/runtime/obj/delay.cpp
```diff
@@ -0,0 +1,103 @@
+#include <jank/runtime/obj/delay.hpp>
+
+namespace jank::runtime
+{
+  obj::delay::static_object(object_ptr const fn)
+    : fn{ fn }
+  {
+    assert(fn);
+    val = nullptr;
```
jeaye, line 9: No need for this; it's zero-initialized by the `{}` in the struct declaration.

compiler+runtime/src/cpp/jank/runtime/obj/delay.cpp
```diff
@@ -0,0 +1,103 @@
...
+    to_string(buff);
+    return native_persistent_string{ reinterpret_cast<native_persistent_string::size_type>(
+                                       buff.data()),
+                                     static_cast<char>(buff.size()) };
+  }
+
+  object_ptr obj::delay::force(object_ptr const &d)
+  {
+    return visit_object(
+      [&]<typename T>(T const typed_s) -> object_ptr {
```
jeaye, line 29: Rather than `<typename T>` here, with a param of type `T`, we need to use a different pattern. Let's make the param type `auto`, and then define this type within the lambda: `using T = typename decltype(typed_d)::value_type;`. Also, the convention is to name the typed object after the original object. The original object is `d`, so the param should be `typed_d`. Finally, your lambda is capturing state with the `[&]`, which results in a closure being generated, not a free function. Since you're not actually capturing anything, we can be explicit about that and get better code generation from it.

The reason why we need to do the `using T =` is that the type is going to be a `native_box<>` of our object type. So your `T` will be a `native_box<obj::delay>`. This is a problem, since that type isn't `behavior::derefable`. The resulting code should look like this:

```cpp
return visit_object(
  [](auto const typed_d) -> object_ptr {
    using T = typename decltype(typed_d)::value_type;

    if constexpr(behavior::derefable<T>)
    {
      return typed_d->deref();
    }
    else
    {
      return d;
    }
  },
  d);
```

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -206,9 +206,9 @@ namespace jank::read::parse
               { "\\u1234", "\\u5678", "\\u90ab", "\\ucdef", "\\uABCD", "\\uEFa0" })
           {
             auto const r(p.next());
```
jeaye, line 208: Looks like there are no actual changes to this file. Just reformatting. Please remove this file from the diff.

compiler+runtime/src/cpp/jank/runtime/obj/delay.cpp
```diff
@@ -0,0 +1,103 @@
...
+          try
+          {
+            throw std::runtime_error{ fmt::format("Invalid expression provided: {}",
+                                                  typed_s->to_string()) };
+          }
+          catch(...)
+          {
+            *delay_exception_ptr = std::current_exception();
+            std::rethrow_exception(*delay_exception_ptr);
+          }
```
jeaye, line 98: This is a good try to get things going. Ultimately, what we want should be simpler.

1. Store the exception as an `object_ptr`, like any other jank object
2. Call the function just using `dynamic_call`; no need to visit it
3. Catch `object_ptr` and `std::exception const &e`; store the former, turn the latter into an `object_ptr` with `make_box(e.what())`
4. If the exception object is non-null, just throw it as it is

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -834,6 +834,10 @@
   [x]
   (native/raw "__value = make_box(~{ x }->type == object_type::volatile_);"))
 
+; Delay and force
+(defn delay [exp] (native/raw "__value = make_box<obj::delay>(~{ exp });"))
+(defn force [exp] (native/raw "__value = try_object<obj::delay>(~{ exp })->deref();"))
```
jeaye, line 839: Good work getting something going here.

1. The comment doesn't provide anything useful; let's remove it
2. We'll need to rewrite delay to be a macro, in order for any of this to work properly; we'll talk about that on Monday
3. The `force` implementation looks good, with two stylistic requests
  a. We don't use `exp` for param names anywhere else; for general objects, we use `o`, but you could use `d` if you're expecting a delay
  b. We always put fn bodies on a new line, not on the same line as the args

compiler+runtime/include/cpp/jank/runtime/obj/delay.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace jank::runtime
+{
+  template <>
+  struct static_object<object_type::delay> : gc
+  {
+    static constexpr native_bool pointer_free{ false };
+
+    static_object() = default;
+    static_object(object_ptr o);
```
jeaye, line 11: CI will complain that the arg name `o` in the declaration doesn't match the arg name `fn` in the definition. We want to update the declaration, since `fn` is much more helpful.

jeaye:
The changes look great, @jianlingzhong! Please format all of the C++ with clang-format from the `build/llvm-install` dir before we merge.

jianlingzhong:
Things should work now:

```
❯ build/jank repl
Bottom of clojure.core
> (def d (delay (println "test")))
#'clojure.core/d
> d
delay@0x1173924d0
> (deref d)
test
nil
> (deref d)
nil
> (def d (delay (println "test")))
#'clojure.core/d
> (force d)
test
nil
> (force d)
nil
> (force 42)
42
> 
```

Thanks @jeaye !

jianlingzhong:
> The changes look great, @jianlingzhong! Please format all of the C++ with clang-format from the `build/llvm-install` dir before we merge.

Done!

jeaye:
Nice job!

PR #125, opened 2024-11-04
feat(repl): adds persistent history with initial multiline support

Samy-33:
[asciicast](https://asciinema.org/a/BrSiSORlkaHHgye98JhTW1fQm)

### DIFF
--- a/compiler+runtime/src/cpp/main.cpp
+++ b/compiler+runtime/src/cpp/main.cpp
@@ -1,5 +1,4 @@
 #include <iostream>
-#include <filesystem>
 
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
@@ -8,9 +7,7 @@
 #include <llvm/Support/CommandLine.h>
 #include <llvm/Support/ManagedStatic.h>
 #include <llvm/Support/TargetSelect.h>
-
-#include <readline/readline.h>
-#include <readline/history.h>
+#include <llvm/LineEditor/LineEditor.h>
 
 #include <folly/FBString.h>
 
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/cpp/main.cpp
```diff
@@ -184,24 +197,32 @@ namespace jank
...
     }
 
-    /* By default, RL will do tab completion for files. We disable that here. */
-    rl_bind_key('\t', rl_insert);
+    llvm::LineEditor le("jank-native", ".jank-native-repl-history");
+    le.setPrompt("native> ");
+    native_transient_string input{};
 
-    while(auto const buf = readline("native> "))
+    while(std::optional<native_persistent_string> buf = le.readLine())
```
jeaye, line 204: I'd still use `auto` here. If we want the type, we can look at the declaration of `le.readLine()`.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -124,27 +121,40 @@ namespace jank
...
       }
 
-      /* TODO: Persist history to disk, á la .lein-repl-history. */
-      /* History is persisted for this session only. */
-      add_history(line.data());
+      input += line;
+      if(input == "(exit)")
+      {
+        break;
+      }
```
jeaye, line 153: Why add this? Clojure doesn't do it. Is there any other precedent?

It prevents the user from defining a fn named exit and calling it. Normal `^d` does the trick.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -124,27 +121,40 @@ namespace jank
       dynamic_call(__rt_ctx->in_ns_var->deref(), make_box<obj::symbol>(opts.target_module));
     }
 
-    /* By default, RL will do tab completion for files. We disable that here. */
-    rl_bind_key('\t', rl_insert);
+    auto const get_prompt([&](native_persistent_string const &suffix) {
```
jeaye, line 124: No need for the `[&]` capture here, I think. `__rt_ctx` is global, `suffix` is a param.

compiler+runtime/src/cpp/main.cpp
Samy-33: I see. I've seen this being used in a REPL. I can't recall where.

But it makes sense to not include it as you mentioned.

compiler+runtime/src/cpp/main.cpp
Samy-33: `auto const &line(*buf);`

This defines a `line` which holds a reference to to the value in the `buf` optional. Is that how we prevent the copies?

Also, it throws an error apparently. No member function `erase` on the `line`

PR #128, opened 2024-11-11
Add native repeat object

jianlingzhong:


compiler+runtime/include/cpp/jank/runtime/obj/repeat.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <jank/runtime/behavior/seqable.hpp>
+
+namespace jank::runtime
+{
+  template <>
+  struct static_object<object_type::repeat> : gc
+  {
+    static constexpr native_bool pointer_free{ false };
+    static constexpr native_integer INFINITE{ -1 };
+    using bounds_check_t = native_bool (*)(object_ptr);
```
jeaye, line 12: This is unused, as far as I can tell.

jianlingzhong: Removed `bounds_check_t`, the other two are used.

compiler+runtime/include/cpp/jank/runtime/obj/repeat.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <jank/runtime/behavior/seqable.hpp>
+
+namespace jank::runtime
+{
+  template <>
+  struct static_object<object_type::repeat> : gc
+  {
+    static constexpr native_bool pointer_free{ false };
+    static constexpr native_integer INFINITE{ -1 };
```
jeaye, line 11: Let's lowercase this. I know the Clojure source had this, but jank doesn't use full caps for anything.

Aside from that, everything else looks good!

jianlingzhong: updated.

jeaye:
Nice work, @jianlingzhong!

PR #132, opened 2024-11-13
Remove ocaml bindings from the clang build and change ubuntu instructions

stmonty:
Removed the unnecessary ocaml bindings and ocamldoc building that hindered building if ocaml was installed.
Added gcc14 (for libstdc++) to build clang on Ubuntu.


jeaye:
Thanks!

PR #133, opened 2024-11-22
feat/support unicode identifiers

### LINKED ISSUE #89: Support Unicode identifiers (symbols/keywords)
Clojure defines symbols (keywords inherit the same behavior) to include the following:

> Symbols begin with a non-numeric character and can contain alphanumeric characters and *, +, !, -, _, ', ?, <, > and = (other characters may be allowed eventually).

Clojure treats "non-numeric character" and "alphanumeric characters" to include pretty much any Unicode character. This includes the various Unicode whitespace characters.

```clojure
; Supported for symbols.
user=> (def 😁 :smile)
#'user/😁

; Supported for keywords.
user=> (def fqkw :🙃/😁)
#'user/fqkw
user=> fqkw
:🙃/😁

; Even Unicode whitespace characters are supported.
user=> (def   2)
#'user/
user=>
2
```

stmonty:
Adds UTF-8 Unicode support (fixes #89).

It currently passes all the tests, though I made some modifications to some lex tests.
However, I think these changes better reflect the desired behavior matching the official Clojure implementation.

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -303,7 +303,7 @@ namespace jank::read
     static native_bool is_special_char(char32_t c)
     {
       return c == '(' || c == ')' || c == '{' || c == '}' || c == '[' || c == ']'
-        || c == ':' || c == '"' || c == '^' || c == '\\';
+        || c == '"' || c == '^' || c == '\\' || c == '`' || c == '~';
     }
     
     static native_bool is_symbol_char(char32_t const c)
@@ -376,7 +376,7 @@ namespace jank::read
           {
             require_space = false;
 
-            
+            //std::cout << "Character Token\n";
             auto const ch(peek());
             pos++;
 
@@ -402,6 +402,7 @@ namespace jank::read
               }
               if(pt.is_err() || !is_symbol_char(cpoint))
               {
+                //std::cout << "This isn't a char: '" << static_cast<char>(cpoint) << "'\n";
                 break;
               }
               pos += size;
@@ -409,7 +410,7 @@ namespace jank::read
 
             native_persistent_string_view const data{ file.data() + token_start,
                                                       ++pos - token_start };
-
+            //std::cout << "Character: " << data << " : " << data.size() << "\n";
             return ok(token{ token_start, pos - token_start, token_kind::character, data });
           }
         case ';':
@@ -697,7 +698,7 @@ namespace jank::read
             {
               return err(error{ token_start, "invalid keyword: starts with /" });
             }
-            else if(name[0] == ':' && name.size() == 1)
+            else if(name[0] == ':' && name[1] == ':')
             {
               return err(error{ token_start, "invalid keyword: incorrect number of :" });
             }
@@ -786,6 +787,7 @@ namespace jank::read
         /* Reader macros. */
         case '#':
           {
+            //std::cout << "Reader Macro Token\n";
             auto &&e(check_whitespace(found_space));
             if(e.is_some())
             {
@@ -797,15 +799,17 @@ namespace jank::read
             size_t size{};
             if (oc.is_err())
             {
+              //std::cout << "c now equals = ' '\n";
               c = ' ';
+              ++pos;
             }
             else
             {
               c = oc.expect_ok().character;
+              size = oc.expect_ok().len;
+              pos += size;
             }
-            size = oc.expect_ok().len;
-            pos += size;
-
+            //std::cout << "Reader char: " << static_cast<char>(c) << "\n";
             switch(c)
             {
               case '_':
@@ -814,19 +818,21 @@ namespace jank::read
                   token{ token_start, pos - token_start, token_kind::reader_macro_comment });
               case '?':
                 {
+                  //std::cout << "Question Mark case in Reader macro!\n";
                   auto const maybe_splice(peek());
                   char32_t c{};
                   size_t size{};
                   if (maybe_splice.is_err())
                   {
                     c = ' ';
+                    ++pos;
                   }
                   else
                   {
                     c = maybe_splice.expect_ok().character;
+                    size = maybe_splice.expect_ok().len;
+                    pos += size;
                   }
-                  size = maybe_splice.expect_ok().len;
-                  pos += size;
                   if(c == '@')
                   {
                     ++pos;
@@ -842,14 +848,16 @@ namespace jank::read
                   }
                 }
               default:
+                //std::cout << "Default case in Reader macro!\n";
                 break;
             }
-
+            //std::cout << "Reader Macro Pos: " << pos << " Token Start: " << token_start << "\n";
             return ok(token{ token_start, pos - token_start, token_kind::reader_macro });
           }
         /* Syntax quoting. */
         case '`':
           {
+            //std::cout << "Syntax quoting token\n";
             auto &&e(check_whitespace(found_space));
             if(e.is_some())
             {
@@ -863,6 +871,7 @@ namespace jank::read
         /* Syntax unquoting. */
         case '~':
           {
+            //std::cout << "Syntax unquoting token\n";
             auto &&e(check_whitespace(found_space));
             if(e.is_some())
             {
@@ -875,13 +884,14 @@ namespace jank::read
             if (oc.is_err())
             {
               c = ' ';
+              ++pos;
             }
             else
             {
               c = oc.expect_ok().character;
+              size = oc.expect_ok().len;
+              pos += size;
             }
-            size = oc.expect_ok().len;
-            pos += size;
 
             switch(c)
             {
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -316,7 +316,6 @@ namespace jank::read
 
     result<token, error> processor::next()
     {
-      //std::cout << "Starting Pos: " << pos << "\n";
       /* Skip whitespace. */
       native_bool found_space{};
       while(true)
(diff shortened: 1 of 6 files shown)


compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -162,6 +162,12 @@ namespace jank::read
     {
     }
 
+    struct codepoint
+    {
+      char32_t character{};
+      size_t len{};
```
jeaye, line 168: We can use a `uint8_t` here, since we know that the character is no longer than 32 bits (`char32_t`).

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -212,6 +218,7 @@ namespace jank::read
     processor::processor(native_persistent_string_view const &f)
       : file{ f }
     {
+      std::setlocale(LC_ALL, "en_US.utf8");
```
jeaye, line 221: I don't think it makes sense for us to hard-code the locale. This will break localization for non-English speakers, for example.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -260,11 +267,51 @@ namespace jank::read
       return none;
     }
 
-    static native_bool is_symbol_char(char const c)
+    static result<codepoint, error> convert_to_codepoint(native_persistent_string_view sv, size_t pos)
```
jeaye, line 270: Please make the params `const` if they're not being mutated.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -260,11 +267,51 @@ namespace jank::read
...
+    {
+      std::mbstate_t state{};
+      wchar_t wc{};
+      size_t len = std::mbrtowc(&wc, sv.data(), sv.size(), &state);
```
jeaye, line 274: Please use direct initialization and avoid spelling out types.

`auto const len{ std::mbrtowc(&wc, sv.data(), sv.size(), &state) }`;

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -289,7 +336,15 @@ namespace jank::read
       }
 
       auto const token_start(pos);
-      switch(file[token_start])
+      auto const oc = convert_to_codepoint(file.substr(token_start), token_start);
```
jeaye, line 339: Please always use direct initialization.

`auto const &oc{ convert_to_codepoint(file.substr(token_start), token_start) };`

compiler+runtime/src/cpp/jank/read/lex.cpp
stmonty: Just curious why not spell out types? I am not disagreeing, just want to understand the rationale.

jeaye: The compiler knows the type. We can find the type by looking at the expression. It's only due to historical reasons that the left hand side needed the type there at all.

Finally, it violates DRY by focusing on nominal typing. It means if we update the return type, even to something structurally compatible, we'll break compilation at every call site. Better to use `auto`.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -289,7 +336,15 @@ namespace jank::read
...
+      char32_t start{};
+      size_t len{};
```
jeaye, line 341: We have a struct for this! You end up doing `c` and `len` (in some form) **six** different times in this file. Just use our `codepoint` struct instead.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -289,7 +336,15 @@ namespace jank::read
...
+      if (oc.is_ok())
+      {
+        start = oc.expect_ok().character;
+        len = oc.expect_ok().len;
+      }
```
jeaye, line 346: What happens if it's not ok? We're checking the wrong thing here. There's a pattern in jank's code which you may not have noticed yet.

Basically every `if` check in jank's code is checking for the negative case. This is for two reasons:

1. It documents and asserts our invariants up front, allowing us to fail fast
2. It keeps the successful path *less nested*

For the second point, nesting code is considered a bad smell. If you find yourself checking for a positive if condition, pause, since you should probably swap it around.

In this case, the whole reason we returned an error from `convert_to_codepoint` was so that we can return it as an error here. With the code as it is, that error gets swallowed. There's a test we could write, which has an error with the starting char, which should then fail since that error isn't returned.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -559,54 +628,58 @@ namespace jank::read
...
               auto const oc(peek());
-              if(oc.is_none())
+              if(oc.is_err())
               {
                 break;
               }
-              auto const c(oc.unwrap());
+
+              auto const c = oc.expect_ok().character;
+              auto const size(oc.expect_ok().len);
```
jeaye, line 667: Just pull out the codepoint.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -260,11 +267,51 @@ namespace jank::read
...
+        return c < 0xD800 || c > 0xDFFF;
+      }
+      else if (c <= 0x10FFFF)
+      {
+        return true;
+      }
+      return false;
+    }
+
+    static native_bool is_special_char(char32_t c)
```
jeaye, line 303: Please make anything `const` which can be const.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -782,21 +889,58 @@ namespace jank::read
             return ok(token{ token_start, pos - token_start, token_kind::deref });
           }
         default:
+          if (is_symbol_char(start))
```
jeaye, line 892: Why would we have a symbol char in this default? Is it only because it's a utf8 char? If so, let's use that predicate.

Also, please add a comment explaining why we need to handle symbols a second time here.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -782,21 +889,58 @@ namespace jank::read
...
-    option<char> processor::peek() const
+    result<codepoint, error> processor::peek() const
     {
       auto const next_pos(pos + 1);
-      if(next_pos >= file.size())
+      if (next_pos >= file.size())
+      {
+        return err(error{pos, "No more characters to peek." });
+      }
+      auto const oc = convert_to_codepoint(file.substr(next_pos), next_pos);
```
jeaye, line 935: Always use direct initialization. `auto const oc{ convert_to_codepoint(file.substr(next_pos), next_pos) };`

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -1336,5 +1305,100 @@ namespace jank::read::lex
         }
       }
     }
+    
+    TEST_CASE("UTF-8")
+    {
+      SUBCASE("UTF-8 symbol")
```
jeaye, line 1311: You're already in the UTF-8 test case, so no need to repeat UTF-8 in each sub case name.

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -1336,5 +1305,100 @@ namespace jank::read::lex
...
+      {
+        processor p{ "👍" };
+        native_vector<result<token, error>> tokens(p.begin(), p.end());
+        CHECK(tokens
+              == make_tokens({
+                  { 0, 4, token_kind::symbol, "👍"sv }
+                }));
+      }
```
jeaye, line 1319: Please add a subcase for qualified symbols and kewords. i.e. `😎/👍` and `:😎/👍`

compiler+runtime/src/cpp/main.cpp
```diff
@@ -231,6 +231,9 @@ try
   using namespace jank;
   using namespace jank::runtime;
 
+  /* Set locale */
```
jeaye, line 234: This is what I call a "what" comment, which is the type of comment to most readily avoid. It says what the code does. In this case, and in most cases with "what" comments, the code is obvious. Instead, comments should generally answer other questions, such as "why", "how", and "when".

For an example, take a look at the comment right below this code, regarding the GC. We don't say "Enable GC interior pointers". That's what the code does. Instead, we provide context about why the code is there and when other GC-related code might happen.

In this case, the useful comment is to explain "why" we're setting this locale. What happens if we don't?

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -700,9 +757,20 @@ namespace jank::read
...
-            ++pos;
-
-            switch(oc.unwrap_or(' '))
+            char32_t c{};
+            size_t size{};
+            if(oc.is_err())
+            {
+              c = ' ';
+              ++pos;
+            }
```
jeaye, line 766: This pattern of replacing an `unwrap_or` with a lot more code shows up a few times. Why not use `unwrap_or(codepoint{ ' ', 1 })` and then remove all of the extra code? We should be working with `codepoint` structs here, not individual `char32_t c` and `size_t size` locals over and over.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -260,11 +266,53 @@ namespace jank::read
...
+        return true;
+      }
+      else if(c <= 0xFFFF)
+      {
+        return c < 0xD800 || c > 0xDFFF;
+      }
+      else if(c <= 0x10FFFF)
+      {
+        return true;
+      }
```
jeaye, line 300: This is good code, but these values are magic! This is the perfect sort of fn which needs a comment inside which at the very least links to the resource used to do this logic and, at best, also describes how it's working.

compiler+runtime/src/cpp/jank/read/lex.cpp
jeaye: Highlighting this one again. All of our `size_t` usages for codepoints can be `uint8_t`. On a 64 bit machine, that's the difference between 1 byte and 8 bytes, which is significant.

jeaye:
Nice work, Monty!

PR #134, opened 2024-11-22
feat(jit-processor): allow adding include, shared libs from cli

Samy-33:


compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -16,6 +16,18 @@
 
 namespace jank::jit
 {
+
+  inline native_persistent_string shared_lib_name(native_persistent_string const &lib)
```
jeaye, line 20: No need for this to be inline. Doesn't buy us anything here.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -148,6 +180,13 @@ namespace jank::jit
     compiler_instance->LoadRequestedPlugins();
 
     interpreter = llvm::cantFail(clang::Interpreter::create(std::move(compiler_instance)));
+
+    auto const &load_result{ load_dynamic_libs(opts.libs) };
+    if(load_result.is_err())
+    {
+      std::cerr << load_result.expect_err().c_str() << "\n";
+      std::exit(1);
```
jeaye, line 188: We want to throw the error as an exception, not exit.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -199,4 +238,54 @@ namespace jank::jit
...
     llvm::logAllUnhandledErrors(std::move(err), llvm::errs(), "error: ");
   }
+
+  option<native_persistent_string>
+  processor::find_dynamic_lib(native_persistent_string const &lib) const
+  {
+    auto const &lib_name{ shared_lib_name(lib) };
+    for(auto const &lib_dir : library_dirs)
+    {
+      auto lib_abs_path{ fmt::format("{}/{}", lib_dir.string(), lib_name) };
```
jeaye, line 248: Should be `const`.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -199,4 +238,54 @@ namespace jank::jit
...
+
+    return none;
+  }
+
+  result<void, native_persistent_string>
+  processor::load_dynamic_libs(native_vector<native_persistent_string> const &libs) const
+  {
+    for(auto const &lib : libs)
+    {
+      auto const &lib_abs_path{ boost::filesystem::absolute(lib.c_str()) };
```
jeaye, line 263: I think we want to be smarter about this. We can check `boost::filesystem:path{ lib }.is_absolute()` and then load it if it exists or fail. If it's a lib name, we don't want to just try to make it absolute, since that's going to try to load the lib name in the current directory, which may not actually be in the library dirs, right?

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -199,4 +238,54 @@ namespace jank::jit
...
+        }
+      }
+    }
+
+    return ok();
+  }
+
+  void processor::load_object(native_persistent_string const &path) const
+  {
+    std::cerr << "Loading dynamic library `" << path << "`\n";
```
jeaye, line 288: We don't want any logging by default. We may add a verbose mode to jank later, but for now we're silent.

compiler+runtime/include/cpp/jank/util/cli.hpp
```diff
@@ -20,6 +20,12 @@ namespace jank::util::cli
     native_bool profiler_enabled{};
     native_transient_string profiler_file{ "jank.profile" };
     native_bool gc_incremental{};
+    
+    /* TODO: New category? */
```
jeaye, line 24: Native dependencies.

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -27,6 +27,20 @@ namespace jank::util::cli
...
+    /* TODO: new category */
+    cli.add_option("-I,--include-dir",
+                   opts.include_dirs,
+                   "Absolute or relative path to the directory for includes resolution");
+    cli.add_option("-L,--library-dir",
+                   opts.library_dirs,
+                   "Absolute or relative path to the directory to search dynamic libraries in");
+    cli.add_option("-D,--define-macro",
+                   opts.define_macros,
+                   "Defines macro value, sets to 1 if omitted");
```
jeaye, line 39: Let's call out that these can each be specified many times.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -16,6 +16,18 @@
...
 {
+
+  inline native_persistent_string shared_lib_name(native_persistent_string const &lib)
+#if defined(__APPLE__)
+  {
+    return fmt::format("{}.dylib", lib);
+  }
+#elif defined(__linux__)
+  {
+    return fmt::format("lib{}.so", lib);
```
jeaye, line 27: There's an interesting thing with shared object versioning and I'm not sure if it's going to bite us. There can be a `libfoo.so.1`, for example, and `libfoo.so.2`. Will you do some digging into this, to see if it's anything we need to support? I'll also do the same.

jeaye: https://unix.stackexchange.com/a/476 implies that `-lfoo` with gcc/clang can't specify the version number either. So we can rely on the `libfoo.so.2` form or the absolute path form for that. We just need to not assume that it ends in `.so` anywhere, since it may not!

compiler+runtime/src/cpp/jank/jit/processor.cpp
Samy-33: We do have some std:cerr's in this file. What's the purpose of them being there? Should I remove them?

jeaye: Ignore anything that was there. I'll be revamping all of the error reporting as my next task. But the happy path shouldn't be logging anything.

compiler+runtime/src/cpp/jank/jit/processor.cpp
Samy-33: done, now we check if `libfoo.so` exists if `-lfoo` was provided as an argument to the cli.

In addition to that, we also check for `dynamic-lib-path/foo`. This handles the case where the library name is given as `-llibfoo.so.1`

jeaye:
Nice work!

PR #137, opened 2024-11-24
Add support for runtime ratio object

### LINKED ISSUE #66: Add ratio support
Clojure supports `3/4` as a ratio. ClojureScript supports this at the reader level, but not at the object level. It will just read it as `(/ 3.0 4)`, whereas Clojure has a specific ratio object.

For jank, let's follow ClojureScript here until we have a reason to add a ratio object.

Official docs on this: https://clojure.org/reference/data_structures#_ratio

jianlingzhong:
everything should work. 

```clojure
❯ build/jank repl
Bottom of clojure.core
clojure.core=> (+ 1/2 1/3)
5/6
clojure.core=> (- 1/2 1/3)
1/6
clojure.core=> (* 1/2 1/3)
1/6
clojure.core=> (/ 1/2 1/3)
3/2
clojure.core=> (+ 1/2 1/2)
1
clojure.core=> (+ 1/2 1)
3/2
clojure.core=> (+ 1 1/3)
4/3
clojure.core=> (+ 1.0 1/3)
1.3333333333333333
clojure.core=> (+ 1.0 -1/3)
0.6666666666666667
clojure.core=> (* 1.0 -1/3)
-0.3333333333333333
clojure.core=> (> 1 1/3)
true
clojure.core=> (< 1 1/3)
false
clojure.core=> (== 1 1/1)
true
clojure.core=> (== 2 1/2)
false
clojure.core=> (mod 4/5 2)
0.8
clojure.core=> (mod 0.8 2)
0.8
clojure.core=> (mod 16/5 2)
1.2000000000000002
clojure.core=> (mod 16/5 3/2)
0.20000000000000018
clojure.core=> (mod (/ 16.0 5) (/ 3.0 2))
0.20000000000000018
clojure.core=> (<= (/ 1.0 3) 1/3)
true
clojure.core=> (== 1/3 0.3333)
false
clojure.core=> (== 1/3 0.3333333333333333333333)
true
```

this closes #66.


compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+    if(denominator == 0)
+    {
+      throw std::invalid_argument("Ratio denominator cannot be zero.");
+    }
+    simplify();
+  }
+
+  object_ptr obj::ratio::create(native_integer const numerator, native_integer const denominator)
+  {
+    auto ratio_data = make_box<obj::ratio_data>(numerator, denominator);
```
jeaye, line 23: `make_box` here is doing a dynamic allocation, via the GC, but `ratio_data` is only used locally here. No need for the GC or dynamic memory. We can just do `obj::ratio_data data{ numerator, denominator };`.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+    if(ratio_data->denominator == 1)
+    {
+      return make_box<obj::integer>(ratio_data->numerator);
+    }
+    auto r(make_box<obj::ratio>());
```
jeaye, line 28: Why not use the ctor here which takes in the numerator/denominator?

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  native_integer obj::ratio::to_integer() const
+  {
+    return data.numerator / data.denominator;
+  }
+
+  void obj::ratio_data::simplify()
+  {
+    int gcd = std::gcd(numerator, denominator);
```
jeaye, line 51: Please always use direct initialization and avoid spelling out obvious types.

```cpp
auto const gcd{ std::gcd(numerator, denominator) };
```

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+  native_bool obj::ratio::equal(object const &o) const
+  {
+    if(o.type == object_type::integer)
+    {
+      auto const i(expect_object<obj::integer>(&o));
+      return this->data == i->data;
+    }
+
+    if(o.type == object_type::real)
+    {
```
jeaye, line 93: Rather than checking the type and then using `expect_object` as a pattern, please use `dyn_cast`.

```cpp
    if(auto const i = dyn_cast<obj::integer>(&o))
    {
      return this->data == i->data;
    }
```

Note that C++ doesn't allow direct initialization in this one case of initializing within an `if`.

Also, there is no need for `this->` here, so please remove that.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+
+  native_integer obj::ratio::compare(static_object const &o) const
+  {
+    return (data > o.data) - (data < o.data);
+  }
+
+  object_ptr operator+(obj::ratio_data const l, obj::ratio_data const r)
+  {
+    native_integer denom = l.denominator * r.denominator;
+    native_integer num = l.numerator * r.denominator + r.numerator * l.denominator;
```
jeaye, line 126: Please always use direct initialization and avoid spelling out obvious types. Also make everything const that can be.

```cpp
    auto const denom{ l.denominator * r.denominator };
    auto const num{ l.numerator * r.denominator + r.numerator * l.denominator };
```

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+    return obj::ratio::create(num, denom);
+  }
+
+  obj::ratio_ptr operator+(obj::integer_ptr l, obj::ratio_data const r)
+  {
+    return expect_object<obj::ratio>(obj::ratio_data(l->data, 1LL) + r);
```
jeaye, line 132: Is `expect_object` needed here? The `ratio_data + integer_ptr` overload should return a `ratio_ptr`, right?

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  obj::ratio_ptr operator+(obj::ratio_data const l, obj::integer_ptr r)
```
jeaye, line 135: Please make `r` const.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+  {
+    return l.to_real() + r;
+  }
+
+  native_real operator+(native_real l, obj::ratio_data const r)
+  {
+    return l + r.to_real();
+  }
+
+  obj::ratio_ptr operator+(obj::ratio_data const l, native_integer r)
```
jeaye, line 160: Please make `r` const.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+  {
+    return expect_object<obj::ratio>(
+      obj::ratio::create(l.numerator + r * l.denominator, l.denominator));
```
jeaye, line 163: Are we 100% certain that calling `ratio::create` here will give us a ratio? If so, we should just make a ratio.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+  {
+    return l - r.to_real();
+  }
+
+  obj::ratio_ptr operator-(obj::ratio_data const l, native_integer r)
+  {
+    return expect_object<obj::ratio>(
+      obj::ratio::create(l.numerator - r * l.denominator, l.denominator));
+  }
+
```
jeaye, line 215: Please make these `r` params const.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,401 @@
+#include "jank/runtime/obj/ratio.hpp"
+#include <doctest/doctest.h>
+#include <stdexcept>
+
+namespace jank::runtime
+{
+  TEST_SUITE("ratio_data")
```
jeaye, line 7: Let's keep all of this in one `ratio` test suite. That'll make it easier to run from the cli, using `-ts=ratio`. You can separate the ratio_data tests with test cases or sub cases, if you'd like.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,401 @@
+#include "jank/runtime/obj/ratio.hpp"
+#include <doctest/doctest.h>
```
jeaye, line 2: This include must always go last. Please put it last and copy the comment from the other test files which says it must always go last.

Also, C and C++ includes always go first in jank, so please move `stdexcept` to the top.

Finally, jank includes always use `<>` so please update that.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,401 @@
...
+{
+  TEST_SUITE("ratio_data")
+  {
+    TEST_CASE("Ratio Constructor")
+    {
+      SUBCASE("Valid Ratio")
+      {
+        obj::ratio_data ratio{ 4, 6 };
+        CHECK(ratio.numerator == 2); // Simplified
+        CHECK(ratio.denominator == 3);
```
jeaye, line 15: I think we want `CHECK_EQ` for all of these equality checks. That will allow doctest to provide better error messages when they're not equal.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,401 @@
...
+      SUBCASE("ToString")
+      {
+        auto ratio_ptr = obj::ratio::create(3, 4);
+        auto ratio = expect_object<obj::ratio>(ratio_ptr);
+        CHECK(ratio->to_string() == "3/4");
+      }
+      SUBCASE("Hashing")
+      {
+        auto ratio1 = expect_object<obj::ratio>(obj::ratio::create(2, 3));
+        auto ratio2 = expect_object<obj::ratio>(obj::ratio::create(2, 3));
```
jeaye, line 126: Please always use direct initialization. `auto const ratio_ptr{ obj::ratio::create(3, 4) };`

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,401 @@
...
+    }
+  }
+
+  TEST_SUITE("ratio")
+  {
+    obj::ratio_ptr make_ptr(native_integer num, native_integer denom)
+    {
+      auto r(make_box<obj::ratio>());
+      r->data.numerator = num;
+      r->data.denominator = denom;
```
jeaye, line 296: Please use a ctor to initialize `obj::ratio` with numerator/denominator instead. Then you don't need this whole `make_ptr` fn, since you can just use `make_box`.

compiler+runtime/include/cpp/jank/runtime/core/make_box.hpp
```diff
@@ -60,6 +60,12 @@ namespace jank::runtime
     return make_box<runtime::obj::real>(r);
   }
 
+  [[gnu::always_inline, gnu::flatten, gnu::hot]]
+  inline auto make_box(obj::ratio_data const r)
+  {
+    return expect_object<runtime::obj::ratio>(obj::ratio::create(r.numerator, r.denominator));
```
jeaye, line 66: So, when we construct with a `ratio_data`, we know for certain that `ratio::create` will always give us back a ratio? Is there any possibility this doesn't happen?

If we know for certain we're getting an `obj::ratio` back, we should just use `make_box` to build it instead.

compiler+runtime/include/cpp/jank/runtime/obj/ratio.hpp
```diff
@@ -0,0 +1,146 @@
...
+    static_object(static_object &&) = default;
+    static_object(static_object const &) = default;
+    static_object(native_integer const numerator, native_integer const denominator);
+
+    native_real to_real() const;
+    void simplify();
+
+    native_integer numerator{};
+    native_integer denominator{};
+  };
```
jeaye, line 19: I don't think you want a new jank object type for `ratio_data`. You didn't add it to `erasure.hpp` for the same reason. We don't ever want a `ratio_data` on its own. It's just a detail of a `ratio`. You also didn't implement any of the `objectable` interface, so this isn't a proper jank object anyway. This should just be a struct with none of the other stuff involved.

```cpp
namespace obj
{
  struct ratio_data
  {
    ratio_data(native_integer numerator, native_integer denominator);

    native_real to_real() const;

    native_integer numerator{};
    native_integer denominator{};
  };
}
```

There is also no need for a `simplify` function. Just do that in the ctor.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -516,20 +529,11 @@ namespace jank::runtime
...
-            if constexpr(std::is_same_v<L, obj::real> || std::is_same_v<R, obj::real>)
-            {
-              return make_box(std::fmod(typed_l, typed_r->data));
-            }
-            else
-            {
-              return make_box(typed_l % typed_r->data);
-            }
+            auto typed_l_data = to_number(typed_l);
+            auto typed_r_data = to_number(typed_r->data);
```
jeaye, line 535: Please use direct initialization everywhere and make everything const.

```cpp
            auto const typed_l_data{ to_number(typed_l) };
            auto const typed_r_data{ to_number(typed_r->data) };
```

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
+#include <jank/runtime/obj/ratio.hpp>
+#include <fmt/compile.h>
+#include <limits>
+
+namespace jank::runtime
+{
+
+  constexpr native_real EPSILON = std::numeric_limits<native_real>::epsilon();
```
jeaye, line 8: Please use direct initialization and `auto` here. Also, in jank's code we don't use SCREAMINGCASE for anything.

```cpp
constexpr static auto epsilon{ std::numeric_limits<native_real>::epsilon() };
```

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: we need to call the ctor for ratio_data to simplify the ratio before we can check if the denominator is 1 or not. i changed this to `auto ratio_data = obj::ratio_data(num, denom)`, would this work better?

jeaye: Please change it to the line I specified: `obj::ratio_data data{ numerator, denominator };`

`auto` is meant to help avoid the type when the right hand expression already determines that type. But, in this case, there is no right hand expression. We're just creating a `obj::ratio_data`. Also, we always use direct initialization.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: i'm trying to not include a ctor for `ratio` that takes num and denom, so that everything goes through `ratio_data` instead. that way we can avoid the possibility that someone accidentally init a ratio with 1 as denom. 

i added a ctor with a `ratio_data` as the only arg. does this work better?

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: done.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated. the reason for that is, we don't want code to terminate in case `expect_object` fails?

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: yeah, i was trying to reuse `ratio_data + ratio_data` (which would return object_ptr). but i suppose we can be more efficient.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated all places to include a `const`.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated all places to include a `const`.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: yes, a `ratio_data` exists only if the denominator is not 1 (its ctor made sure of that). so a ratio + int will always be another ratio. 

updated to be more efficient.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  native_real operator/(native_real l, obj::ratio_data const r)
+  {
+    return l / r.to_real();
+  }
+
+  obj::ratio_ptr operator/(obj::ratio_data const l, native_integer r)
+  {
+    return l / make_box<obj::integer>(r);
```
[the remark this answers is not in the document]
jianlingzhong: i was trying reuse the boxed version. but i guess the other way around makes more sense. updated. the whole ratio.cpp file should not have any use of `expect_object` now.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated all. using `CHECK_EQ`, `CHECK_LT`, etc., instead.  for certain cases, this does not seem to work once `ratio_data` is not a jank object.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated.

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated.

compiler+runtime/include/cpp/jank/runtime/core/make_box.hpp
jianlingzhong: yes, it should always be a ratio since we made sure a ratio_data is created only when its denominator does not simplify to 1. updated.

compiler+runtime/include/cpp/jank/runtime/obj/ratio.hpp
jianlingzhong: i've made a large change to the use of `ratio_data`, changing it to a non-jank object.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1277,10 +1278,19 @@ namespace jank::read::parse
...
       return err(error{ token.pos, "Divide by zero" });
     }
-    return object_source_info{ make_box<obj::real>(static_cast<native_real>(ratio_data.numerator)
-                                                   / ratio_data.denominator),
-                               token,
-                               token };
+    auto ratio = obj::ratio::create(ratio_data.numerator, ratio_data.denominator);
+    if(ratio->type == object_type::ratio)
+    {
+      return object_source_info{ expect_object<obj::ratio>(ratio), token, token };
```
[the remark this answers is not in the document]
jianlingzhong: keep this as well, since dyn_cast is not available.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1277,10 +1278,19 @@ namespace jank::read::parse
     {
       return err(error{ token.pos, "Divide by zero" });
     }
-    return object_source_info{ make_box<obj::real>(static_cast<native_real>(ratio_data.numerator)
-                                                   / ratio_data.denominator),
-                               token,
-                               token };
+    auto ratio = obj::ratio::create(ratio_data.numerator, ratio_data.denominator);
```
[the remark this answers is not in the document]
jianlingzhong: updated.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
jianlingzhong: updated.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: updated.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: sg. updated.

compiler+runtime/src/cpp/jank/runtime/obj/ratio.cpp
jeaye: Yep, that can work!

compiler+runtime/test/cpp/jank/runtime/obj/ratio.cpp
jianlingzhong: > updated all. using CHECK_EQ, CHECK_LT, etc., instead. for certain cases, this does not seem to work once ratio_data is not a jank object.

this turns out to be caused by operators not being implemented in the `jank::runtime::obj` (but `ratio_data` is). everything works once operators are also moved into `jank::runtime::obj`.

compiler+runtime/test/cpp/jank/jit/processor.cpp
```diff
@@ -182,6 +182,11 @@ namespace jank::jit
           std::cerr.flush();
         }
       }
+      for(auto const f : failures)
+      {
+        std::cout << f.path << std::endl;
+        std::cout << f.error << std::endl;
+      }
```
jeaye, line 189: Why is this here?

compiler+runtime/bin/compile
```diff
@@ -2,4 +2,6 @@
 
 set -euo pipefail
 
+git_root=$(git rev-parse --show-toplevel)
+"$git_root"/compiler+runtime/bin/format
```
jeaye, line 6: Please use a pre-commit hook for this instead. Nobody else needs this hooked into the compile script; our editors are running the formatter for us.

compiler+runtime/include/cpp/jank/read/lex.hpp
```diff
@@ -150,14 +150,19 @@ namespace jank::read::lex
...
     iterator begin();
     iterator end();
 
     size_t pos{};
-    /* Whether or not the previous token requires a space after it. */
+    native_integer radix{ 10 };
+    /* The 'r' used in arbitrary radix (prefixed with N and then r, where N is the radix (2 <= radix <= 36); */
+    /* e.g. 2r10101 for binary, 16rebed00d for hex) */
+    native_bool found_r{};
+    /* Whether the previous token requires a space after it. */
```
jeaye, line 165: This was a ratio PR, but it now contains some of your radix changes too. Any reason these were merged, especially when you have a separate radix PR?

compiler+runtime/test/cpp/jank/jit/processor.cpp
jianlingzhong: should be removed now.

compiler+runtime/bin/compile
jianlingzhong: removed.

compiler+runtime/include/cpp/jank/read/lex.hpp
jianlingzhong: i dropped the integer related commit.

jeaye:
Nice work, Jianling!

PR #138, opened 2024-11-25
feat(objs): adds native_pointer_wrapper obj

### LINKED ISSUE #136: Create native_pointer_wrapper
jank needs a way to store arbitrary native pointers as boxed objects. Until we get to a point where the jank compile/runtime can understand and codegen arbitrary C or C++ types, hiding things behind an opaque pointer will be necessary.

native_pointer_wrapper will store a `void*`. We should be able to construct it with a `void*` and then pull the pointer out as some other type. We don't need any jank fns for this. This will be used for storing arbitrary native pointers in a jank object. When you pull it out, you can cast it to the type you want and it's up to you to make sure it's correct.

```cpp
auto box{ make_box<obj::native_pointer_wrapper>(my_database_obj) };

// later on
void foo(object_ptr const o)
{
  auto box{ try_object<obj::native_pointer_wrapper>(o) };
  database *db{ box->as<database>() };
}
```

Samy-33:
closes #136 
```cpp
➜  compiler+runtime git:(feat/native_pointer_wrapper) ✗ ./build/jank cpp-repl
Bottom of clojure.core
native> using namespace jank;
native> using namespace jank::runtime;
native> uint32_t a{22};
native> uint32_t *pa{&a};
native>
native> fmt::println("native: {}", fmt::ptr(pa));
native: 0x74b63d92d000
native>
native> auto boxed{make_box(pa)};
native> fmt::println("boxed: {}", boxed->to_string());
boxed: native_pointer_wrapper@0x74b62e8dcfc0
native>
native> auto pointer_wrapper{try_object<obj::native_pointer_wrapper>(boxed)};
native> auto unboxed{pointer_wrapper->as<uint32_t>()};
native>
native> fmt::println("unboxed: {}", fmt::ptr(unboxed));
unboxed: 0x74b63d92d000
native> fmt::println("unboxed value: {}", *unboxed);
unboxed value: 22
native>
```

After the changes:
```cpp
➜  compiler+runtime git:(feat/native_pointer_wrapper) ✗ ./build/jank cpp-repl
Bottom of clojure.core
native> using namespace jank;
native> using namespace jank::runtime;
native> uint32_t a{22};;
native> uint32_t *pa{&a};
native>
native> fmt::println("native: {}", fmt::ptr(pa));
native: 0x79e0b59e5000
native>
native> auto boxed{make_box<obj::native_pointer_wrapper>(pa)};
native> fmt::println("boxed: {}", boxed->to_string());
boxed: native_pointer_wrapper@0x79e0a6ad5fc0
native>
native> auto pointer_wrapper{try_object<obj::native_pointer_wrapper>(boxed)};
native> auto unboxed{pointer_wrapper->as<uint32_t>()};
native>
native> fmt::println("unboxed: {}", fmt::ptr(unboxed));
unboxed: 0x79e0b59e5000
native> fmt::println("unboxed value: {}", *unboxed);
unboxed value: 22
native>
```

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/core/make_box.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core/make_box.hpp
@@ -83,14 +83,6 @@ namespace jank::runtime
   }
 
   template <typename T>
-  requires std::is_pointer_v<T>
-  [[gnu::always_inline, gnu::flatten, gnu::hot]]
-  inline auto make_box(T const d)
-  {
-    return make_box<obj::native_pointer_wrapper>(static_cast<void *>(d));
-  }
-
-  template <typename T>
   requires std::is_floating_point_v<T>
   [[gnu::always_inline, gnu::flatten, gnu::hot]]
   inline auto make_box(T const d)
--- a/compiler+runtime/include/cpp/jank/runtime/core/make_box.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core/make_box.hpp
@@ -83,6 +83,14 @@ namespace jank::runtime
   }
 
   template <typename T>
+  requires std::is_pointer_v<T>
+  [[gnu::always_inline, gnu::flatten, gnu::hot]]
+  inline auto make_box(T const d)
+  {
+    return make_box<obj::native_pointer_wrapper>(static_cast<void *>(d));
+  }
+
+  template <typename T>
   requires std::is_floating_point_v<T>
   [[gnu::always_inline, gnu::flatten, gnu::hot]]
   inline auto make_box(T const d)
--- a/compiler+runtime/src/cpp/jank/runtime/obj/native_pointer_wrapper.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/native_pointer_wrapper.cpp
@@ -2,7 +2,7 @@
 
 namespace jank::runtime
 {
-  obj::native_pointer_wrapper::static_object(void *d)
+  obj::native_pointer_wrapper::static_object(void * const d)
     : data{ d }
   {
   }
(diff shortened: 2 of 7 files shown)


compiler+runtime/include/cpp/jank/runtime/core/make_box.hpp
```diff
@@ -82,6 +82,14 @@ namespace jank::runtime
     return make_box<runtime::obj::persistent_list>(l);
   }
 
+  template <typename T>
+  requires std::is_pointer_v<T>
+  [[gnu::always_inline, gnu::flatten, gnu::hot]]
+  inline auto make_box(T const d)
+  {
+    return make_box<obj::native_pointer_wrapper>(static_cast<void *>(d));
+  }
```
jeaye, line 91: I thought about this, and it seemed clever at first, but I think it's too clever. Anything we don't catch in another overload gets here, even if we were previously relying on implicit conversions. Let's remove this and rely on explicit `make_box<obj::native_pointer_wrapper>()` for now.

compiler+runtime/src/cpp/jank/runtime/obj/native_pointer_wrapper.cpp
```diff
@@ -0,0 +1,45 @@
+#include <jank/runtime/obj/native_pointer_wrapper.hpp>
+
+namespace jank::runtime
+{
+  obj::native_pointer_wrapper::static_object(void *d)
```
jeaye, line 5: Please make `d` const.

compiler+runtime/src/cpp/jank/runtime/obj/native_pointer_wrapper.cpp
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  native_bool obj::native_pointer_wrapper::equal(object const &o) const
+  {
+    if(o.type != object_type::native_pointer_wrapper)
+    {
+      return false;
+    }
+
+    auto const c(expect_object<obj::native_pointer_wrapper>(&o));
```
jeaye, line 17: Please use `dyn_cast` instead of the pattern of checking for the type and then using `expect_type`.

```cpp
if(auto const c = dyn_cast<obj::native_pointer_wrapper>(&o))
{
  return data == c->data;
}
```

jeaye:
Well done! :muscle:

PR #140, opened 2024-12-03
Support octal, hex, and arbitrary radix numbers

jianlingzhong:
code is a bit ugly but should work:

```clojure
❯ build/jank repl
Bottom of clojure.core
clojure.core=> 071
57
clojure.core=> -071
-57
clojure.core=> 08
Read error (0 - 2): invalid number: char 8 are invalid for radix 8
clojure.core=> 08.9
8.9
clojure.core=> 08e2
800
clojure.core=> 0-7.1
0
clojure.core=> 0xabc
2748
clojure.core=> 0Xabc
2748
clojure.core=> -0xef
-239
clojure.core=> 0xg
Read error (0 - 3): invalid number: char g are invalid for radix 16
clojure.core=> 0x-a
Read error (0 - 4): invalid number: char - are invalid for radix 16
clojure.core=> 2r011110
30
clojure.core=> 8r71
57
clojure.core=> -8r71
-57
clojure.core=> 1r0
Read error (0 - 3): invalid number: radix 1 is out of range
clojure.core=> -36razxyu
-18473142
clojure.core=> -30razxyu
Read error (0 - 9): invalid number: char zxyu are invalid for radix 30
clojure.core=> 37rzb
Read error (0 - 5): invalid number: radix 37 is out of range
clojure.core=> 36r4.5
Read error (0 - 6): invalid number: char . are invalid for radix 36
clojure.core=> 36r-4.5
Read error (0 - 7): invalid number: char -. are invalid for radix 36
clojure.core=> 3e-4
0.0003
clojure.core=> 16r3e4
996
clojure.core=>  36r3e4
4396
clojure.core=> 3e2/3
Read error (0 - 3): invalid ratio
clojure.core=> 7r3e4
Read error (0 - 5): invalid number: char e are invalid for radix 7
clojure.core=>  0xe
14
clojure.core=> 16re3
227
clojure.core=> 16r3e
62
clojure.core=> 0x3e
62
clojure.core=>  0x3e3
995
clojure.core=> 16r3e3
995
```

### DIFF
--- a/compiler+runtime/include/cpp/jank/read/lex.hpp
+++ b/compiler+runtime/include/cpp/jank/read/lex.hpp
@@ -213,14 +213,14 @@ namespace jank::read::lex
     processor(native_persistent_string_view const &f);
 
     result<token, error> next();
-    result<codepoint, error> peek() const;
+    result<codepoint, error> peek(native_integer const ahead = 1) const;
     option<error> check_whitespace(native_bool const found_space);
 
     iterator begin();
     iterator end();
 
     size_t pos{};
-    /* Whether or not the previous token requires a space after it. */
+    /* Whether the previous token requires a space after it. */
     native_bool require_space{};
     /* True when seeing a '/' following a number. */
     native_bool found_slash_after_number{};
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -1,5 +1,6 @@
 #include <iostream>
 #include <iomanip>
+#include <fmt/format.h>
 
 #include <jank/read/lex.hpp>
 
@@ -305,6 +306,17 @@ namespace jank::read
       return false;
     }
 
+    static int safe_cast_char32_t_to_int(char32_t input)
+    {
+      // Check if the value exceeds the limit for wint_t
+      if(input > static_cast<char32_t>(INT_MAX))
+      {
+        throw std::runtime_error("value out of range for int");
+      }
+      // Perform the safe conversion
+      return static_cast<int>(input);
+    }
+
     static native_bool is_special_char(char32_t const c)
     {
       return c == '(' || c == ')' || c == '{' || c == '}' || c == '[' || c == ']' || c == '"'
@@ -313,10 +325,50 @@ namespace jank::read
 
     static native_bool is_symbol_char(char32_t const c)
     {
-      return !std::iswspace(c) && !is_special_char(c)
+      return !std::iswspace(safe_cast_char32_t_to_int(c)) && !is_special_char(c)
         && (std::iswalnum(static_cast<wint_t>(c)) != 0 || c == '_' || c == '-' || c == '/'
             || c == '?' || c == '!' || c == '+' || c == '*' || c == '=' || c == '.' || c == '&'
             || c == '<' || c == '>' || c == '#' || c == '%' || is_utf8_char(c));
+    }
+
+    static native_bool is_lower_letter(char32_t const c)
+    {
+      return c >= 'a' && c <= 'z';
+    }
+
+    static native_bool is_upper_letter(char32_t const c)
+    {
+      return c >= 'A' && c <= 'Z';
+    }
+
+    static native_bool is_letter(char32_t const c)
+    {
+      return is_lower_letter(c) || is_upper_letter(c);
+    }
+
+    static native_bool is_valid_num_char(char32_t const c, native_integer const radix)
+    {
+      if(c == '-' || c == '+' || c == '.')
+      {
+        return radix == 10;
+      }
+      if(radix == 10 && (c == 'e' || c == 'E'))
+      {
+        return true;
+      }
+      if(radix <= 10)
+      {
+        return c >= '0' && c < '0' + radix;
+      }
+      if(is_upper_letter(c))
+      {
+        return c < 'A' + radix - 10;
+      }
+      if(is_lower_letter(c))
+      {
+        return c < 'a' + radix - 10;
+      }
+      return c >= '0' && c <= '9';
     }
 
     result<token, error> processor::next()
@@ -330,8 +382,7 @@ namespace jank::read
           return ok(token{ pos, token_kind::eof });
         }
 
-        auto const c(file[pos]);
-        if(std::isspace(c) == 0 && c != ',')
+        if(auto const c(file[pos]); std::isspace(c) == 0 && c != ',')
         {
           break;
         }
@@ -340,6 +391,7 @@ namespace jank::read
         ++pos;
       }
 
+      native_bool found_r{};
       auto const token_start(pos);
       auto const oc{ convert_to_codepoint(file.substr(token_start), token_start) };
       if(oc.is_err())
@@ -377,7 +429,7 @@ namespace jank::read
             auto const ch(peek());
             pos++;
 
-            if(ch.is_err() || std::iswspace(ch.expect_ok().character))
+            if(ch.is_err() || std::iswspace(safe_cast_char32_t_to_int(ch.expect_ok().character)))
             {
               return err(error{ token_start, "Expecting a valid character literal after \\" });
             }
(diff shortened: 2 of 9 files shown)


compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -668,8 +868,8 @@ namespace jank::read
             pos++;
 
             /* Unescaped strings can be read right from memory, but escaped strings require
-             * some processing first, to turn the escape sequences into the necessary characters.
-             * We use distinct token types for these so we can optimize for the typical case. */
+                     * some processing first, to turn the escape sequences into the necessary characters.
+                     * We use distinct token types for these so we can optimize for the typical case. */
```
jeaye, line 872: This doesn't look right.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -542,26 +681,79 @@ namespace jank::read
             }
 
             /* Tokens beginning with - are ambiguous; it's only a negative number if it has numbers
-             * to follow.
-             * TODO: handle numbers starting with `+` */
+                     * to follow.
+                     * TODO: handle numbers starting with `+` */
```
jeaye, line 685: Looks like an incorrect change.

jeaye: We can remove that TODO now, right? We handle numbers starting in `+`. Hmm, I don't see any tests for them though.

compiler+runtime/include/cpp/jank/read/lex.hpp
```diff
@@ -151,14 +151,19 @@ namespace jank::read::lex
     processor(native_persistent_string_view const &f);
 
     result<token, error> next();
-    result<codepoint, error> peek() const;
+    result<codepoint, error> peek(native_integer const ahead = 1) const;
     option<error> check_whitespace(native_bool const found_space);
-    
+    native_bool is_valid_num_char(char const c) const;
+
```
jeaye, line 157: This doesn't need to be a member fn, from what I can see. It can just be a static fn in the cpp file.

compiler+runtime/include/cpp/jank/read/lex.hpp
```diff
@@ -149,14 +149,19 @@ namespace jank::read::lex
...
     iterator begin();
     iterator end();
 
     size_t pos{};
-    /* Whether or not the previous token requires a space after it. */
+    native_integer radix{ 10 };
+    /* The 'r' used in arbitrary radix (prefixed with N and then r, where N is the radix (2 <= radix <= 36); */
+    /* e.g. 2r10101 for binary, 16rebed00d for hex) */
+    native_bool found_r{};
```
jeaye, line 163: Can these be stored as locals within the `next` fn? If so, let's prefer that.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -524,15 +662,16 @@ namespace jank::read
                   return err(
                     error{ token_start, pos, "unexpected end of real, expecting exponent" });
                 }
-                break;
+                if(!is_letter(c) || (pos == token_start && file[token_start] != '0'))
+                {
+                  break;
+                }
```
jeaye, line 668: What're we checking here? Can you put whatever explanation you have into a comment, too, please? It's not obvious to me.

jianlingzhong: added some comments for explanation.

compiler+runtime/include/cpp/jank/read/lex.hpp
jianlingzhong: updated them as locals.

compiler+runtime/include/cpp/jank/read/lex.hpp
jianlingzhong: udpated.

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: right, i didn't do anything regarding '+', so i'll keep it here.

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: udpated.

compiler+runtime/include/cpp/jank/read/lex.hpp
```diff
@@ -150,14 +150,16 @@ namespace jank::read::lex
...
+    result<codepoint, error> peek(native_integer const ahead = 1) const;
     option<error> check_whitespace(native_bool const found_space);
 
     iterator begin();
     iterator end();
 
     size_t pos{};
-    /* Whether or not the previous token requires a space after it. */
+    /* The 'r' used in arbitrary radix (prefixed with N and then r, where N is the radix (2 <= radix <= 36); */
+    /* e.g. 2r10101 for binary, 16rebed00d for hex) */
```
jeaye, line 161: We've kept the comment from when we had a new member here.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -440,81 +484,192 @@ namespace jank::read
...
+            auto r_pos{ pos }; /* records the 'r' position if one is found */
+            native_bool found_beginning_negative{};
+
+            if(auto const [c, l]{ peek().unwrap_or({ ' ', 1 }) };
+               file[token_start] == '-' && c == '0' && !found_slash_after_number)
+            {
+              contains_leading_digit = true;
+              if(auto const [f, l]{ peek(2).unwrap_or({ ' ', 1 }) }; f != ' ')
+              {
+                // auto const[n, _] {peek(3).unwrap_or({' ', 1})};
```
jeaye, line 517: Please remove this commented code, as well as the others in the various branches below.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -440,81 +484,192 @@ namespace jank::read
...
+            {
+              contains_leading_digit = true;
+              // auto const[n, _] {peek(2).unwrap_or({' ', 1})};
+              if(auto const [f, l]{ peek().unwrap_or({ ' ', 1 }) }; f != ' ')
+              {
+                if(f != 'x' && f != 'X')
+                {
+                  radix = 8;
+                }
+                if(f == 'x' || f == 'X')
```
jeaye, line 540: Shouldn't this just be an `else`?

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -542,27 +733,91 @@ namespace jank::read
...
+              if(found_r)
+              {
+                return err(error{ token_start,
+                                  pos,
+                                  "unexpected end of radix number, expecting more digits" });
+              }
+              return err(error{
+                token_start,
+                pos,
+                fmt::format("unexpected end of radix {} number, expecting more digits", radix) });
```
jeaye, line 748: Why do we have these two different returns? Does the last one not make sense if `found_r` is true?

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -4,6 +4,8 @@
 #include <jank/read/lex.hpp>
 
 /* This must go last; doctest and glog both define CHECK and family. */
+#include <iostream>
+#include <clang/Basic/Diagnostic.h>
```
jeaye, line 8: Please put these at the top of the file. Also, the comment on line 6 is specifically for the doctest include.

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -467,6 +487,104 @@ namespace jank::read::lex
...
+        CHECK(tokens
+              == make_results({
+                error{  0,  3,        "invalid number: char 3 are invalid for radix 2" },
+                error{  4,  8,       "invalid number: char z are invalid for radix 35" },
+                error{  9, 14,        "invalid number: char e are invalid for radix 8" },
+                error{ 15, 22,       "invalid number: char - are invalid for radix 19" },
+                error{ 23, 25, "unexpected end of radix number, expecting more digits" },
+                error{ 26, 29, "unexpected end of radix number, expecting more digits" }
+        }));
+      }
```
jeaye, line 586: I really like how you've captured both positive and negative tests here. Let's dig deeper into potential negative cases, since I think there are more. Off the top of my head, to get the ball rolling:

* `0r0` (is radix 0 supported?) -- In Clojure, no
* `1r1` (is radix 1 supported?) -- In Clojure, no
* `2048r0` (how high can radix actually go?) -- In Clojure, `36r0` is the highest
* `2.0r1` (what if radix is a real?)
* `2rr1` (what if `r` is specified multiple times?)
* `2r1r` (what if `r` is specified at the end of a valid number?)
* `2r.0` (what if a decimal is specified after `r`?)

jeaye: Please also give the same treatment to octal and hex. I can help with some test cases if you need more suggestions.

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: i don't think so. if the string is "0)", then i don't want radix to become 16.

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: yeah, when found_r is true, we would not know what `radix` is yet until later so the msg could be confusing.

compiler+runtime/test/cpp/jank/read/lex.cpp
jianlingzhong: we actually don't need them. removed.

compiler+runtime/test/cpp/jank/read/lex.cpp
jianlingzhong: updated with more unit tests.

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -460,17 +478,332 @@ namespace jank::read::lex
...
+                error{ 21, 21, "unexpected character: ." },
+                error{ 22, 22, "expected whitespace before next token" },
+                token{ 22, 1, token_kind::integer, 2ll },
+        }));
+      }
+
+      SUBCASE("Arbitrary radix edge cases")
+      {
+        /* exceeds 64-bit integer max */
+        processor p{ "36r0123456789abcdefghijklmnopqrstuvwxyz" };
```
jeaye, line 556: Seems like this should be an error, until we support big numbers. The token we create is lossy.

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -460,17 +478,332 @@ namespace jank::read::lex
...
+        native_vector<result<token, error>> const tokens(p.begin(), p.end());
+        CHECK(tokens
+              == make_tokens({
+                { 0, 39, token_kind::integer, 9223372036854775807ll }
+        }));
+
+        processor p2{ "2r1111111111111111111111111111111111111111111" };
```
jeaye, line 563: There's a neat edge case of a binary number with more bits than our integer. That should also be an error.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -313,12 +325,52 @@ namespace jank::read
 
     static native_bool is_symbol_char(char32_t const c)
     {
-      return !std::iswspace(c) && !is_special_char(c)
+      return !std::iswspace(safe_cast_char32_t_to_int(c)) && !is_special_char(c)
```
jeaye, line 328: I don't think this is what we want. What we want is this:

* We have a `char32_t` and we want to check if it's a space (true or false)

With this change, we have this:

* We have a `char32_t` and we want to check if it's a space (true or false), but we throw an exception if it's larger than an int

Instead, let's just provide our own `iswspace` which takes a `char32_t`. The docs for it are here: https://en.cppreference.com/w/cpp/string/wide/iswspace Looks like a simple switch with 6 cases should work.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -540,27 +737,95 @@ namespace jank::read
...
+                {
+                  invalid_digits.emplace_back(file[i]);
+                }
+              }
+              if(!invalid_digits.empty())
+              {
+                return err(
+                  error{ token_start,
+                         pos,
+                         fmt::format("invalid number: char {} are invalid for radix {}",
```
jeaye, line 803: Let's clarify the error message here and fix the grammar by saying `invalid number: chars '{}' are invalid for radix {}`.

jianlingzhong: updated

compiler+runtime/src/cpp/jank/read/lex.cpp
jianlingzhong: updated to a static cast instead.

compiler+runtime/test/cpp/jank/read/lex.cpp
jianlingzhong: this one actually is not out of range.

compiler+runtime/test/cpp/jank/read/lex.cpp
jianlingzhong: updated to an error instead.

jeaye:
Great work, Jianling!

PR #145, opened 2024-12-12
Feature: Integer Range (Clojure's LongRange)

### LINKED ISSUE #135: Create integer_range
Clojure JVM has separate range implementations for integers (Long, in JVM parlance) and anything else. The main difference is that `integer_range` (`LongRange`, in JVM parlance) would implement the `countable` behavior (`Counted`, in JVM parlance). This allows for O(1) counting, compared to O(n). You can see the difference here very clearly:

```clojure
user=> (time (count (range 0 100000.0)))
"Elapsed time: 16.55396 msecs"
100000
user=> (time (count (range 0 100000)))
"Elapsed time: 0.056999 msecs"
```

jank's `integer_range` should do the same and optimize for integers wherever possible.

stmonty:
Fixes #135 

Implements integer-range for ranges that have all integer parameters.
Notably, implements the 'countable' behavior for faster counting.

Todo: We discussed if integer-range's should hold onto their count as a member like in Clojure's JVM version. I can still implement this.

Some initial benchmarking shows that 'count' is much faster on integer-range, which is to be expected. Notably, the sum-even function benchmark shows a big discrepancy. Maybe this is because of the lack of chunking / chunkable behavior?

```clojure
;; The function used for last benchmarks
(defn sum-even [rng]
  (->> rng
       (filter even?)
       (reduce +)))

;; (sum-even (range 1000)
```

|               ns/op |                op/s |    err% |          ins/op |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|---------------:|--------:|----------:|:----------
|                0.00 |          808,751.50 |    0.4% |        4,852.32 |         940.30 |    0.6% |      0.01 | `Create Range`
|                0.00 |          966,250.42 |    2.1% |        4,035.40 |         783.76 |    0.6% |      0.01 | `Create Integer-Range`
|                3.14 |              318.15 |    0.2% |   16,452,765.57 |   2,916,268.59 |    0.2% |      0.77 | `Range Count`
|                0.00 |          580,015.16 |    0.6% |        6,737.07 |       1,304.10 |    1.2% |      0.01 | `Integer-Range Count`
|                3.16 |              316.91 |    0.0% |   16,454,158.10 |   2,916,528.24 |    0.2% |      0.77 | `Range Count w/ Start`
|                0.00 |          479,403.86 |    0.9% |        8,311.59 |       1,608.00 |    0.8% |      0.01 | `Integer-Range Count w/ Start`
|                0.67 |            1,503.25 |    1.4% |    3,324,459.80 |     597,542.25 |    0.2% |      0.17 | `Range Count w/ Start and Step`
|                0.00 |          415,022.64 |    1.0% |        9,817.09 |       1,883.11 |    0.7% |      0.01 | `Integer-Range Count w/ Start and Step`
|                5.07 |              197.37 |    0.2% |   18,904,369.00 |   3,568,382.61 |    1.8% |      1.23 | `Sum-Even Range`
|               34.46 |               29.02 |    0.1% |  128,385,974.91 |  24,370,527.91 |    3.7% |      8.34 | `Sum-Even Integer-Range`


compiler+runtime/include/cpp/jank/runtime/obj/integer_range.hpp
```diff
@@ -0,0 +1,96 @@
+#pragma once
+
+#include "jank/type.hpp"
```
jeaye, line 3: Please remove this.

compiler+runtime/include/cpp/jank/runtime/obj/integer_range.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <jank/runtime/object.hpp>
+#include <jank/runtime/behavior/seqable.hpp>
+#include <jank/runtime/behavior/countable.hpp>
```
jeaye, line 6: These are not needed.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+    : start{ start }
+    , end{ end }
+    , step{ step }
+    , bounds_check{ bounds_check }
+  {
+  }
+
+  object_ptr obj::integer_range::create(native_integer const end)
+  {
+    if(end)
```
jeaye, line 58: There's a bug here, with `if(end)`. Can you reason through it?

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+                                        native_integer const end,
+                                        native_integer const step)
+  {
+    if((step > 0 && end < start) || (step < 0 && start < end) || start == end)
+    {
+      return obj::persistent_list::empty();
+    }
+    /* TODO: Repeat object. */
+    //else if(is_zero(step))
+    //{ return make_box<obj::repeat>(start); }
```
jeaye, line 80: We have a repeat object.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+                                        end,
+                                        step,
+                                        step > 0 ? positive_step_bounds_check
+                                                 : negative_step_bounds_check);
+  }
+
+  obj::integer_range_ptr obj::integer_range::seq()
+  {
+    return this;
+  }
```
jeaye, line 91: This can be const.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+  native_persistent_string obj::integer_range::to_string()
+  {
+    return runtime::to_string(seq());
+  }
+
+  native_persistent_string obj::integer_range::to_code_string()
+  {
+    return runtime::to_code_string(seq());
+  }
```
jeaye, line 160: These should all be able to be const.

compiler+runtime/src/cpp/clojure/core_native.cpp
```diff
@@ -702,6 +702,27 @@ jank_object_ptr jank_load_clojure_core_native()
...
+    fn->arity_2 = [](object * const start, object * const end) -> object * {
+      return obj::integer_range::create(expect_object<obj::integer>(start)->data,
+                                        expect_object<obj::integer>(end)->data);
+    };
+    fn->arity_3 = [](object * const start, object * const end, object * const step) -> object * {
+      return obj::integer_range::create(expect_object<obj::integer>(start)->data,
+                                        expect_object<obj::integer>(end)->data,
+                                        expect_object<obj::integer>(step)->data);
+    };
+    intern_fn_obj("integer_range", fn);
```
jeaye, line 723: When you intern a var here, you're entering Clojure land. In Clojure land, we use kebab-case, not snake_case. Look at all of the other examples in this file.

compiler+runtime/src/cpp/clojure/core_native.cpp
```diff
@@ -702,6 +702,27 @@ jank_object_ptr jank_load_clojure_core_native()
...
+      return obj::integer_range::create(expect_object<obj::integer>(end)->data);
+    };
+    fn->arity_2 = [](object * const start, object * const end) -> object * {
+      return obj::integer_range::create(expect_object<obj::integer>(start)->data,
+                                        expect_object<obj::integer>(end)->data);
+    };
+    fn->arity_3 = [](object * const start, object * const end, object * const step) -> object * {
+      return obj::integer_range::create(expect_object<obj::integer>(start)->data,
+                                        expect_object<obj::integer>(end)->data,
+                                        expect_object<obj::integer>(step)->data);
```
jeaye, line 721: You're using `expect_object` here. Are you 100% sure that these fns can never be called with anything other than an integer for each arg? If they are, `expect_object` will not check and we'll have undefined behavior. At best, it will crash. At worst, it will launch missiles.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    auto const diff{ end - start };
+    auto const offset{ step > 0 ? -1 : 1 };
+
+    if((step > 0 && diff > std::numeric_limits<native_integer>::max() - step + offset)
+       || (step < 0 && diff < std::numeric_limits<native_integer>::min() - step + offset))
+    {
+      throw std::runtime_error("[Integer-Range] Overflow occurred in arithmetic");
+    }
```
jeaye, line 189: Have you tested this case? Can I see a repl session for it working?

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -1918,20 +1918,32 @@
...
+     (clojure.core-native/integer_range start end step)
+     (clojure.core-native/range start end step))))
+
+  (def nth
+   "Returns the value at the index. get returns nil if index out of
    bounds, nth throws an exception unless not-found is supplied.  nth
    also works for strings, Java arrays, regex Matchers and Lists, and,
    in O(n) time, for sequences."
-  clojure.core-native/nth)
+   clojure.core-native/nth)
```
jeaye, line 1946: Looks to me like `(def nth` was indented here, for some reason.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+  {
+    return this;
+  }
+
+  obj::integer_range_ptr obj::integer_range::fresh_seq() const
+  {
+    return make_box<obj::integer_range>(start, end, step, bounds_check);
+  }
+
+  object_ptr obj::integer_range::first() const
```
jeaye, line 98: We know the return type of this will always be `obj::integer_ptr`, so let's use that.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
+#include <jank/runtime/obj/integer_range.hpp>
+#include <jank/runtime/obj/number.hpp>
```
jeaye, line 2: No need for this include, since you include visit.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  obj::integer_range_ptr obj::integer_range::fresh_seq() const
+  {
+    return make_box<obj::integer_range>(start, end, step, bounds_check);
+  }
+
+  object_ptr obj::integer_range::first() const
+  {
+    return make_box<obj::integer>(start);
```
jeaye, line 100: Since `start` is not boxed, we need to box it every time we pull it out of the sequence. This ends up meaning we do just as many allocations as we would if it were not boxed. In some cases, we may do more? Can you reason through why that might be?

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  obj::integer_range_ptr obj::integer_range::next() const
+  {
+    if(count() < 1)
```
jeaye, line 105: What you want to check is if `count() == 0`. Why are we checking if it's `< 1` instead? Same effect, but much less clear, since the implication here might be that it could be less than zero (which it cannot be, given that it's an unsigned type).

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+    {
+      return nullptr;
+    }
+    return make_box<obj::integer_range>(start + step, end, step, bounds_check);
+  }
+
+  obj::integer_range_ptr obj::integer_range::next_in_place()
+  {
+    if(count() <= 1)
+    {
```
jeaye, line 115: In `next`, you check `count() < 1`, but in `next_in_place`, you check `count() <= 1`. Which is it you want? These fns should behave the same, excepting the mutation.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
```diff
@@ -0,0 +1,193 @@
...
+    auto const diff{ end - start };
+    auto const offset{ step > 0 ? -1 : 1 };
+
+    if((step > 0 && diff > std::numeric_limits<native_integer>::max() - step + offset)
+       || (step < 0 && diff < std::numeric_limits<native_integer>::min() - step + offset))
+    {
+      throw std::runtime_error("[Integer-Range] Overflow occurred in arithmetic");
+    }
+
+    return static_cast<size_t>((diff + offset + step) / step);
```
jeaye, line 191: Have you tested this for negative steps? i.e. stepping from 0 to -10 by -1? From what I can tell, that would be broken.

compiler+runtime/src/cpp/clojure/core_native.cpp
stmonty: These fns should only have integer args because for an integer-range to be called it uses the integer? function to make sure the function arguments are all integers.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
stmonty: I think 'end' should be positive only, so that would be an issue here because in Clojure JVM -> (range -10) = (). So fixing this to be "if (end >0)"

compiler+runtime/src/cpp/clojure/core_native.cpp
jeaye: You're exposing a public var here: `clojure.core-native/integer-range`

Sure, in the one place you're calling it, all of the inputs are boxed integers. But let's talk about good practices. Is this the assumption we want to make for all potential callers of this function? What're the pros/cons of the choice?

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
stmonty: Sorry, I think both should be count() <= 1, because a range returns a less than or equal to range, so (range 3)  = (0 1 2). If this was == 0, it would return (0, 1, 2, 3).

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
stmonty: Similar reasoning to the comment above.

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
stmonty: I tested it like this:
clojure.core=> (range 0 -9223372036854775808 -1)
Exception: [Integer-Range] Overflow occurred in arithmetic
clojure.core=> (range 9223372036854775807)
Exception: [Integer-Range] Overflow occurred in arithmetic
clojure.core=>

compiler+runtime/src/cpp/jank/runtime/obj/integer_range.cpp
stmonty: It seems to work properly:
clojure.core=> (range 0 -10 -1)
(0 -1 -2 -3 -4 -5 -6 -7 -8 -9)
clojure.core=>

compiler+runtime/src/jank/clojure/core.jank
stmonty: Formatting issue, fixed it!

compiler+runtime/include/cpp/jank/runtime/obj/integer_range.hpp
```diff
@@ -0,0 +1,92 @@
+#pragma once
+
+#include "jank/runtime/obj/number.hpp"
```
jeaye, line 3: Please remove.

jeaye:
Nice work!

Issue #139, opened 2024-11-27
Make module system more robust


jeaye:
More details to be filled in soon.

# JIT
## Add dynamic var for loaded libs
Clojure uses a `clojure.core/*loaded-libs*` dynamic var, which contains a set of loaded libs. It's initialized in `clojure.core`, like so:

```clojure
(defonce ^:dynamic
  ^{:private true
     :doc "A ref to a sorted set of symbols representing loaded libs"}
  *loaded-libs* (ref (sorted-set)))
```

We can do the same with jank but make it an atom and leave a TODO for `ref` usage. Then follow all of the usages of `*loaded-libs*` in `clojure.core` and make sure jank is doing the same work in the various functions.

## Add cyclical dep check
Clojure also uses another dynamic var, `*pending-paths*`, to track cyclical deps. It's defined like so:

```clojure
(defonce ^:dynamic
  ^{:private true
     :doc "A stack of paths currently being loaded by this thread"}
  *pending-paths* ())
```

We'll want to add the same thing, along with the `check-cyclic-dependency` function and its usages. Manually test this to ensure it's working well.

## Add transactionality
Based on my testing, Clojure will still mark a namespace as loaded if you throw an exception while requiring it. This means we don't need anything clever other than thread safety when updating the loaded libs set. Please do your own testing (check loaded libs, try to load a ns which isn't loaded, have it fail, and then check again) and ensure jank matches Clojure's behavior. The loaded libs var is private, but you can get to it like so:

```clojure
(deref (intern 'clojure.core '*loaded-libs*))
```

Again, an atom will do fine here, as far as I can tell. I think it'd be good to ask about why a ref is used here, compared to an atom, in the Clojurian Slack; we might learn something neat.

## Add ability to look up module by source
I have recently added an `origin` enum to `load_module`. It allows us to explicitly load from source or load from the latest, where latest is either binary or source depending on timestamps (timestamp checking not yet implemented).

For future functionality, it would be helpful to extract some of the behavior we have in `load_module` into a `find_module` function which also takes an `origin`. It can then return some data containing both the entry and which part of the entry should be used (based on the `origin`).

## Add reloading support
When a pass `:reload` along to require, we want to load the module from the latest origin again, even if it's already loaded. This will require passing in a flag to ignore the early exit for skipping modules which are already loaded. This also needs to work with `:reload-all`, which reloads the specified module and all of its dependencies from their latest origins. Look into how Clojure does this in the `load-lib` function.

# AOT
[...]

PR #152, opened 2024-12-24
feat(module-caching): loads module based on origin

Samy-33:
[asciicast](https://asciinema.org/a/4jjQOPDJbJsRPIrvelsQ5OAvl)

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/module/loader.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/module/loader.hpp
@@ -27,6 +27,14 @@ namespace jank::runtime::module
     latest,
   };
 
+  enum class module_type : uint8_t
+  {
+    o,
+    cpp,
+    jank,
+    cljc
+  };
+
   struct file_entry
   {
     object_ptr to_runtime_data() const;
@@ -68,6 +76,24 @@ namespace jank::runtime::module
       option<file_entry> cljc;
     };
 
+    struct find_result
+    {
+      /*
+       * All the sources for a module
+       */
+      entry sources;
+      /*
+       * On the basis of origin, source that should be loaded.
+       */
+      option<module_type> to_load;
+
+      find_result(entry const &sources, module_type const to_load)
+        : sources{ sources }
+        , to_load{ to_load }
+      {
+      }
+    };
+
     /* These separators match what the JVM does on each system. */
 #ifdef _WIN32
     static constexpr char module_separator{ ';' };
@@ -80,6 +106,8 @@ namespace jank::runtime::module
     native_bool is_loaded(native_persistent_string_view const &) const;
     void set_loaded(native_persistent_string_view const &);
 
+    string_result<find_result>
+    find(native_persistent_string_view const &module, origin const ori);
     string_result<void> load(native_persistent_string_view const &module, origin const ori);
 
     string_result<void>
--- a/compiler+runtime/include/cpp/jank/runtime/module/loader.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/module/loader.hpp
@@ -38,6 +38,8 @@ namespace jank::runtime::module
   struct file_entry
   {
     object_ptr to_runtime_data() const;
+    native_bool exists() const;
+    std::time_t last_modified_at() const;
 
     /* If the file is within a JAR, this will be the path to the JAR. */
     option<native_persistent_string> archive_path;
@@ -78,20 +80,10 @@ namespace jank::runtime::module
 
     struct find_result
     {
-      /*
-       * All the sources for a module
-       */
+      /* All the sources for a module */
       entry sources;
-      /*
-       * On the basis of origin, source that should be loaded.
-       */
+      /* On the basis of origin, source that should be loaded. */
       option<module_type> to_load;
-
-      find_result(entry const &sources, module_type const to_load)
-        : sources{ sources }
-        , to_load{ to_load }
-      {
-      }
     };
 
     /* These separators match what the JVM does on each system. */
@@ -106,8 +98,7 @@ namespace jank::runtime::module
     native_bool is_loaded(native_persistent_string_view const &) const;
     void set_loaded(native_persistent_string_view const &);
 
-    string_result<find_result>
-    find(native_persistent_string_view const &module, origin const ori);
+    string_result<find_result> find(native_persistent_string_view const &module, origin const ori);
     string_result<void> load(native_persistent_string_view const &module, origin const ori);
 
     string_result<void>
--- a/compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
@@ -1,3 +1,4 @@
+#include <boost/filesystem/operations.hpp>
 #include <regex>
 #include <iostream>
 
@@ -139,7 +140,7 @@ namespace jank::runtime::module
     }
 
     auto const &zip_entry(zf.getEntry(std::string{ entry.path }));
-    fn(zip_entry.readAsText());
+    fn(zip_entry);
   }
 
   static void register_entry(native_unordered_map<native_persistent_string, loader::entry> &entries,
@@ -317,6 +318,31 @@ namespace jank::runtime::module
       make_box(path));
   }
 
+  native_bool file_entry::exists() const
+  {
+    auto const is_archive{ archive_path.is_some() };
+    if(is_archive && !boost::filesystem::exists(native_transient_string{ archive_path.unwrap() }))
+    {
+      return false;
+    }
+    else
+    {
+      native_bool source_exists{};
+      if(is_archive)
+      {
+        visit_jar_entry(*this, [&](auto const &zip_entry) { source_exists = zip_entry.isFile(); });
+      }
+
+      return source_exists || boost::filesystem::exists(native_transient_string{ path });
+    }
+  }
+
+  std::time_t file_entry::last_modified_at() const
+  {
+    auto const source_path{ archive_path.unwrap_or(path) };
+    return boost::filesystem::last_write_time(native_transient_string{ source_path });
+  }
+
   native_bool loader::is_loaded(native_persistent_string_view const &module) const
   {
     return loaded.contains(module);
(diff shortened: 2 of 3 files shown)


compiler+runtime/include/cpp/jank/runtime/module/loader.hpp
```diff
@@ -68,6 +76,24 @@ namespace jank::runtime::module
...
+      /*
+       * On the basis of origin, source that should be loaded.
+       */
+      option<module_type> to_load;
+
+      find_result(entry const &sources, module_type const to_load)
+        : sources{ sources }
+        , to_load{ to_load }
+      {
+      }
```
jeaye, line 94: No need for this core. Aggregate initialization will work the same without it. By having it, though, you prevent aggregate initialization.

compiler+runtime/include/cpp/jank/runtime/module/loader.hpp
```diff
@@ -68,6 +76,24 @@ namespace jank::runtime::module
       option<file_entry> cljc;
     };
 
+    struct find_result
+    {
+      /*
+       * All the sources for a module
+       */
```
jeaye, line 83: Please keep these comments to one line.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -101,7 +101,7 @@ namespace jank
     using namespace jank;
     using namespace jank::runtime;
 
-    //__rt_ctx->load_module("/clojure.core", module::origin::latest).expect_ok();
+    __rt_ctx->load_module("/clojure.core", module::origin::latest).expect_ok();
```
jeaye, line 104: We can't just uncomment this unconditionally yet. This is why you saw the duplicate symbol error when compiling clojure.core.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -338,47 +339,101 @@ namespace jank::runtime::module
...
-    /* When we're compiling, we always load from source. Otherwise, we
-     * load from source if we specifically chose to do so. */
-    bool const compiling{ truthy(rt_ctx.compile_files_var->deref()) };
-    if(compiling || ori == origin::source)
+    if(ori == origin::source)
     {
       if(entry->second.jank.is_some())
       {
-        res = load_jank(entry->second.jank.unwrap());
+        return with_module_type(module_type::jank);
```
jeaye, line 353: No need for `with_module_type`. Just do:

```cpp
return { entry->second, module_type::jank };
```

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -338,47 +339,101 @@ namespace jank::runtime::module
...
     {
-      /* TODO: origin must be latest, so we need to find that. Even if there
-       * is a binary, we need to only choose it if it's as new, or newer, than
-       * the source. */
-      if(entry->second.o.is_some())
+      if(entry->second.o.is_some()
+         && (entry->second.jank.is_some() || entry->second.cljc.is_some()))
       {
-        res = load_o(module, entry->second.o.unwrap());
+        auto const o_file_path{ native_transient_string{ entry->second.o.unwrap().path } };
```
jeaye, line 365: There's a case where `path` is empty because the object file came for a JAR, right? We want to ignore object files from JARs.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -338,47 +339,101 @@ namespace jank::runtime::module
...
+          module_type = module_type::jank;
+        }
+        else
+        {
+          source_path = entry->second.cljc.unwrap().path;
+          module_type = module_type::cljc;
+        }
+
+        if(boost::filesystem::last_write_time(o_file_path)
+           >= boost::filesystem::last_write_time(source_path))
```
jeaye, line 382: Do these throw if the file no longer exists? It's possible the file has been deleted since we first learned about it. We need to be sure not to throw here. Check the docs and protect these calls accordingly. If a file has gone missing, or we can't access it, we should use the other. If both are bonked, then we can return an error.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -338,47 +339,101 @@ namespace jank::runtime::module
...
+        if(entry->second.jank.is_some())
+        {
+          source_path = entry->second.jank.unwrap().path;
+          module_type = module_type::jank;
+        }
+        else
+        {
+          source_path = entry->second.cljc.unwrap().path;
+          module_type = module_type::cljc;
+        }
```
jeaye, line 379: You're not taking into account that JAR usecase here. Each file entry has two fields (`archive_path` and `path`). Only one of them will be set, but it won't always be `path`.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -338,47 +339,101 @@ namespace jank::runtime::module
...
 
+    return err(fmt::format("no sources for registered module: {}", module));
+  }
+
+  string_result<void> loader::load(native_persistent_string_view const &module, origin const ori)
+  {
+    auto const &found_module{ loader::find(module, ori) };
+    if(found_module.is_err())
+    {
+      return err(std::move(found_module.expect_err()));
```
jeaye, line 413: There is `expect_err_move` for this.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
Samy-33: If the jank/cljc souces are in archive, how do we determine the modified time of these files?

I think we can just consider jar's modified time as the source's start time because they will only be used as a dep?

Is there any other case, I am missing?

compiler+runtime/src/cpp/main.cpp
Samy-33: Without it, compiling any module fails. The compiler can't recognize any clojure.core symbol used in the module being compiled.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
Samy-33: How will that work.

```
 'this' argument to member function 'expect_err_move' has type 'const string_result<loader::find_result>' (aka 'const result<jank::runtime::module::loader::find_result, native_persistent_string>'), but function is not marked const [member_function_call
 _bad_cvr]
```

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
Samy-33: Updated, not loading object file from jar at all.

compiler+runtime/src/cpp/main.cpp
jeaye: Yep, that makes sense. We need to add a special case for compiling `clojure.core` then, to not also load `/clojure.core`. For now, this works around a bug (the duplicate symbols), but it also makes sense just from an efficiency standpoint.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
jeaye: > I think we can just consider jar's modified time as the source's start time because they will only be used as a dep?

Exactly, just the JAR's time.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
jeaye: Right now, `found_module` is `const`, which means that you can't move from it. As the code is written now, you're calling `expect_err`, which returns an `E const  &` and then you're calling `std::move` on that, which does nothing since you can't move from an `E const &` since it's a ref to `const`.

Thus, the error using `expect_err_move` here is pointing out the actual error with you wanting to move here.

Overall, this is isn't a correct application of move, the way you have it. If you want it to be correct, use `expect_err_move`, but also, this is premature optimization in a case that my gut tells me is not worth the additional complexity.

Lastly, using `std::move` with return values is generally not a good idea unless those values can from parameters. If they're locals, you'll just get in the way of NRVO. If they're function call results, you'll either try to move something which can't be moved (like this code does) or get in the way of existing prvalues.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -338,53 +365,109 @@ namespace jank::runtime::module
...
+        return find_result{ entry->second, module_type::cljc };
       }
     }
     else
     {
-      /* TODO: origin must be latest, so we need to find that. Even if there
-       * is a binary, we need to only choose it if it's as new, or newer, than
-       * the source. */
-      if(entry->second.o.is_some())
+      if(entry->second.o.is_some() && entry->second.o.unwrap().archive_path.is_none()
```
jeaye, line 381: Let's get a comment here explaining why we're not including object files from JARs.

Samy-33: Done

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
Samy-33: TIL:
- Can't move consts
- std::move can disable NRVOs (Although I don't quite understand how NRVO works.)

I didn't understand this part:
> Lastly, using std::move with return values is generally not a good idea unless those values can from parameters

jeaye: No, that particular pattern is *not* a good idea, since `my_type &param1` doesn't convey that a move will happen. Instead, this would be better:

```cpp
my_type my_function(my_type &&param1)
{
   do_something_with_param1(param1);
   return std::move(param1);
}
```

For more info on why not to use `std::move` for returning non-params, see this:
https://stackoverflow.com/questions/19267408/why-does-stdmove-prevent-rvo-return-value-optimization

jeaye:
Great job!

Samy-33:
Handles #139: `Load binaries only if the source isn't newer`

PR #155, opened 2024-12-28
feat(modules): reload modules and compilation cache

Samy-33:
[asciicast](https://asciinema.org/a/696718)


Samy-33:
@jeaye

compiler+runtime/include/cpp/jank/jit/processor.hpp
```diff
@@ -34,11 +35,17 @@ namespace jank::jit
...
+    string_result<T> find_symbol(native_persistent_string const &name) const
     {
-      auto const sym(interpreter->getSymbolAddress(name.c_str()).get());
-      return sym.toPtr<T>();
+      if(auto symbol{ interpreter->getSymbolAddress(name.c_str()) })
+      {
+        return symbol.get().toPtr<T>();
+      }
+
+      return err("Failed to find symbol");
```
jeaye, line 48: Please use `Failed to find symbol '{}'` and format in the symbol. We can all be helping with error reporting by generating better errors. :)

compiler+runtime/include/cpp/jank/jit/processor.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <boost/filesystem/path.hpp>
```
jeaye, line 3: Why is this here? I don't see it being used in this header.

compiler+runtime/include/cpp/jank/runtime/core/munge.hpp
```diff
@@ -1,10 +1,14 @@
 #pragma once
 
+#include <regex>
```
jeaye, line 3: Because of this, everyone who includes `munge.hpp` now needs to include all of the C++ regex headers. That's another ~5k lines of C++ source. Since this header is part of `runtime/core.hpp`, it's included very often and now every single one of those (dozens of) translation units needs to pay the cost of parsing the C++ regex library, even though `munge_extra` is used only in two of them.

Instead, consider:

1. Forward declaring the regex bits you need (this is very tough, in this case, since `std::regex` is an alias of a template)
2. Moving `std::regex` into the `.cpp` file by instead taking a string for the search and turning it into a regex
3. Moving `munge_extra` into a new `munge_ext.hpp`, like I did with `seq_ext.hpp`

In this case, I would likely go with option 2 if I were writing it. However, I write all of this up to get you thinking about compile times when you're writing code. In jank, compile times aren't just compile time, either, since we will need to include C++ headers at runtime for seamless C++ interop. We want those headers to be as light as possible. But even without that, we want our project to compile as fast as possible. With where we are, I want jank to be *faster* to compile in 6 months, not slower. We will get there by applying more mindful methodologies to what we put in headers.

compiler+runtime/include/cpp/jank/runtime/ns.hpp
```diff
@@ -50,6 +50,8 @@ namespace jank::runtime
     /* TODO: Benchmark the use of atomics here. That's what Clojure uses. */
     folly::Synchronized<obj::persistent_hash_map_ptr> vars;
     folly::Synchronized<obj::persistent_hash_map_ptr> aliases;
+
+    std::atomic_uint64_t symbol_counter{};
```
jeaye, line 54: Any reason you switched from `atomic_size_t` to `atomic_uint64_t`?

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -75,9 +75,17 @@ extern "C"
     return __rt_ctx->read_string(s_obj->data);
   }
 
+  void jank_ns_set_symbol_counter(char const * const ns, uint64_t count)
```
jeaye, line 78: Please make `count` const.

compiler+runtime/include/cpp/jank/c_api.h
```diff
@@ -24,6 +24,8 @@ extern "C"
   jank_object_ptr jank_eval(jank_object_ptr s);
   jank_object_ptr jank_read_string(jank_object_ptr s);
 
+  void jank_ns_set_symbol_counter(char const * const ns, uint64_t count);
```
jeaye, line 27: Why did you change this from a `size_t` to a `uint64_t`?

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -135,6 +135,18 @@ namespace jank::codegen
     {
       auto const global_ctor_fn(ctx->global_ctor_block->getParent());
       ctx->builder->CreateCall(global_ctor_fn, {});
+
```
jeaye, line 138: Please write a comment here for the future us which explains why this whole dance is needed. It's not at all obvious.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -185,6 +186,20 @@ namespace jank::jit
     load_ir_module(std::move(ir_module), std::move(ctx));
   }
 
+  string_result<void> processor::remove_symbol(native_persistent_string const &name) const
+  {
+    auto &ee{ interpreter->getExecutionEngine().get() };
+    llvm::orc::SymbolNameSet to_remove{};
+    to_remove.insert(ee.mangleAndIntern(name.c_str()));
+    auto const error = ee.getMainJITDylib().remove(to_remove);
```
jeaye, line 194: Please always use direct initialization.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -185,6 +186,20 @@ namespace jank::jit
...
+
+    if(error.isA<llvm::orc::SymbolsCouldNotBeRemoved>())
+    {
+      return err(fmt::format("Failed to remove the symbol: {}", name));
```
jeaye, line 198: Please use the format: "Failed to remove the symbol '{}'"

compiler+runtime/src/cpp/jank/runtime/core/munge.cpp
```diff
@@ -1,3 +1,5 @@
+#include <regex>
```
jeaye, line 1: I left a comment about you including this header in `munge.hpp`, but I also want to now point out that you're *also* including it in `munge.cpp`, (and other files in this PR) which makes the problem even worse. :joy:

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -674,7 +674,8 @@ namespace jank::runtime
         {
           return typed_s->contains(key);
         }
-        if constexpr(std::same_as<S, obj::persistent_hash_set>)
+        if constexpr(std::same_as<S, obj::persistent_hash_set>
+                     || std::same_as<S, obj::persistent_sorted_set>)
```
jeaye, line 678: We want `behavior::set_like` here.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -477,15 +457,30 @@ namespace jank::runtime::module
...
   {
     profile::timer const timer{ fmt::format("load object {}", module) };
+
+    /* While loading an object, if the main ns loading symbol exists, then
+     * we don't need to load the object file again.
+     *
+     * Existence of the `jank_load_<module>` symbol (also a function),
+     * means that all the required symbols exist and are already defined.
+     * We call this symbol to re-initialize all the vars in the namespace.
+     * */
```
jeaye, line 474: So this means that it'll be impossible, in jank, to:

1. Load a module from an object file
2. Change the source behind that module
3. Try to require the module again, to reload it
4. Get the updated source

What does Clojure do here?

jeaye (changes requested):

compiler+runtime/include/cpp/jank/jit/processor.hpp
Samy-33: Clangd shows error, without this `include`

Being used at the bottom of this file.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
Samy-33: If the source is more recent than the object file, we will always load from source.

Clojure doesn't load from `class` files, while `reloading` in the repl.

compiler+runtime/include/cpp/jank/jit/processor.hpp
jeaye: Ok, not new code. Please make sure stdlib headers are first. Then dependency headers. Then jank headers.

compiler+runtime/include/cpp/jank/runtime/ns.hpp
Samy-33: I couldn't figure out how to generate IR for `size_t` while calling the `jank_set_ns_symbol_counter`.

For `atomic_uint64_t` I could use the `ctx->builder->getInt64Ty()`. But couldn't find an equivalent for size_t.

Would appreciate if you could point me to the correct place.

compiler+runtime/src/cpp/jank/runtime/core/munge.cpp
Samy-33: How does including it into cpp much worse than just including it in hpp?

I understand why including `regex` in hpp would be a problem and also including it in a `cpp` would be worse. But your comment feels like, it's extremely bad on top of just `hpp`?

Could you help me understand ?

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
Samy-33: Let me know, if this makes sense?

compiler+runtime/include/cpp/jank/runtime/ns.hpp
jeaye: Ah, I see. The IR is what led to this. Ok, we can keep it as is in order to be a better 1:1 fit with the IR.

compiler+runtime/include/cpp/jank/jit/processor.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <boost/filesystem/path.hpp>
+#include <fmt/format.h>
```
jeaye, line 4: `fmt` is not permitted within normal headers. Please use our `util/string_builder` instead. `fmt` is too heavy and I have stripped it from every normal header and replaced it with a custom string builder which is much lighter to compile.

The commit for that is here: https://github.com/jank-lang/jank/commit/23aba5503bd5c3455511bf5e4ccc9e1b50e623ea

General usage:

```cpp
return err(util::string_builder{}("Failed to find the symbol: '")(name)("'")).release();
```

Samy-33: Done

Samy-33:
Handles #139 : `Load binaries only if the source isn't newer`

PR #156, opened 2024-12-29
manual delete for PointerFreeGC is not necessary

jianlingzhong:


jeaye:
This is not the fix. Deleting is important for memory usage, even with a GC.

PR #160, opened 2025-01-11
add nix flake for building jank

skykanin:
adds a nix flake which provides a devShell
with all the packages required to build jank
(it's not neccessary to build clang/llvm yourself)


flake.nix
```diff
@@ -0,0 +1,51 @@
...
+            # Libs.
+            boehmgc
+            boost
+            cli11
+            fmt
+            immer
+            libzip
+            magic-enum
+            openssl
+            readline
```
jeaye, line 46: I think we can strip this down.

* `immer`, `fmt`, and `cli11` are all built along with jank as submodules, so should not be needed
* `magic-enum` should not be needed at all
* `readline` should not be needed at all

jeaye:
Thank you! This is great. I left one comment about trimming things down. Aside from that, would you mind explaining how this is used now, given that `shell.nix` is gone? Updating the `compiler+runtime/doc/build.md` would be best, but if not then explaining to me so I can then update the doc will also work for me.

flake.nix
skykanin: These were just copied from the old `shell.nix`. I've removed them

jeaye:
Tested this locally. Works like a charm. It needs some changes for our various configurations, which have different deps, but I'll handle those now that you've set everything up. Thanks so much!

jeaye:
@skykanin We really need to be able to build from source, since we have some local LLVM changes. I've been looking into how to override the source for:

- clang
- llvm
- llvmPackages.libclang
- llvmPackages.clang-tools

No success so far. Are you open to looking into this? This is what we need to build from (based on LLVM main, which is unreleased LLVM 21).

This is what I've been trying: https://gist.github.com/jeaye/a23006759029efcba6b1529b5b0ffc1e

PR #164, opened 2025-01-17
Fill out clojure.core fns

### LINKED ISSUE #163: Fill out clojure.core fns
We need to fill out jank's clojure.core support more. We're currently only at around 40% parity.

I'm putting out the call for someone who wants to help get us closer to 100%. To do this, we need to:

1. Identify all core fns missing from jank's clojure.core
2. Copy those fns over to jank
3. If there are any issues with compiling that fn in jank, comment out the body `#_` and replace it with a `(throw "TODO: port")`

I suspect another 25% or so should port over cleanly. The rest will need some form of changes, to address Java interop or implement new backing objects. But the main goal is to get every single one of those fns into jank's clojure.core..

## Workflow tip
Just use `./bin/watch ./bin/compile` to automatically compile `clojure.core` when you change it. Port over missing core fns, save, and see if there are any compilation issues. If so, add the TODO and move on.

djblue:
Resolves #163

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -1,5 +1,20 @@
 (def ^:dynamic *ns*)
+(def ^:dynamic *in* nil)
 (def ^:dynamic *out* nil)
+(def ^:dynamic *command-line-args* nil)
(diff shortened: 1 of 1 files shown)


jeaye:
Great job! :muscle:

PR #171, opened 2025-01-18
Fill out fn implementations

djblue:
- `abs`
- `int?`
- `double?`
- `map-entry?`
- remove `assert-args`
- `find-var`
- `eval`
- `ratio?`
- `numerator`
- `denominator`

### DIFF
--- a/compiler+runtime/src/cpp/clojure/core_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/core_native.cpp
@@ -200,6 +200,12 @@ namespace clojure::core_native
       ?: obj::nil::nil_const();
   }
 
+  static object_ptr find_var(object_ptr const sym)
+  {
+    return __rt_ctx->find_var(try_object<obj::symbol>(sym)).unwrap_or(nullptr)
+      ?: obj::nil::nil_const();
+  }
+
   static object_ptr remove_ns(object_ptr const sym)
   {
     return __rt_ctx->remove_ns(try_object<obj::symbol>(sym)).unwrap_or(nullptr)
@@ -264,6 +270,11 @@ namespace clojure::core_native
   {
     __rt_ctx->compile_module(runtime::to_string(path)).expect_ok();
     return obj::nil::nil_const();
+  }
+
+  static object_ptr eval(object_ptr const expr)
+  {
+    return expr;
   }
 }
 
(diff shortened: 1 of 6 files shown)


compiler+runtime/src/cpp/clojure/core_native.cpp
```diff
@@ -265,6 +271,11 @@ namespace clojure::core_native
     __rt_ctx->compile_module(runtime::to_string(path)).expect_ok();
     return obj::nil::nil_const();
   }
+
+  static object_ptr eval(object_ptr const expr)
+  {
+    return expr;
```
djblue, line 277: I'm struggling here with how to covert the expr to something that can be passed to [`evaluate::eval()`](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/include/cpp/jank/evaluate.hpp#L19).

jeaye: Oh, you're right. Poor direction on my part. First you'll need to analyze the object. `runtime::context::eval_string` shows how this is done. Let's add a `runtime::context::eval` which takes the `object_ptr`. It'll then use `an_prc` to analyze that object in value position. Something like:

```cpp
static object_ptr eval(object_ptr const o)
{
  auto const expr(an_prc.analyze(o, analyze::expression_position::value));
  return evaluate::eval(expr.expect_ok());
}
```

jeaye:
Everything else in this PR is spot on, for the record. :100:

compiler+runtime/src/cpp/clojure/core_native.cpp
djblue: Done, thanks @jeaye!

jeaye:
Looks great to me. Tested everything from a repl? We don't yet have jank hooked up to https://github.com/jank-lang/clojure.core-test

djblue:
@jeaye I did do some manual verification at the repl 👍 
```clojure
% ./build/jank repl
Bottom of clojure.core
clojure.core=> (abs -1)
1
clojure.core=> (int? 1)
true
clojure.core=> (double? 1.0)
true
clojure.core=> (map-entry? [:k :v])
true
clojure.core=> (find-var 'clojure.core/find-var)
#'clojure.core/find-var
clojure.core=> (eval '(+ 1 2 3))
6
clojure.core=> (ratio? 1/2)
true
clojure.core=> (numerator 1/2)
1
clojure.core=> (denominator 1/2)
2
clojure.core=> 
```

jeaye:
Brilliant. Nice work, Chris!

PR #175, opened 2025-01-20
fix: clojure.core loads twice

### LINKED ISSUE #166: `jank repl clojure.core` loads clojure.core twice
```clojure
$ jank repl clojure.core
Bottom of clojure.core
Bottom of clojure.core
clojure.core=>
```

Samy-33:
fixes #166

### DIFF
--- a/compiler+runtime/src/cpp/main.cpp
+++ b/compiler+runtime/src/cpp/main.cpp
@@ -119,6 +119,7 @@ namespace jank
       throw std::runtime_error{ "Not yet implemented: REPL server" };
     }
 
+    if(opts.target_module != "clojure.core")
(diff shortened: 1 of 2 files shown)


compiler+runtime/src/cpp/main.cpp
```diff
@@ -119,6 +119,7 @@ namespace jank
       throw std::runtime_error{ "Not yet implemented: REPL server" };
     }
 
+    if(opts.target_module != "clojure.core")
```
jeaye, line 122: Why is this necessary? If we try to `load_module` `clojure.core` twice, it should already be loaded the second time and do nothing, right?

PR #176, opened 2025-01-21
Adjust CI to handle build executables for supported platforms

elken:
Marked as draft as this has conflicts with a recent change, and I am offering these changes as a proposal. Aware I also have to re-enable tests as well. Figured I'd get your input early 😄 

If you want me to drop certain bits I can, and I think there's probably a cleaner way we can handle the various on/off states we have across runs.

Once this is merged, I have changes needed ready to go for homebrew-jank to pick this up.

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -14,48 +14,15 @@ outputs:
 runs:
   using: "composite"
   steps:
-    - name: Install brew packages
-      working-directory: ${{ github.workspace }}/compiler+runtime
-      shell: bash
-      run: |
-        if [[ "$OSTYPE" == "linux"* ]]; then
-          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
-          sudo apt-get install -y libzip-dev libbz2-dev libgc-dev
-        fi
-        HOMEBREW_NO_AUTO_UPDATE=1 brew install ninja libzip doctest boost llvm@19 double-conversion bdw-gc clang-format
-
     - name: Compile and test
       shell: bash
       working-directory: ${{ github.workspace }}/compiler+runtime
-      run: |
-        if [[ "$OSTYPE" == "darwin"* ]]; then
-            export SDKROOT=$(xcrun --sdk macosx --show-sdk-path)
-        fi
-
-        if [[ "$OSTYPE" == "linux"* ]]; then
-          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
-        fi
-
-        export PATH="$(brew --prefix)/opt/llvm/bin:${PATH}"
-        export LDFLAGS="-Wl,-rpath,$(brew --prefix)/opt/llvm/lib ${LDFLAGS}"
-        export CPPFLAGS="-I$(brew --prefix)/opt/llvm/include ${CPPFLAGS}"
-
-        ${{ github.workspace }}/compiler+runtime/bin/configure -GNinja \
-                                                               -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} #\
-                                                               #-Djank_tests=on \
-                                                               # -Djank_analysis="${ANALYSIS:-off}" \
-                                                               # -Djank_coverage="${COVERAGE:-off}"
-
-        ${{ github.workspace }}/compiler+runtime/bin/compile
-        # ${{ github.workspace }}/compiler+runtime/bin/test
-        # LLVM_PROFILE_FILE=build/test.profraw ./bin/test
-        # llvm-profdata merge --sparse build/test.profraw -o build/test.profdata
-        # llvm-cov show ./build/jank-test --instr-profile build/test.profdata > coverage.txt
+      run: ./bin/ci/test
 
     - name: Validate formatting
       shell: bash
       working-directory: ${{ github.workspace }}/compiler+runtime
-      run: ${{ github.workspace }}/compiler+runtime/bin/ci/validate-formatting
+      run: ./bin/ci/validate-formatting
 
     - if: ${{ env.CODECOV }} == "on"
       name: Upload report to Codecov
@@ -66,13 +33,4 @@ runs:
       id: release
       shell: bash
       working-directory: ${{ github.workspace }}/compiler+runtime
-      run: |
-        export OS_NAME=$(uname | tr "[:upper:]" "[:lower:]")
-        export FOLDER=jank-${OS_NAME}-$(uname -m)
-        export DESTDIR=${{ github.workspace }}/${FOLDER}
-        ${{ github.workspace }}/compiler+runtime/bin/install
-
-        cp -R ${DESTDIR} .
-
-        tar czf ${DESTDIR}.tar.gz ${FOLDER}/
-        echo "archive=${DESTDIR}.tar.gz" >> $GITHUB_OUTPUT
+      run: ./bin/ci/install
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -14,49 +14,65 @@ outputs:
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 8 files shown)


.github/actions/build/action.yml
```diff
@@ -14,49 +14,65 @@ outputs:
...
+                                                               -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} #\
+                                                               #-Djank_tests=on \
+                                                               # -Djank_analysis="${ANALYSIS:-off}" \
+                                                               # -Djank_coverage="${COVERAGE:-off}"
+
+        ${{ github.workspace }}/compiler+runtime/bin/compile
+        # ${{ github.workspace }}/compiler+runtime/bin/test
+        # LLVM_PROFILE_FILE=build/test.profraw ./bin/test
+        # llvm-profdata merge --sparse build/test.profraw -o build/test.profdata
+        # llvm-cov show ./build/jank-test --instr-profile build/test.profdata > coverage.txt
```
jeaye, line 53: This was in `bin/ci/test` so that we can easily benefit from linting it (we have a CI job for this). In general, I'm opposed to writing large amounts of bash in yaml files, so I tend to pull those into scripts.

.github/actions/build/action.yml
```diff
@@ -14,49 +14,65 @@ outputs:
...
+        export FOLDER=jank-${OS_NAME}-$(uname -m)
+        export DESTDIR=${{ github.workspace }}/${FOLDER}
         ${{ github.workspace }}/compiler+runtime/bin/install
-        tar czf $DESTDIR.tar.gz $DESTDIR/
-        echo "archive=$DESTDIR.tar.gz" >> $GITHUB_OUTPUT
+
+        cp -R ${DESTDIR} .
+
+        tar czf ${DESTDIR}.tar.gz ${FOLDER}/
+        echo "archive=${DESTDIR}.tar.gz" >> $GITHUB_OUTPUT
```
jeaye, line 78: This would be good in a `bin/ci/install` script which we can lint, too.

.gitmodules
```diff
@@ -1,24 +1,24 @@
 [submodule "third-party/vcpkg"]
-	path = compiler+runtime/third-party/vcpkg
-	url = https://github.com/jank-lang/vcpkg.git
+        path = compiler+runtime/third-party/vcpkg
+        url = https://github.com/jank-lang/vcpkg.git
 [submodule "compiler+runtime/third-party/folly"]
```
jeaye, line 4: Very odd to see changes in here.

compiler+runtime/CMakeLists.txt
```diff
@@ -257,6 +257,8 @@ include(cmake/dependency/fmt.cmake)
 include(cmake/dependency/libzippp.cmake)
 
 find_package(OpenSSL REQUIRED COMPONENTS Crypto)
+set(Boost_USE_STATIC_LIBS ON)
+set(Boost_USE_STATIC_RUNTIME OFF)
```
jeaye, line 261: What's the static runtime? Will you please put it as a comment so I can also know in 6 months when I ask myself the same thing? :joy:

compiler+runtime/bin/ci/validate-formatting
```diff
@@ -3,7 +3,11 @@
 set -euo pipefail
 
 success=true
-clang_format="./build/llvm-install/usr/local/bin/clang-format"
+if [[ -e "./build/llvm-install" ]]; then
+    clang_format="./build/llvm-install/usr/local/bin/clang-format"
+else
+    clang_format="$(command -v clang-format)"
+fi
```
jeaye, line 10: Please drop the `then` down, Allman style. Also, this file has 2 space indentation.

.github/workflows/build.yml
```diff
@@ -10,18 +10,20 @@ jobs:
       fail-fast: false
       matrix:
         # TODO: sanitize: [On, Off]
-        os: [ubuntu-24.04]
+        os: [ macos-15]
```
jeaye, line 13: For macOS, given the new matrix on main, we should probably just include a Debug + Release job. All of the analysis, codecov, and sanitization can happen on Ubuntu. We have 5 build jobs (1 lint) right now, so this should put us at 7 build jobs.

jeaye (changes requested):
This is great to see, Ellis! I think it's excellent to build for macOS, but let's keep Ubuntu using apt and use bash scripts which we can lint. I've left notes for how we can fit this into the existing build matrix, too.

.github/actions/build/action.yml
```diff
@@ -14,49 +14,65 @@ outputs:
...
-      uses: actions/cache/restore@v4
-      with:
-        path: |
-          ${{ github.workspace }}/compiler+runtime/build
-        key: ${{ runner.os }}-${{ hashFiles('compiler+runtime/bin/build-clang') }}
+      run: |
+        if [[ "$OSTYPE" == "linux"* ]]; then
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+          sudo apt-get install -y libzip-dev libbz2-dev libgc-dev
+        fi
```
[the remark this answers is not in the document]
jeaye: Note: If we do add Linux brew separately, that would be our 8th build job. Nothing wrong with that, just clarifying.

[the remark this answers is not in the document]
elken: Won't necesserily break at all no, you would just still need the deps installed which will be taken care of when we ship .deb and .rpms. Linux can load the objects from anywhere, they don't have to be installed via brew and in fact if they weren't then the LD paths will already be correctly loaded. I'll prove this next time I'm on my desktop.

The only difference in package managers is brew has a good enough version of clang which prevents us having to spend several hours building it. Once `ubuntu-24.10` images start shipping we can use those as it ships with `clang-19`.

I did try every variation of Clang I could but for some reason it always failed for weird errors.

.github/actions/build/action.yml
elken: I did debate it as it was getting quite large, no worries then will amend

.gitmodules
elken: Tabs to spaces. My editor converts them automatically on save.

compiler+runtime/bin/ci/validate-formatting
elken: Worth adding an editorconfig file, I can add also if you want

.github/workflows/build.yml
elken: I want to have a re-think on adjusting the pipeline to better streamline the steps while still handling the Debug/Release steps.

Are you sure there's no value in having macOS also do those?

.github/actions/build/action.yml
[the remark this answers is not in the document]
elken: Think I have a way around this to have the best of both worlds here, but my point above still stands about the package manager in use.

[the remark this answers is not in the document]
jeaye: Ok, gotcha. I'm confident that you know more about packaging that I do. :upside_down_face:  Just wanted to make sure we considered this. If it works and it's simpler, great!

compiler+runtime/bin/ci/validate-formatting
jeaye: Up to you if it's worth it.

.github/workflows/build.yml
jeaye: I don't think there's *no value* in macOS doing sanitization, but I'm also wary of exploding our CI jobs. If you'd like to try with all the jobs first, we can do that.

.github/actions/build/action.yml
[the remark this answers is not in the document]
elken: No worries! You're right to question it as it does add some potential confusion 😄

jeaye:
Closing this for cleanup. A lot has happened since January. I'll be adding support for macOS in CI soon.

PR #177, opened 2025-01-23
Add lein-jank test

frenchy64:
Since this broke recently I thought it would help to add a test for `lein-jank`.


lein-jank/test-project/project.clj
```diff
@@ -0,0 +1,7 @@
+(defproject test-project "0.1.0-SNAPSHOT"
+  :description "FIXME: write description"
+  :url "http://example.com/FIXME"
+  :license {:name "EPL-2.0 OR GPL-2.0-or-later WITH Classpath-exception-2.0"
+            :url "https://www.eclipse.org/legal/epl-2.0/"}
```
jeaye, line 5: Let's remove these lines, rather than updating the license to MPL (what jank uses).

lein-jank/test.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
jeaye, line 1: Please use `#!/usr/bin/env bash`, which makes it more portable. Those of us on NixOS appreciate it. :joy:

lein-jank/test.sh
```diff
@@ -0,0 +1,5 @@
...
+set -xe
```
jeaye, line 2: `set -euo pipefail` is a good default. Do we need `-x` here?

.github/actions/build/action.yml
```diff
@@ -50,6 +58,12 @@ runs:
     - if: ${{ env.CODECOV }} == "on"
       name: Upload report to Codecov
       uses: codecov/codecov-action@v3
+    - name: Test Leiningen plugin
+      shell: bash
+      working-directory: ${{ github.workspace }}/lein-jank
+      run: |
+        ${{ github.workspace }}/compiler+runtime/bin/install
+        ./test.sh
```
jeaye, line 66: Please move the script to `lein-jank/bin/ci/test` (no `.sh`) to fit our existing structure and naming.

jeaye (changes requested):
I definitely like to see us testing this workflow. Soon, we'll have jank testing the full https://github.com/jank-lang/clojure.core-test repo, which uses leiningen as well.

lein-jank/test.sh
frenchy64: We don't need -x.

lein-jank/test
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
jeaye, line 1: Will you please move this to `lein-jank/bin/ci/test`? All of our scripts are in `bin` dirs. Anything CI specific is in `bin/ci`.

frenchy64: Whoops, misread that.

lein-jank/bin/ci/test
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -euo pipefail
+lein install
+cd test-project
+lein jank run 1 | grep 'lein jank works!'
```
jeaye, line 5: Everything looks good. Just wanted to double check this, though. lein-jank is going to expect that `jank` is available on the `PATH` and I don't think it is. We'd need to export a changed `PATH` before running this, right?

frenchy64: The invariant I originally assumed was that jank would be installed before calling this script, which is why I added the install step in the workflow. Are there other similar scripts with different conventions?

jeaye: Oh, right. You're installing globally, not to a `DESTDIR`. This may just work then.

PR #179, opened 2025-01-23
Add `NaN?` and `infinite?`

djblue:


compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1666,4 +1666,26 @@ namespace jank::runtime
...
+
+  native_bool is_nan(object_ptr const o)
+  {
+    if(!is_real(o))
+    {
+      return false;
+    }
+    return visit_number_like(
+      [](auto const typed_l) -> native_bool { return std::isnan(typed_l->to_real()); },
+      o);
```
jeaye, line 1678: With this setup, you'll need to inspect the object type twice. When you visit, though, you have all of the info you need. Also, when you visit an object, you have its full type, at compile-time. That means you can just reach in and pull out the `native_real` within an `obj::real_ptr`. Check it out:

```cpp
    return visit_number_like(
      [=](auto const typed_o) -> native_bool {
        using T = typename decltype(typed_o)::value_type;
        
        if constexpr(std::same_as<T, obj::real>)
        {
          return std::isnan(typed_l->data);
        }
        else
        {
          return false;
        }
      },
      o);
```

Note that, when you visit, the typed param always matches the name of the input param, but with a `typed_` prefix. So here we're visiting `o` and thus we get `typed_o`.

djblue: Is the reason you have the type info at compile-time because the lambda gets expanded for multiple types at compile-time?

jeaye: Exactly! The lambda has an `auto` param, which is a shorthand for making it a template. When we visit, we actually instantiate the lambda for every possible object type, at compile-time. At runtime, we'll call the one matching the type. So we can use compile-time programming tricks (like `if constexpr`) to change the implementations of certain instantiations of our lambda.

djblue: Dang, that's pretty cool. I feel like I'm starting to understand things a bit more now. Thanks for taking the time to explain it!

jeaye:
Excellent job!

PR #180, opened 2025-01-24
Initial support for `tagged-literal`

djblue:
Not done yet, but just wanted to put this up to make sure I was headed in the right direction.

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <jank/runtime/object.hpp>
-#include <jank/option.hpp>
 
 namespace jank::runtime::obj
 {
@@ -27,11 +26,9 @@ namespace jank::runtime::obj
 
     object base{ obj_type };
 
-    object_ptr tag;
-    object_ptr form;
+    object_ptr tag{};
+    object_ptr form{};
 
     mutable native_hash hash{};
   };
-
-  native_bool is_tagged_literal(object_ptr o);
 }
--- a/compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
@@ -19,6 +19,12 @@ namespace jank::runtime::obj
     native_persistent_string to_code_string() const;
     native_hash to_hash() const;
 
+    /* behavior::associatively_readable */
+    object_ptr get(object_ptr const key) const;
+    object_ptr get(object_ptr const key, object_ptr const fallback) const;
+    object_ptr get_entry(object_ptr key) const;
+    native_bool contains(object_ptr key) const;
+
     object base{ obj_type };
 
     object_ptr tag;
--- a/compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
@@ -1,0 +1,31 @@
+#pragma once
+
+#include <jank/runtime/object.hpp>
+#include <jank/option.hpp>
+
+namespace jank::runtime::obj
+{
+  struct tagged_literal : gc
+  {
+    static constexpr object_type obj_type{ object_type::tagged_literal };
+    static constexpr native_bool pointer_free{ false };
+
+    tagged_literal(object_ptr tag, object_ptr form);
+
+    /* behavior::object_like */
+    native_bool equal(object const &) const;
+    native_persistent_string to_string() const;
+    void to_string(util::string_builder &buff) const;
+    native_persistent_string to_code_string() const;
+    native_hash to_hash() const;
+
+    object base{ obj_type };
+
+    object_ptr tag;
+    object_ptr form;
+
+    mutable native_hash hash{};
+  };
+
+  native_bool is_tagged_literal(object_ptr o);
+}
--- a/compiler+runtime/include/cpp/jank/runtime/visit.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/visit.hpp
@@ -48,11 +48,11 @@
 #include <jank/runtime/obj/volatile.hpp>
 #include <jank/runtime/obj/delay.hpp>
 #include <jank/runtime/obj/reduced.hpp>
+#include <jank/runtime/obj/tagged_literal.hpp>
 #include <jank/runtime/ns.hpp>
 #include <jank/runtime/var.hpp>
 #include <jank/runtime/rtti.hpp>
 #include <jank/runtime/core/to_string.hpp>
-#include <jank/runtime/obj/tagged_literal.hpp>
 
 namespace jank::runtime
 {
--- a/compiler+runtime/include/cpp/jank/runtime/visit.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/visit.hpp
@@ -52,6 +52,7 @@
 #include <jank/runtime/var.hpp>
 #include <jank/runtime/rtti.hpp>
 #include <jank/runtime/core/to_string.hpp>
+#include <jank/runtime/obj/tagged_literal.hpp>
 
 namespace jank::runtime
 {
@@ -350,6 +351,11 @@ namespace jank::runtime
       case object_type::var_unbound_root:
         {
           return fn(expect_object<var_unbound_root>(erased), std::forward<Args>(args)...);
+        }
+        break;
+      case object_type::tagged_literal:
+        {
+          return fn(expect_object<obj::tagged_literal>(erased), std::forward<Args>(args)...);
         }
         break;
       default:
--- a/compiler+runtime/src/cpp/clojure/core_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/core_native.cpp
@@ -287,11 +287,6 @@ namespace clojure::core_native
   {
     return make_box(JANK_VERSION);
   }
-
-  static object_ptr tagged_literal(object_ptr const tag, object_ptr const form)
-  {
-    return make_box<obj::tagged_literal>(tag, form);
-  }
 }
 
 jank_object_ptr jank_load_clojure_core_native()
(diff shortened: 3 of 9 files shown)


compiler+runtime/src/cpp/clojure/core_native.cpp
```diff
@@ -499,6 +504,8 @@ jank_object_ptr jank_load_clojure_core_native()
   intern_fn("hash-unordered-coll", static_cast<native_hash (*)(object const *)>(&hash::unordered));
   intern_fn("read-string", &core_native::read_string);
   intern_fn("jank-version", &core_native::jank_version);
+  intern_fn("tagged-literal", &core_native::tagged_literal);
```
djblue, line 507: Not sure where these two fns should live.

compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
```diff
@@ -0,0 +1,31 @@
...
+    /* behavior::object_like */
+    native_bool equal(object const &) const;
+    native_persistent_string to_string() const;
+    void to_string(util::string_builder &buff) const;
+    native_persistent_string to_code_string() const;
+    native_hash to_hash() const;
+
+    object base{ obj_type };
+
+    object_ptr tag;
```
djblue, line 24: Technically the tag should always be a symbol, but this was easier to get working for me.

compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
```diff
@@ -0,0 +1,31 @@
...
+namespace jank::runtime::obj
+{
+  struct tagged_literal : gc
+  {
+    static constexpr object_type obj_type{ object_type::tagged_literal };
+    static constexpr native_bool pointer_free{ false };
+
+    tagged_literal(object_ptr tag, object_ptr form);
+
+    /* behavior::object_like */
```
djblue, line 15: Still need to add behavior for `:form` and `:tag` selectors.

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    return fallback;
+  }
+
+  object_ptr tagged_literal::get(object_ptr const key) const
+  {
+    return get(key, obj::nil::nil_const());
+  }
+
+  object_ptr tagged_literal::get_entry(object_ptr /*key*/) const
```
djblue, line 85: I don't think `get_entry` and `contains` are supported by `tagged-literals` on the jvm, so not sure what to do here.

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  native_bool tagged_literal::equal(object const &o) const
+  {
+    if(o.type != object_type::tagged_literal)
+    {
+      return false;
+    }
+
+    auto const s(expect_object<tagged_literal>(&o));
+    return tag == s->tag && form == s->form;
```
djblue, line 23: This doesn't appear to be working:

```clojure
% ./build/jank repl             
Bottom of clojure.core
clojure.core=> (= (tagged-literal 'foo {}) (tagged-literal 'foo {}))
false
```

frenchy64: By looking at similar code, I'm guessing `runtime::equal` is more appropriate to compare tag and form as they are not interned.

djblue: Thanks @frenchy64, that was it! I don't know why I assumed that `==` would be overloaded instead of providing reference equality.

compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
```diff
@@ -0,0 +1,37 @@
...
+    /* behavior::associatively_readable */
+    object_ptr get(object_ptr const key) const;
+    object_ptr get(object_ptr const key, object_ptr const fallback) const;
+    object_ptr get_entry(object_ptr key) const;
+    native_bool contains(object_ptr key) const;
+
+    object base{ obj_type };
+
+    object_ptr tag;
+    object_ptr form;
```
jeaye, line 31: ALWAYS initialize pointers with `{}`. i.e. `object_ptr form{};` Otherwise, their contents is undefined.

compiler+runtime/include/cpp/jank/runtime/obj/tagged_literal.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+    mutable native_hash hash{};
+  };
+
+  native_bool is_tagged_literal(object_ptr o);
```
jeaye, line 36: This should go in `runtime/core.hpp`

compiler+runtime/include/cpp/jank/runtime/visit.hpp
```diff
@@ -52,6 +52,7 @@
 #include <jank/runtime/var.hpp>
 #include <jank/runtime/rtti.hpp>
 #include <jank/runtime/core/to_string.hpp>
+#include <jank/runtime/obj/tagged_literal.hpp>
```
jeaye, line 55: Please move this up with the other object includes.

compiler+runtime/src/cpp/clojure/core_native.cpp
```diff
@@ -287,6 +287,11 @@ namespace clojure::core_native
   {
     return make_box(JANK_VERSION);
   }
+
+  static object_ptr tagged_literal(object_ptr const tag, object_ptr const form)
```
jeaye, line 291: This should be defined in `runtime/core.cpp` and should not be static.

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
```diff
@@ -0,0 +1,99 @@
...
+      return hash;
+    }
+
+    return hash = hash::combine(hash::visit(tag), hash::visit(form));
+  }
+
+  object_ptr tagged_literal::get(object_ptr const key, object_ptr const fallback) const
+  {
+    object_ptr _tag = __rt_ctx->intern_keyword("tag").expect_ok();
+    object_ptr _form = __rt_ctx->intern_keyword("form").expect_ok();
```
jeaye, line 65: Please make these `const`. Also, call them `tag_kw` or something, rather than `_tag`. Also, just use `auto` when the type isn't needed. Also, always use direct initialization. i.e.

```cpp
auto const tag_kw{ __rt_ctx->intern_keyword("tag").expect_ok() };
```

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
jeaye: `get_entry` returns a vector of key and value. Something like:

```cpp
return make_box<persistent_vector>(std::in_place, tag, form);
```

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  object_ptr tagged_literal::get_entry(object_ptr /*key*/) const
+  {
+    return obj::nil::nil_const();
+  }
+
+  native_bool tagged_literal::contains(object_ptr /*key*/) const
+  {
+    return false;
```
jeaye, line 92: We can check if the key is equal to the tag here.

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
```diff
@@ -0,0 +1,99 @@
+#include <jank/runtime/obj/symbol.hpp>
+#include <jank/runtime/core/to_string.hpp>
+#include <jank/runtime/visit.hpp>
+#include <jank/runtime/context.hpp>
+#include <jank/runtime/obj/tagged_literal.hpp>
```
jeaye, line 5: `visit.hpp` is a *very* heavy include. You're not using it here, either. Please trim these down to what you're using. Also put `tagged_literal.hpp` first, unless you have a good reason not to.

jeaye (changes requested):
For having no intro into jank's structure, this is a great PR for a new object. :)

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
djblue: After playing with it for a bit, I think it makes more sense to me for `contains?` to be true for `:tag` and `:form` since:
```clojure
clojure.core=> (:tag (tagged-literal 'foo {}))
foo
clojure.core=> (:form (tagged-literal 'foo {}))
{}
```

djblue:
If `codecov` isn't a hard requirement and you're okay with how I implemented `tagged_literal::contains` and `tagged_literal::get_entry`, this PR should be ready.

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    return fallback;
+  }
+
+  object_ptr tagged_literal::get(object_ptr const key) const
+  {
+    return get(key, obj::nil::nil_const());
+  }
+
+  object_ptr tagged_literal::get_entry(object_ptr key) const
```
jeaye, line 84: Please make `key` const here and in `contains`.

jeaye:
> If `codecov` isn't a hard requirement and you're okay with how I implemented `tagged_literal::contains` and `tagged_literal::get_entry`, this PR should be ready.

All good, after those params become const.

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
djblue: I'm not sure how I keep messing this up 🤦

jeaye: Most C++ devs don't even think about that. I'm just a stickler for const-correctness and have trained my eye for it. You'll get there!

jeaye:
Great work!

Issue #173, opened 2025-01-18
Metadata isn't being preserved on conj


djblue:
In jank:
```clojure
% ./build/jank repl
Bottom of clojure.core
clojure.core=> (meta ^:meta {})
{:meta true}
clojure.core=> (meta (conj ^:meta {} [:k :v]))
nil
clojure.core=> 
```

In clj:
```clojure
% clj
Clojure 1.12.0
user=> (meta ^:meta {})
{:meta true}
user=> (meta (conj ^:meta {} [:k :v]))
{:meta true}
user=> 
```

I haven't investigated further, but this might apply to other operations and collection types.

jeaye:
Yep, you're right about this. Meta preservation is likely missing in a lot of places, since metadata was implemented pretty late. We do have all of the behaviors for it now, though. Just need to identify where it's needed.

shantanu-sardesai:
@jeaye 
Should this one be closed?

```jank
user=> (meta (conj ^:meta {} [:k :v]))
{:meta true}
```

And I see Chris already worked on it in one of his previous PR.

shantanu-sardesai:
I suppose this also could do with a few new test cases in the test suite?

jeaye:
Yes, we should make sure that meta preservation is tested as well. This is something we're lacking, for coverage.

shantanu-sardesai:
It looks like this still requires some work after all:

```jank
user=> (let [meta-data {:foo 42} \
user=>...    apply-meta #(-> % (with-meta meta-data) (conj [:k :v]) meta)] \
user=>... (apply-meta '()))
nil
```

PR #184, opened 2025-01-25
Preserve (hash/array)map meta on assoc, dissoc and conj

djblue:
I was playing around with solving #173 and this is what I came up with. I still need to do the other collection types and their various operations, but wanted to get some feedback on the approach before I make too many changes.

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/obj/persistent_hash_map.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/persistent_hash_map.hpp
@@ -27,7 +27,8 @@ namespace jank::runtime::obj
     persistent_hash_map() = default;
     persistent_hash_map(persistent_hash_map &&) noexcept = default;
     persistent_hash_map(persistent_hash_map const &) = default;
-    persistent_hash_map(runtime::detail::native_persistent_array_map const &m,
+    persistent_hash_map(option<object_ptr> meta,
+                        runtime::detail::native_persistent_array_map const &m,
                         object_ptr key,
                         object_ptr val);
     persistent_hash_map(value_type &&d);
--- a/compiler+runtime/include/cpp/jank/runtime/obj/persistent_hash_map.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/persistent_hash_map.hpp
@@ -17,6 +17,10 @@ namespace jank::runtime::obj
                                        runtime::detail::native_persistent_hash_map>
   {
     static constexpr object_type obj_type{ object_type::persistent_hash_map };
+    using parent_type
+      = obj::detail::base_persistent_map<persistent_hash_map,
+                                         persistent_hash_map_sequence,
+                                         runtime::detail::native_persistent_hash_map>;
 
     using transient_type = transient_hash_map;
 
--- a/compiler+runtime/include/cpp/jank/runtime/obj/persistent_hash_map.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/persistent_hash_map.hpp
@@ -10,6 +10,7 @@ namespace jank::runtime::obj
   using persistent_array_map_ptr = native_box<struct persistent_array_map>;
   using transient_hash_map_ptr = native_box<struct transient_hash_map>;
   using persistent_hash_map_ptr = native_box<struct persistent_hash_map>;
+  using meta_option = option<object_ptr>;
 
   struct persistent_hash_map
     : obj::detail::base_persistent_map<persistent_hash_map,
@@ -29,6 +30,7 @@ namespace jank::runtime::obj
     persistent_hash_map(value_type &&d);
     persistent_hash_map(value_type const &d);
     persistent_hash_map(object_ptr meta, value_type &&d);
+    persistent_hash_map(meta_option meta, value_type &&d);
 
     template <typename... Args>
     persistent_hash_map(runtime::detail::in_place_unique, Args &&...args)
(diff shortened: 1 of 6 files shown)


compiler+runtime/include/cpp/jank/runtime/obj/persistent_hash_map.hpp
```diff
@@ -10,6 +10,7 @@ namespace jank::runtime::obj
   using persistent_array_map_ptr = native_box<struct persistent_array_map>;
   using transient_hash_map_ptr = native_box<struct transient_hash_map>;
   using persistent_hash_map_ptr = native_box<struct persistent_hash_map>;
+  using meta_option = option<object_ptr>;
```
djblue, line 13: I went with this approach since I assume we don't want to unwrap the option, only propagate it forward.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_hash_map.cpp
```diff
@@ -39,6 +39,12 @@ namespace jank::runtime::obj
     this->meta = meta;
   }
 
+  persistent_hash_map::persistent_hash_map(meta_option meta, value_type &&d)
+    : data{ std::move(d) }
+  {
+    this->meta = meta;
```
jeaye, line 45: Use this:

```cpp
: data{ std::move(d) }
, meta{ meta }
```

Also please make `meta` const.

compiler+runtime/include/cpp/jank/runtime/obj/persistent_hash_map.hpp
jeaye: Yeah, option makes sense. Let's remove the type alias, though. The others are there because they're forward declaring (the `struct` portion). Just use `option<object_ptr>` everywhere instead.

jeaye:
Couple of comments, but looks good!

compiler+runtime/src/cpp/jank/runtime/obj/persistent_hash_map.cpp
djblue: I get this error which I don't understand:
```
error: member initializer 'meta' does not name a non-static data member or base class
```

jeaye: Oh, `meta` comes in from the base. You'll need something like:

```
: obj::detail::base_persistent_map{ meta }
, data{ std::move(d) }
```

It's possible you need the [full type](https://github.com/jank-lang/jank/pull/184/files/b0b846961ce26929508221e0da368f4dd5f221df#diff-f13107755d1250573233217600a41a5c9f5ae7cb1f20f1dae0cbdd390371c45aR16) there. If so, I'd alias it under `obj_type` called `parent_type`.

We'll need a ctor in `base_persistent_map` which initilializes `meta`.

djblue: I think https://github.com/jank-lang/jank/pull/184/commits/b892c52db3155523dc412a5664315a2544e56059 is implementing what you described, but it doesn't appear to work:
```clojure
% ./build/jank repl            
Bottom of clojure.core
clojure.core=> (def x ^:meta {:x :y})
#'clojure.core/x
clojure.core=> (meta x)
{:meta true}
clojure.core=> (meta (assoc x :x 1))
nil
clojure.core=> (meta (dissoc x :x))
nil
```

jeaye: Your changes look correct, in that commit. I would either add some logs or use a debugger to see what's going on with the meta here. Using `this->meta = meta;` may work, but that's not the way we do it in C++; we use member initializer lists. Why it's not working for you... I'm not yet sure. Maybe try a `./bin/clean` to be safe.

djblue: So after getting a debugger setup, I noticed the code paths weren't getting hit and it turns out implementing metadata preservation for `"persistent_hash_map"` isn't useful for `"persistent_array_map"` 😂 After pushing up https://github.com/jank-lang/jank/pull/184/commits/055edf8d5b6d2fb70413936df48f627d09789fd2, things are working. I think my earlier testing was done explicitly with `(hash-map)` and I forgot 🤦

djblue:
For metadata preservation, do you want all the types fixed in one PR, or can they be fixed in multiple PRs? I tend to prefer smaller more focused PRs.

jeaye:
> For metadata preservation, do you want all the types fixed in one PR, or can they be fixed in multiple PRs? I tend to prefer smaller more focused PRs.

Smaller, more focused PRs are preferred.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_hash_map.cpp
jeaye: Good catch! I didn't spot that either. :grin:

djblue:
Ok, I think this PR should be good to go then. It will preserve metadata for maps on `assoc`, `dissoc` and `conj`.

I'll work on getting a complete list of data structures and operations and and put it in #173.

jeaye:
> Ok, I think this PR should be good to go then. It will preserve metadata for maps on `assoc`, `dissoc` and `conj`.
> 
> I'll work on getting a complete list of data structures and operations and and put it in #173.

I know that CI is *always* failing right now, but please take a look at this here: https://github.com/jank-lang/jank/actions/runs/12971282509/job/36177518651?pr=184

You can reproduce them locally by configuring with `-Djank_analysis=on`.

djblue:
Ok, I think https://github.com/jank-lang/jank/pull/184/commits/278a473afa8080301fe1f140efcfbec5f2485b25 should resolve those errors.

jeaye:
Nice!

PR #186, opened 2025-01-25
Implement `parse-long` and `parse-double`

djblue:


compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1702,4 +1702,40 @@ namespace jank::runtime
...
+      [](auto const typed_o) -> native_real {
+        using T = typename decltype(typed_o)::value_type;
+
+        if constexpr(std::same_as<T, obj::persistent_string>)
+        {
+          return std::stol(typed_o->data);
+        }
+        else
+        {
+          throw erase(make_box("Expected string, got nil"));
```
djblue, line 1718: Is there a way to get the type name statically here?

jeaye: You have `T`, which is an object type. Every object type has a static `obj_type`, which is its enum value. From there, we have the constexpr `object_type_str` fn. So, this will get you a string of the object type at compile-time.

```cpp
object_type_str(T::obj_type)
```

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1702,4 +1702,40 @@ namespace jank::runtime
...
+
+  native_integer parse_long(object_ptr const o)
+  {
+    return visit_object(
+      [](auto const typed_o) -> native_integer {
+        using T = typename decltype(typed_o)::value_type;
+
+        if constexpr(std::same_as<T, obj::persistent_string>)
+        {
+          return std::stol(typed_o->data);
```
jeaye, line 1714: We want `std::stoll` here.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1702,4 +1702,40 @@ namespace jank::runtime
...
       o);
   }
+
+  native_integer parse_long(object_ptr const o)
+  {
+    return visit_object(
+      [](auto const typed_o) -> native_integer {
+        using T = typename decltype(typed_o)::value_type;
+
+        if constexpr(std::same_as<T, obj::persistent_string>)
```
jeaye, line 1712: You're visiting *every single object type*. This lambda template will be instantiated 52 times (the number of objects we have). Only one of those does anything useful. In other words visit is not what you want here.

Either check the `o->type` or use `dyn_cast`. Note that C++ doesn't allow direct initialization in this one particular syntax.

```cpp
if(auto const typed_o = dyn_cast<obj::persistent_string>(o))
{
  return std::stoll(typed_o->data);
}
else
{
  throw erase(make_box("Expected string, got <TODO: type>"));
}
```

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
djblue: Good to know, although I opted for `object_type_str(o->type)` since we dropped `visit_object`.

djblue:
I removed the `visit_object` (didn't realize how expensive it was) and resolved the type name via `object_type_str(o->type)`, so this one should also be good if the usage/import of `fmt::format` is correct.

jeaye:
These are awesome PRs, Chris! Thanks so much.

PR #187, opened 2025-01-25
Improve support of `##Inf`, `##-Inf` and `##NaN`

djblue:
```clojure
% ./build/jank repl
Bottom of clojure.core
clojure.core=> ##Inf
##Inf
clojure.core=> ##-Inf
##-Inf
clojure.core=> ##NaN
##NaN
clojure.core=> (infinite? ##Inf)
true
clojure.core=> (infinite? ##-Inf)
true
clojure.core=> (NaN? ##NaN)
true
```


compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -693,6 +695,53 @@ namespace jank::read::parse
...
+    else if(sym_result.expect_ok().unwrap().ptr->type != object_type::symbol)
+    {
+      return err(
+        error{ start_token.pos, native_persistent_string{ "value after ## must be a symbol" } });
+    }
+
+    auto const sym(expect_object<obj::symbol>(sym_result.expect_ok().unwrap().ptr));
+    auto const sym_end(sym_result.expect_ok().unwrap().end);
+
+    native_real n;
```
jeaye, line 723: Always zero-initialize primitives. Otherwise, their values are undefined. Even if you're initializing it in every branch below now, if those branches change later, we'll be in for undefined behavior.

jeaye (changes requested):
Nice! It's impressive how well you figure our jank's systems on your own.

We'll need tests for this. We can do this using jank's JIT test runner. That just involves adding `.jank` files to `test/jank`. Take a look at `test/jank/reader-macro/var-quote` for examples. Maybe we'll make a `test/jank/reader-macro/symbolic-value`. Test files need to start with `pass-` or `fail-`. The former needs to result in `:success`.

compiler+runtime/src/cpp/jank/read/parse.cpp
djblue: Good point, I wasn't thinking about how the code could change later 👍

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -720,7 +720,7 @@ namespace jank::read::parse
     auto const sym(expect_object<obj::symbol>(sym_result.expect_ok().unwrap().ptr));
     auto const sym_end(sym_result.expect_ok().unwrap().end);
 
-    native_real n;
+    auto n = 0.0;
```
jeaye, line 723: Please use `native_real n{};`. *Always* use direct initialization. You could use `auto n{ 0.0 };` but you specifically want a `native_real` here.

compiler+runtime/test/jank/reader-macro/symbolic-value/pass-inf.jank
```diff
@@ -0,0 +1,3 @@
+(assert (infinite? ##Inf))
```
jeaye, line 1: What about some other failure cases? Off the top of my head:

* `##Infinite`
* `##INF`
* `###Inf`
* `##NAN`
* What about `-##NaN`?

djblue: I think `-##NaN` would go down a different code path, but I can still add it if you like.

djblue:
If you are okay without adding `-##NaN`, this PR should be good.

jeaye:
Great work!

PR #188, opened 2025-01-25
Implement `sorted?`

djblue:


compiler+runtime/include/cpp/jank/runtime/core.hpp
```diff
@@ -83,4 +83,6 @@ namespace jank::runtime
   object_ptr get_thread_bindings();
 
   object_ptr force(object_ptr o);
+
+  native_bool is_sorted(object_ptr o);
```
jeaye, line 87: This should be in `runtime/seq.{h,c}pp`. Otherwise works for me.

djblue:
Moved the fn to `seq.{h,c}pp` and merged with main. Should be good now.

jeaye:
Looks good!

PR #189, opened 2025-01-25
Implement `ex-message`, `ex-data` and `ex-cause`

djblue:


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -33,9 +33,29 @@
   "Create an instance of ExceptionInfo, a RuntimeException subclass
    that carries a map of additional data."
   ; TODO: Arity 3
```
jeaye, line 35: We can remove the TODO here.

djblue:
Removed the TODO, so this should be good to go.

PR #190, opened 2025-01-26
Fix `with-bindings` for persistent_array_map

djblue:
I considered implementing the binding mechanics in terms of `persistent_array_map` but wasn't sure it was worth it.

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -4405,7 +4405,7 @@
   Pops the installed bindings after f returned. Returns whatever f returns."
   [binding-map f & args]
   (push-thread-bindings (if (= "persistent_array_map" (type binding-map))
-                          (into {} binding-map)
+                          (into (hash-map) binding-map)
                           binding-map))
   (try
     (apply f args)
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -4404,7 +4404,9 @@
   values as thread-local bindings. Then calls f with the supplied arguments.
   Pops the installed bindings after f returned. Returns whatever f returns."
   [binding-map f & args]
-  (push-thread-bindings binding-map)
+  (push-thread-bindings (if (= "persistent_array_map" (type binding-map))
+                          (into {} binding-map)
```
jeaye, line 4408: Is this meant to put it into a hash map? `(type {})` is also `"persistent_array_map"`. Could use `(hash-map)`.

djblue: Yes, I'm trying to convert all input maps to a `"persistent_hash_map"`. In my local testing `(type (into {} {:x :y})) => "persistent_hash_map"` but maybe that's an implementation detail we shouldn't leverage here. Are you thinking something like `(into (hash-map) {:x :y})` is better here?

jeaye: Yep, I think that's safer and more explicit. I'm surprised `(into {} ...)` gives a hash map.

djblue: Done 👍

jeaye:
Nice work!

PR #193, opened 2025-01-28
Add uuid support

djblue:


compiler+runtime/include/cpp/jank/runtime/obj/uuid.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <boost/uuid/uuid.hpp>
```
djblue, line 3: I tried to forward declare the type I needed here but couldn't figure out how.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -742,6 +744,46 @@ namespace jank::read::parse
     return object_source_info{ wrapped, start_token, sym_end };
   }
 
+  processor::object_result processor::parse_reader_macro_tagged()
+  {
+    auto const token((*token_current).expect_ok());
+    auto const tag(boost::get<native_persistent_string_view>(token.data));
```
djblue, line 750: I really struggled to get the tag here, not sure if I did it correctly.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -600,4 +600,32 @@ namespace jank::runtime
     return o->type == object_type::tagged_literal;
   }
 
+  object_ptr parse_uuid(object_ptr const o)
```
djblue, line 603: `parse-uuid` on the jvm is odd in that it will throw if the input value is the wrong type, but will return nil if the string is invalid. However, using `#uuid "invalid-string"` will throw an exception. Not sure how much of the exception behavior we want to mirror of the jvm, but I tried to stay close.

compiler+runtime/include/cpp/jank/runtime/obj/uuid.hpp
```diff
@@ -0,0 +1,30 @@
...
+    /* behavior::object_like */
+    native_bool equal(object const &) const;
+    native_persistent_string to_string() const;
+    void to_string(util::string_builder &buff) const;
+    native_persistent_string to_code_string() const;
+    native_hash to_hash() const;
+
+    object base{ obj_type };
+
+    boost::uuids::uuid value{};
```
djblue, line 26: I went with this representation as I assume boost has some internal compact representation, however I did not verify. Also, not sure what to call this field.

djblue: Internal storage: 128-bit data stored as a byte array  - according to google 😂

djblue: Actually I can just goto definition 🤦 
```cpp
using repr_type = std::uint8_t[ 16 ];
```

compiler+runtime/test/jank/reader-macro/uuid/pass-random-uuid.jank
```diff
@@ -0,0 +1,3 @@
+(assert (uuid? (random-uuid)))
```
jeaye, line 1: The tests here are only for the compiler, not the runtime. We have a whole other repo of tests for the runtime (clojure.core-test). We can remove this test and change the one above it to not check `uuid?`. I think we can have more negative cases, though, such as:

* Empty string
* Valid UUID string with extra characters at the end

compiler+runtime/src/cpp/jank/runtime/obj/uuid.cpp
```diff
@@ -0,0 +1,67 @@
...
+    if(o.type != object_type::uuid)
+    {
+      return false;
+    }
+
+    auto const s(expect_object<uuid>(&o));
+    return s->value == value;
+  }
+
+  static void to_string_impl(boost::uuids::uuid value, util::string_builder &buff)
```
jeaye, line 34: `boost::uuids::uuid` is a struct type, so please pass it by `const &` to avoid copying.

compiler+runtime/include/cpp/jank/runtime/obj/uuid.hpp
jeaye: This is just fine. If the uuid header shows up as a heft include, we can later change the struct to store something opaque and only include boost's uuid in the cpp file. At this point, I don't think we can do that without affecting runtime perf.

compiler+runtime/include/cpp/jank/runtime/obj/uuid.hpp
jeaye: We need to store a value of this type in our struct, so we need the definition. We can only get away with a forward decl when we're dealing with only refs and pointers, since they're always the same size.

compiler+runtime/include/cpp/jank/runtime/obj/uuid.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+    mutable native_hash hash{};
+  };
```
jeaye, line 29: Every object `foo` needs a `foo_ptr` alias after it. Check with the other object types for how it works.

jeaye (changes requested):
Looks great. Some minor comments. The only major comment is this: you're missing something. :)

Please add a test which creates an anonymous fn which returns a uuid and calls the fn immediately. You'll find it fails. The reason is that we need IR gen for this new object, since we want to carry it into IR land.

To do that, we need to:

* Add a new uuid fn to `c-api.h`
* Add a `gen_global` overload for `uuid_ptr`
* Verify your test passes

djblue:
@jeaye I think a problem with https://github.com/jank-lang/jank/pull/193#pullrequestreview-2581989844 is that the reader is open, users can provide their own fns and return any object they want. In those instances, how would it work if the type isn't part of the the `c-api.h`?

compiler+runtime/include/cpp/jank/runtime/obj/uuid.hpp
jeaye: Note that boost has been removed as a dependency. For this, we can use stduuid: https://github.com/mariusbancila/stduuid

We'll need to bring it in as a submodule, under `compiler+runtime/third-party`.

jeaye:
@djblue Any interest in pushing this across the finish line, or are you done with it?

djblue:
Hi @jeaye, I don't think I can get this over the line. I think the scope of this PR is a little larger than I have bandwidth for currently, sorry.

jeaye:
> Hi @jeaye, I don't think I can get this over the line. I think the scope of this PR is a little larger than I have bandwidth for currently, sorry.

All good, my dude. Thanks for getting it this far! I'll make sure it gets finished.

Issue #197, opened 2025-01-29
`def` meta should be evaluated and support arbitrary values


frenchy64:
At least if we want Clojure parity, not sure if CLJS does this. `clojure.test/deftest` uses this.

```clojure
;jank
clojure.core=> (def ^{:test (fn [])} a)
#'clojure.core/a
clojure.core=> (-> #'a meta :test)
(fn [])

;clj
user=> (def ^{:test (fn [])} a)
#'user/a
user=> (-> #'a meta :test)
#object[user$fn__134 0x4a05d8ae "user$fn__134@4a05d8ae"]
```

Can't set manually with `alter-meta!` either:

```
Exception: invalid meta: clojure.core/clojure_core-fn-4375 (jit_function@0x1117d2008)
```

jeaye:
Yep, we should follow Clojure here.

PR #200, opened 2025-01-29
clojure.test

frenchy64:
Workarounds noted in comment at the top of `clojure/test.jank`.

Needed `clojure.core/ns-interns` from clojure.core, but I had to add the function name to every `(throw "TODO: port")` statement to figure that out.


compiler+runtime/test/bash/module/clojure-test/pass-test
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
jeaye, line 1: 1. You've put these tests under the `module` dir, but they are unrelated to module loading; moving this up to be `test/bash/clojure.test` should be fine.
2. Why use `clojure-test` instead of `clojure.test`?

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -4094,7 +4094,7 @@
   c. Returns true or false"
   [c x]
   ;; (. c (isInstance x))
-  (throw "TODO: port"))
+  (throw "TODO: port instance?"))
```
jeaye, line 4097: It's fine to change all of these, but it wasn't needed to find your error. For that, I'd just hop into gdb and tell it to stop when an exception is thrown.

```
gdb --args ./build/jank run foo.jank
> catch throw
> run
... wait for an exception to be thrown and gdb will stop ...
> backtrace
... see where you are ...
```

compiler+runtime/src/jank/clojure/test.jank
```diff
@@ -0,0 +1,843 @@
...
+;; by Stuart Sierra
+;; March 28, 2009
+
+;; Thanks to Chas Emerick, Allen Rohner, and Stuart Halloway for
+;; contributions and suggestions.
+
+;; jank notes:
+;; - use atom instead of ref in *report-counters*
+;; - workaround https://github.com/jank-lang/jank/issues/199
+;;   - wrote -workaround-var->ns to translate var to qualified symbol
```
jeaye, line 20: This isn't needed anymore, right?

jeaye (changes requested):

compiler+runtime/src/jank/clojure/test.jank
frenchy64: Correct, now we just need `-workaround-var->ns` to workaround #197 and its own implementation does require using a workaround :)

```clojure
 (defn -workaround-var->ns [v]
-  (-> v str (subs 2) symbol namespace symbol the-ns))
+  (-> v symbol namespace symbol the-ns))
```

compiler+runtime/test/bash/module/clojure-test/pass-test
frenchy64: That makes a lot more sense, I misunderstood the convention.

compiler+runtime/src/jank/clojure/core.jank
frenchy64: Aha, so `(byte 1)` will give `#2  0x00007fffe8df885d in clojure_core_byte_3074_1 ()`.

Very instructive thanks.

frenchy64:
Putting this back to draft actually, `clojure.test/are` doesn't work yet.

frenchy64:
Opened and worked around https://github.com/jank-lang/jank/issues/237

WDYT about the approach here of gradually adding the clojure-test suite? Just requires a deps.edn to be added to clojure-test: `{:paths ["test"]}`.

compiler+runtime/src/jank/clojure/core.jank
jeaye: Not quite. `clojure_core_byte_3074_1` means `clojure.core/byte <gensym count> <arity 1>`.

jeaye:
> Opened and worked around #237
> 
> WDYT about the approach here of gradually adding the clojure-test suite? Just requires a deps.edn to be added to clojure-test: `{:paths ["test"]}`.

I think we can gradually add for now, but note that the clojure-test suite is specifically designed to support missing vars. It will skip them. So, we should just `#_` out our unimplemented defns and then run all tests.

Syntactically, I *think* the only thing we're missing is big number support. Will need to get someone on that.

compiler+runtime/src/jank/clojure/core.jank
frenchy64: Ok. I pointed it out because AFAICT this is the only frame that gives a hint of which function needs implementing. Is there a way to configure source mapping? e.g.,

```clojure
;;foo.jank
(ns foo)
(defn -main [] (byte 1)) 
```

```cpp
#8  0x00007ffff33850e0 in foo.main_1_0 >>here>> at foo.jank:2 <<here<< ()
```

jeaye: There will be, but it's a *large* feature which involves a lot of LLVM IR work. Perhaps I know the right guy for it, though. I'll reach out to him.

frenchy64:
I think this is ready to go. I tried loading and testing every clojure-test namespace and documented their failures.

This needs to go in first: https://github.com/jank-lang/clojure-test/pull/71

Then we need to update the clojure-test dependency sha to point to the official repo.

compiler+runtime/test/bash/clojure-test-suite/deps.edn
```diff
@@ -0,0 +1,4 @@
+{:paths ["src"]
+ :deps {io.github.jank-lang/clojure-test {;; TODO update to official repo
+                                          :git/url "https://github.com/frenchy64/clojure-test.git"
+                                          :git/sha "d5c1a319251b25e3afd3acd07648c3950f882fa3"}}}
```
jeaye, line 4: We're using submodules for everything right now. I'd prefer to stick with that. Let's add the submodule here. I've also renamed that repo to clojure-test-suite, so we'll have `clojure-test-suite/clojure-test-suite`, but that's ok.

compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
```diff
@@ -0,0 +1,137 @@
+(ns jank-test.run-clojure-test-suite
+  (:require [clojure.test :as t]))
+
+;; most of these are blocked by parsing of #?(:clj Long/MAX_VALUE)
+;; https://github.com/jank-lang/jank/issues/195
+(def namespaces
```
jeaye, line 6: I really appreciate the rigor put into this.

jeaye (changes requested):
Looks great. Just one comment about how to handle the clojure-test-suite dep.

compiler+runtime/test/bash/clojure-test-suite/deps.edn
frenchy64: Done.

compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
frenchy64: Thanks for saying that :)

jeaye:
Great to see this! :muscle:

frenchy64:
Nice!

PR #201, opened 2025-01-29
Implement 9 of 21 clojure.string vars

frenchy64:
Implements:
- `clojure.string/blank?`
- `clojure.string/capitalize`
- `clojure.string/ends-with?`
- `clojure.string/escape`
- `clojure.string/includes?`
- `clojure.string/lower-case`
- `clojure.string/reverse`
- `clojure.string/starts-with?`
- `clojure.string/upper-case`


djblue:
Very cool! If you start looking at regex, I started some of that work [here](https://github.com/jank-lang/jank/compare/main...djblue:jank:regex?expand=1). I'm still looking through various performance characteristic of different regex libraries.

compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -4,320 +4,318 @@
...
+;  ;       result)))
+;  ;  ([s value from-index]
+;  ;   (let [result (if (instance? Character value)
+;  ;                  (.lastIndexOf (.toString s) (.charValue value) (unchecked-int from-index))
+;  ;                  (.lastIndexOf (.toString s) value (unchecked-int from-index)))]
+;  ;     (if (= result -1)
+;  ;       nil
+;  ;       result))))
+;  
 (defn starts-with?
```
djblue, line 304: In `core.jank`, the pattern I've noticed is the use a def here instead of a defn. I feel like it was intentional to avoid the extra fn, but not sure.

jeaye: You're correct. We already have a lot of indirection. We don't want even more.

compiler+runtime/test/bash/module/clojure-string/src/clojure_string/self_test.jank
```diff
@@ -0,0 +1,55 @@
+(ns clojure-string.self-test
```
jeaye, line 1: Please put these tests in the https://github.com/jank-lang/clojure-test repo instead. We should have one place for testing all of the official Clojure namespaces.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -295,6 +296,7 @@ try
   __rt_ctx = new(GC) runtime::context{ opts };
 
   jank_load_clojure_core_native();
+  jank_load_clojure_string_native();
```
jeaye, line 299: It's fine to put this here.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include <fmt/format.h>
+
+#include <jank/native_persistent_string/fmt.hpp>
+#include <clojure/core_native.hpp>
+#include <jank/runtime/convert.hpp>
+#include <jank/runtime/core.hpp>
+#include <jank/runtime/context.hpp>
+#include <jank/runtime/behavior/callable.hpp>
+#include <jank/runtime/visit.hpp>
```
jeaye, line 11: You're not using any visting fns, while this is an incredibly expensive include. Please audit all of the includes here to ensure they're needed.

jeaye (changes requested):

compiler+runtime/test/bash/module/clojure-string/src/clojure_string/self_test.jank
frenchy64: https://github.com/jank-lang/clojure-test/pull/70

compiler+runtime/src/cpp/clojure/string_native.cpp
frenchy64: Still working on this.

compiler+runtime/test/bash/module/clojure-string/src/clojure_string/self_test.jank
frenchy64: Would you accept these tests with this disclaimer? Every test namespace fails use to #226. 

(Line 4)

```
;;TODO delete this namespace and use clojure-test-suite
;;https://github.com/jank-lang/jank/issues/226
```

jeaye: That is a critical bug which will be fixed shortly. It's already assigned to someone. We don't benefit from having these tests between now and then. I would rather we not duplicate tests from clojure--test-suite at all.

We'll get there, so that we can run the whole test suite. We don't want to accrue a bunch of duplication along the way.

frenchy64: Sounds good, removed.

jeaye:
Great work, Ambrose! This is a good chunk of `clojure.string`. :muscle:

frenchy64:
Hell yeah!

PR #202, opened 2025-01-29
Close #199: symbol on var should be qualified

### LINKED ISSUE #199: `symbol` on var should include namespace
```
; jank
clojure.core=> (symbol #'+)
+

; clojure
user=> (symbol #'+)
clojure.core/+
```

frenchy64:
Closes #199

### DIFF
--- a/compiler+runtime/test/jank/form/var/pass-symbol-on-var.jank
+++ b/compiler+runtime/test/jank/form/var/pass-symbol-on-var.jank
@@ -1,0 +1,3 @@
+(assert (= 'clojure.core/+ (symbol #'+)))
+
+:success
--- a/compiler+runtime/test/jank/form/var/pass-symbol-on-var.jank
+++ b/compiler+runtime/test/jank/form/var/pass-symbol-on-var.jank
@@ -1,3 +1,0 @@
-(assert (= 'clojure.core/+ (symbol #'+)))
-
-:success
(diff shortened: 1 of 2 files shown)


djblue:
I didn't even know you could do this, I always dig out the name via metadata 😂

compiler+runtime/test/jank/form/var/pass-symbol-on-var.jank
```diff
@@ -0,0 +1,3 @@
+(assert (= 'clojure.core/+ (symbol #'+)))
+
+:success
```
jeaye, line 3: These tests are for the compiler (lexer, parser, analysis, eval). The bug was in `clojure.core/symbol`, so the test should be in clojure.core-test.

Looks like we don't have a case there for vars. Up for adding it? https://github.com/jank-lang/clojure.core-test/blob/main/test/clojure/core_test/symbol.cljc

jeaye (changes requested):
If you add `Closes #199` or `Fixes #199` to the PR description, the two will be linked and the PR will automagically close the ticket when merged.

compiler+runtime/test/jank/form/var/pass-symbol-on-var.jank
frenchy64: Sure I can do that. IIUC this PR will then not include any test validating the fix.

frenchy64: https://github.com/jank-lang/clojure.core-test/pull/66

jeaye:
Thanks, Ambrose!

PR #227, opened 2025-02-01
String associatively_readable implementation

### LINKED ISSUE #225: `get` does not respond to strings
```clojure
;; jank
clojure.core=> (get "a" 0)
nil

;; clojure
user=> (get "a" 0)
\a
```

frenchy64:
Closes #225 

Followed Clojure on the implementation of `get_entry`, which throws. CLJS returns nil on `(find "a" 0)`.

```clojure
clojure.core=> (get "a" 0)
\a
clojure.core=> (contains? "a" 0)
true
clojure.core=> (find "a" 0)
Exception: get_entry not supported on string
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
@@ -72,25 +72,27 @@ namespace jank::runtime::obj
   {
     if(key->type == object_type::integer)
     {
-      auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
-      if(data.size() <= i)
+      auto const i(expect_object<integer>(key)->data);
+      if(i < 0 || data.size() <= static_cast<size_t>(i))
       {
-        return nil::nil_const();
+        return fallback;
       }
-      return make_box<runtime::obj::character>(data[i]);
+      return make_box<character>(data[i]);
     }
-    return fallback;
+    else
+    {
+      return fallback;
+    }
   }
 
   native_bool persistent_string::contains(object_ptr const key) const
   {
     if(key->type == object_type::integer)
     {
-      auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
-      return 0 <= i && i < data.size();
+      auto const i(expect_object<integer>(key)->data);
+      return 0 <= i && static_cast<size_t>(i) < data.size();
     }
-    throw std::runtime_error{ fmt::format("contains? not supported on string: {}",
-                                          runtime::to_string(key)) };
+    return false;
   }
 
   object_ptr persistent_string::get_entry(object_ptr const) const
--- a/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
@@ -1,6 +1,7 @@
 #include <fmt/format.h>
 
 #include <jank/native_persistent_string/fmt.hpp>
+#include <jank/runtime/behavior/associatively_readable.hpp>
 #include <jank/runtime/obj/persistent_string.hpp>
 #include <jank/runtime/obj/persistent_string_sequence.hpp>
 #include <jank/runtime/rtti.hpp>
@@ -60,6 +61,40 @@ namespace jank::runtime::obj
   native_integer persistent_string::compare(persistent_string const &s) const
   {
     return data.compare(s.data);
+  }
+
+  object_ptr persistent_string::get(object_ptr const key) const
+  {
+    return get(key, nil::nil_const());
+  }
+
+  object_ptr persistent_string::get(object_ptr const key, object_ptr const fallback) const
+  {
+    if(key->type == object_type::integer)
+    {
+      auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
+      if(data.size() <= i)
+      {
+        return nil::nil_const();
+      }
+      return make_box<runtime::obj::character>(data[i]);
+    }
+    return fallback;
+  }
+
+  native_bool persistent_string::contains(object_ptr const key) const
+  {
+    if(key->type == object_type::integer)
+    {
+      auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
+      return data.size() <= i;
+    }
+    throw std::runtime_error{ fmt::format("contains? not supported on string: {}", runtime::to_string(key)) };
+  }
+
+  object_ptr persistent_string::get_entry(object_ptr const) const
+  {
+    throw std::runtime_error{ fmt::format("get_entry not supported on string") };
   }
 
   string_result<persistent_string_ptr> persistent_string::substring(native_integer start) const
--- a/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
@@ -1,7 +1,6 @@
 #include <fmt/format.h>
 
 #include <jank/native_persistent_string/fmt.hpp>
-#include <jank/runtime/behavior/associatively_readable.hpp>
 #include <jank/runtime/obj/persistent_string.hpp>
 #include <jank/runtime/obj/persistent_string_sequence.hpp>
 #include <jank/runtime/rtti.hpp>
--- a/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
@@ -87,7 +87,7 @@ namespace jank::runtime::obj
     if(key->type == object_type::integer)
     {
       auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
-      return data.size() <= i;
+      return 0 <= i && i < data.size();
     }
     throw std::runtime_error{ fmt::format("contains? not supported on string: {}", runtime::to_string(key)) };
   }
--- a/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
@@ -89,7 +89,8 @@ namespace jank::runtime::obj
       auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
       return 0 <= i && i < data.size();
     }
-    throw std::runtime_error{ fmt::format("contains? not supported on string: {}", runtime::to_string(key)) };
+    throw std::runtime_error{ fmt::format("contains? not supported on string: {}",
+                                          runtime::to_string(key)) };
   }
 
   object_ptr persistent_string::get_entry(object_ptr const) const
--- a/compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
+++ b/compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
@@ -1,0 +1,40 @@
+/* clang-format off */
+#include <jank/runtime/obj/persistent_string.hpp>
+#include <jank/runtime/obj/character.hpp>
+
+/* This must go last; doctest and glog both define CHECK and family. */
+#include <doctest/doctest.h>
+
+namespace jank
+{
+  TEST_SUITE("persistent_string")
+  {
+    TEST_CASE("Index")
+    {
+      SUBCASE("Present no default")
+      {
+        runtime::obj::persistent_string const s{ "foo bar" };
+        CHECK_EQ(s.get(s, runtime::make_box<runtime::obj::integer>(0)), runtime::make_box<runtime::obj::character>("f"));
+      }
+
+      SUBCASE("Absent no default")
+      {
+        runtime::obj::persistent_string const s{ "foo bar" };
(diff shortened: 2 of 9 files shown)


compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -62,6 +63,41 @@ namespace jank::runtime::obj
...
+
+  native_bool persistent_string::contains(object_ptr const key) const
+  {
+    if(key->type == object_type::integer)
+    {
+      auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
+      return 0 <= i && i < data.size();
+    }
+    throw std::runtime_error{ fmt::format("contains? not supported on string: {}",
+                                          runtime::to_string(key)) };
```
jeaye, line 93: This is a confusing error message. It says to me that the thing which comes after `:` is a string which doesn't support `contains?`.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -0,0 +1,70 @@
+/* clang-format off */
```
jeaye, line 1: Why is this here? None of the C++ files in the jank repo have formatting disabled.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -0,0 +1,70 @@
...
+        runtime::obj::persistent_string const s{ "foo bar" };
+        try
+        {
+          s.get_entry(runtime::make_box<runtime::obj::integer>(0));
+          CHECK(false);
+        }
+        catch(std::exception const &e)
+        {
+          std::string const actual = e.what();
+          std::string const expected = "get_entry not supported on string";
```
jeaye, line 64: Please *always* use direct initialization. Also, we can use `auto` here for both values.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -62,6 +63,41 @@ namespace jank::runtime::obj
...
+  object_ptr persistent_string::get(object_ptr const key) const
+  {
+    return get(key, nil::nil_const());
+  }
+
+  object_ptr persistent_string::get(object_ptr const key, object_ptr const fallback) const
+  {
+    if(key->type == object_type::integer)
+    {
+      auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
```
jeaye, line 75: Converting a signed value to an unsigned value will result in underflows. We'll want to handle the `< 0` case before casting, I think.

Makes sense to have a test case for negative values as well.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <jank/runtime/obj/persistent_string.hpp>
+#include <jank/runtime/obj/character.hpp>
+#include <jank/runtime/obj/number.hpp>
+#include <jank/runtime/obj/nil.hpp>
+#include <jank/runtime/core.hpp>
+
+/* This must go last; doctest and glog both define CHECK and family. */
+#include <doctest/doctest.h>
+
+namespace jank
```
jeaye, line 11: You're testing `jank::runtime::obj::persistent_string`, so we should be in `jank::runtime::obj`. That would allow you to get rid of all of the namespacing below, too.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -62,6 +63,41 @@ namespace jank::runtime::obj
...
+    }
+    return fallback;
+  }
+
+  native_bool persistent_string::contains(object_ptr const key) const
+  {
+    if(key->type == object_type::integer)
+    {
+      auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
+      return 0 <= i && i < data.size();
```
jeaye, line 90: `i is always > 0` since `i` is unsigned here. We'll need the `< 0` check to be done prior to the cast.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+  TEST_SUITE("persistent_string")
+  {
+    TEST_CASE("Index")
```
jeaye, line 15: We're missing test cases for the `throw` in each `get`, where the input is not an integer.

jeaye (changes requested):
Behavior-wise, looks great. Just some stylistic, messaging, and testing comments.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
frenchy64: I used the one C++ file where that is not true as a starting point and assumed that was normal without checking.

https://github.com/jank-lang/jank/blob/676cd4f98a7514edb87839065a9719c1bdf362ef/compiler%2Bruntime/test/cpp/jank/native_persistent_string.cpp#L1

jeaye: Fair point. Can't fault you for following the precedent. Neither of these should have that set, at this point, though.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
frenchy64: `get` does not throw.

frenchy64: But there's definitely a bug in the `get` fallback, I need another case.

jeaye: Oh, whoops. I meant `contains` throwing.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
frenchy64: Ah, I based it on persistent_vector.cpp, does that need the same fix? The vector `contains` impl looks more like what you're saying is correct but both `get` and `get_entry` seem suspicious.

https://github.com/jank-lang/jank/blob/86c16899c446f5ef5d54cf3c6317e7290dd72ca8/compiler%2Bruntime/src/cpp/jank/runtime/obj/persistent_vector.cpp#L232

frenchy64: Actually maybe it's just the `get` without fallback on `persistent_vector.cpp` that seems off.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
frenchy64: What do you think about returning `false` instead of throwing for `(contains? "a" non-int)`? Clojure throws, CLJS returns false.

jeaye: If CLJS sets a precedent for `false`, I'd follow that. Exceptions for predicates are bonkers.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
jeaye: Yeah, I'd say so. Better to have more checks to start with.

frenchy64: I found the same problem in several other files and attempted fixes.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
frenchy64: Changed `(contains? "a" non-int)` to return false and added your suggested test cases and a couple more.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
frenchy64: Returns `false` instead of throwing now.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_vector.cpp
```diff
@@ -0,0 +1,73 @@
...
+  TEST_SUITE("persistent_vector")
+  {
+    static persistent_vector_ptr const v(persistent_vector::empty()
+                                           ->conj(make_box('f'))
+                                           ->conj(make_box('o'))
+                                           ->conj(make_box('o'))
+                                           ->conj(make_box(' '))
+                                           ->conj(make_box('b'))
+                                           ->conj(make_box('a'))
+                                           ->conj(make_box('r')));
```
jeaye, line 21: 1. We can use `auto`
2. We can use `make_box` and the `in_place` ctor to provide all values at once

```cpp
  static auto const v{
    make_box<persistent_vector>(std::in_place,
      make_box('f'),
      make_box('o'),
      make_box('o'),
      make_box(' '),
      make_box('b'),
      make_box('a'),
      make_box('r') };
```

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -0,0 +1,69 @@
...
+    TEST_CASE("get_entry not implemented")
+    {
+      try
+      {
+        s.get_entry(min);
+        CHECK(false);
+      }
+      catch(std::exception const &e)
+      {
+        auto const actual{ fmt::format("{}", e.what()) };
```
jeaye, line 63: Shouldn't need to format anything here. Just `e.what()` should work.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_vector.cpp
```diff
@@ -0,0 +1,73 @@
...
+      CHECK(!contains(v, non_int));
+    }
+    TEST_CASE("find")
+    {
+      CHECK(equal(find(v, min),
+                  persistent_vector::empty()->conj(make_box(min))->conj(make_box(min_char))));
+      CHECK(equal(find(v, mid),
+                  persistent_vector::empty()->conj(make_box(mid))->conj(make_box(mid_char))));
+      CHECK(equal(find(v, max),
+                  persistent_vector::empty()->conj(make_box(max))->conj(make_box(max_char))));
```
jeaye, line 67: Using the `in_place` ctor maybe be cleaner here, too.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -1,6 +1,7 @@
 #include <fmt/format.h>
 
 #include <jank/native_persistent_string/fmt.hpp>
+#include <jank/runtime/behavior/associatively_readable.hpp>
```
jeaye, line 4: I don't think this is needed.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_vector.cpp
```diff
@@ -0,0 +1,73 @@
+#include <jank/runtime/obj/character.hpp>
+#include <jank/runtime/obj/number.hpp>
+#include <jank/runtime/obj/nil.hpp>
+#include <jank/runtime/core.hpp>
+#include <jank/runtime/context.hpp>
```
jeaye, line 5: I suspect we don't need to include anything other than `<jank/runtime/obj/persistent_vector.hpp>`. Let's start there, though, and add only as needed. Keeping jank's compile-time low is important for our sanity.

jeaye (changes requested):
Nice changes! Minor comments.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_vector.cpp
frenchy64: Couldn't figure out how to use less than

```cpp
#include <jank/runtime/obj/persistent_vector.hpp>
#include <jank/runtime/core/make_box.hpp>
```

jeaye: Yep, those look good for this file.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -0,0 +1,64 @@
...
+    static auto const min{ make_box(0) };
+    static auto const min_char{ make_box('f') };
+    static auto const mid{ make_box(3) };
+    static auto const mid_char{ make_box(' ') };
+    static auto const max{ make_box(6) };
+    static auto const max_char{ make_box('r') };
+    static auto const over{ make_box(7) };
+    static auto const under{ make_box(-1) };
+    static auto const nil{ nil::nil_const() };
+    static auto const non_int{ make_box('z') };
```
jeaye, line 21: We don't benefit from these being static. Same goes with the other test files.

jeaye (changes requested):
One last comment, which I noticed in another PR of yours and made me realize I had seen it here, too. Otherwise looks good!

Issue #216, opened 2025-01-30
`(nthnext c 0)` should return `(seq c)`


frenchy64:
This was fixed in Clojure around 1.10-1.12.

```clojure
;; jank
clojure.core=> (nthnext "ab" 0)
"ab"

;; clojure
user=> (nthnext "ab" 0)
(\a \b)
```

PR #229, opened 2025-02-03
(nthnext c 0) returns (seq c)

bpiel:
Closes https://github.com/jank-lang/jank/issues/216

```
clojure.core=> (nthnext "abc" 0)
(\a \b \c)
```

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -2001,7 +2001,7 @@
   [coll n]
   (if (and (seq coll) (pos? n))
     (recur (next coll) (dec n))
-    coll))
+    (seq coll)))
 
 (defn nthrest
   "Returns the nth rest of coll, coll when n is 0."


frenchy64:
I think if we use Clojure's version via an inner loop it'll save a call to `seq` when recurring:

https://github.com/clojure/clojure/blob/fb22fd778a272b034684a4ee94509552b46ee8a9/src/clj/clojure/core.clj#L3181-L3184

frenchy64:
Oh and the original jank implementation wasn't correct anyway because it should call `next` on the result of `seq`.

jeaye:
Hi Bill! Thanks for the PR.

I agree with Ambrose that we should follow Clojure here. jank's `nthrest` wasn't using `loop` since we hadn't yet implemented `loop` when I brought `nthrest` in. We can leave a TODO for `IDrop` support, too.

bpiel:
Ok, thanks for the feedback.

frenchy64 (approved):
LGTM

```
clojure.core=> (nthnext "a" 0)
(\a)
clojure.core=> (nthnext "" 0)
nil
clojure.core=> (nthnext "ab" 1)
(\b)
```

jeaye:
Great work, Bill!

PR #230, opened 2025-02-04
Don't escape `?` or `\?` in strings

### LINKED ISSUE #221: String escaping issues
https://clojurians.slack.com/archives/C03SRH97FDK/p1738274725111159

```clojure
clojure.core=> "asdf?"
"asdf\?"

clojure.core=> "\?"
"\?"
```

frenchy64:
Close #221 

Because Clojure strings don't recognize the escape sequence `\?`, we don't want to print strings that contain it. In C++ strings, `\?` and `?` are the same character, so for Clojure compatibility, this PR has jank print both `?` and `\?` as `?` instead of the current `\?`.

It's handy to be able to write `"\?"` in `.jank` files coming from C++, so the lexer still considers it a valid escape sequence and still unescapes it to `?`. A cosmetic change is included in `unparse` to emphasize that `\?` and `?` are the same character in C++.

```
clojure.core=> "?"
"?"
clojure.core=> "\?"
"?"
clojure.core=> "\\?"
"\\?"
clojure.core=> (count "\\?")
2
clojure.core=> (str \?)
"?"
clojure.core=> (str \\ \?)
"\\?"
clojure.core=> (count "\?")
1
clojure.core=> (count "?")
1
clojure.core=> (= "?" "\?")
true
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/util/escape.cpp
+++ b/compiler+runtime/src/cpp/jank/util/escape.cpp
@@ -5,18 +5,7 @@
 
 namespace jank::util
 {
-  /* Converts escape sequences (2 characters) to their mapped (single) character:
-   *  \n => '\n'
-   *  \t => '\t'
-   *  \r => '\r'
-   *  \\ => '\\'
-   *  \" => '"'
-   *  \a => '\a'
-   *  \v => '\v'
-   *  \? => '?'
-   *  \f => '\f'
-   *  \b => '\b'
-   * */
+  /* Converts escape sequences starting with backslash to their mapped character. e.g., \" => " */
   result<native_persistent_string, unescape_error> unescape(native_persistent_string const &input)
   {
     util::string_builder sb{ input.size() };
@@ -79,17 +68,7 @@ namespace jank::util
     return ok(sb.release());
   }
 
-  /* Converts the following (single) characters to their escape sequences (2 characters):
-   *  '\n' => \n
-   *  '\t' => \t
-   *  '\r' => \r
-   *  '\\' => \\
-   *  '"'  => \"
-   *  '\a' => \a
-   *  '\v' => \v
-   *  '\f' => \f
-   *  '\b' => \b
-   * */
+  /* Converts special characters to their escape sequences. e.g., " => \" */
   native_persistent_string escape(native_persistent_string const &input)
   {
     /* We can expect on relocation, since escaping anything will result in a larger string.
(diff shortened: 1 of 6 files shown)


compiler+runtime/src/cpp/jank/util/escape.cpp
```diff
@@ -69,7 +69,7 @@ namespace jank::util
 
   native_persistent_string escape(native_persistent_string const &input)
   {
-    /* We can expect on relocation, since escaping anything will result in a larger string.
+    /* We can expect on relocation, since escaping results in an equal or larger length string.
```
frenchy64, line 72: I'm still mulling over this comment update. Getting confused on which direction is {un}escape.

I think the original comment still holds.

frenchy64: I added documentation that helped me understand the difference between unescape and escape.

compiler+runtime/test/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include <doctest/doctest.h>
+
+namespace jank::runtime::obj
+{
+  TEST_SUITE("persistent_string")
+  {
+    TEST_CASE("to_code_string")
+    {
+      static auto const s{ make_box("?") };
+      static auto const expected{ make_box("\"?\"") };
```
jeaye, line 14: We don't benefit from these being static. Especially since the test will only be run once.

compiler+runtime/src/cpp/jank/util/escape.cpp
```diff
@@ -5,6 +5,18 @@
...
+   *  \n => '\n'
+   *  \t => '\t'
+   *  \r => '\r'
+   *  \\ => '\\'
+   *  \" => '"'
+   *  \a => '\a'
+   *  \v => '\v'
+   *  \? => '?'
+   *  \f => '\f'
+   *  \b => '\b'
```
jeaye, line 18: (applies to escape and unescape)

I understand the goal of documenting these fns to avoid confusion. Let's not exhaustively duplicate the entire mapping into the comments, though; that's very clearly seen in the `switch` statement. We can give an example of '\n' becoming \n and vice versa.

jeaye (changes requested):

jeaye:
Thanks, Ambrose!

Issue #232, opened 2025-02-04
Add support for shebangs


jeaye:
Clojure has reader support for `#!`. When it sees that pattern, it just ignores the rest of the line. We should update jank to treat `#!` the same way it treats `;`.

This involves updating the lexer to output a `comment` token when `#!` is found. We already have the code for handling comments and for detecting `#`. We just need to add a `!` case following `#` which then reuses the comment code (pull it into a fn named `build_comment`).

We then need lex tests for the shebang showing up in various places. We can extend the lex comment tests for this.

bpiel:
I'll take this.

PR #236, opened 2025-02-05
Add support for shebangs

bpiel:
Closes https://github.com/jank-lang/jank/issues/232

```
clojure.core=> #! hello
nil
clojure.core=> (+ 1 1) #! addition
2
```

From the issue:
> We just need to add a ! case following # which then reuses the comment code (pull it into a fn named build_comment).

The code I ended up writing for shebang, while heavily inspired by the existing comment code, ended up being different in a few places throughout the code, such that it didn't seem appropriate to me to try to reuse a common function. Let me know if I'm missing something and should change my approach.

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -1072,7 +1072,7 @@ namespace jank::read
                   }
                   else
                   {
-                    const size_t length = pos - token_start - 2;
+                    size_t const length = pos - token_start - 2;
                     native_persistent_string_view const comment{ file.data() + token_start + 2,
                                                                  length };
                     return ok(token{ token_start, length, token_kind::comment, comment });
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -1047,6 +1047,37 @@ namespace jank::read
                                      token_kind::reader_macro_conditional });
                   }
                 }
+              case '!':
+                {
+                  while(true)
+                  {
+                    auto const oc(peek());
+                    if(oc.is_err())
+                    {
+                      break;
+                    }
+                    auto const c(oc.expect_ok().character);
+                    if(c == '\n')
+                    {
+                      break;
+                    }
+
+                    ++pos;
+                  }
+
+                  ++pos;
+                  if(pos == token_start + 2)
+                  {
+                    return ok(token{ token_start, 1, token_kind::comment, ""sv });
+                  }
+                  else
+                  {
+                    size_t length = pos - token_start - 2;
+                    native_persistent_string_view const comment{ file.data() + token_start + 2,
+                                                                 length };
+                    return ok(token{ token_start, length, token_kind::comment, comment });
+                  }
+                }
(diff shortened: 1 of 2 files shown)


compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -1574,6 +1574,85 @@ namespace jank::read::lex
         }
       }
 
+      SUBCASE("Shebang Comments")
+      {
+        SUBCASE("Empty")
+        {
```
jeaye, line 1580: All of your tests have a space after `#!`, which is not how shebangs are used by default. We should have an actual test for a real shebang.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -1047,6 +1047,37 @@ namespace jank::read
...
+                  }
+
+                  ++pos;
+                  if(pos == token_start + 2)
+                  {
+                    return ok(token{ token_start, 1, token_kind::comment, ""sv });
+                  }
+                  else
+                  {
+                    size_t const length = pos - token_start - 2;
```
jeaye, line 1075: 1. We can use auto here
2. Please *always* use direct initialization

```cpp
auto const length{ pos - token_start - 2 };
```

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -1574,6 +1574,85 @@ namespace jank::read::lex
...
+          native_vector<result<token, error>> const tokens(p.begin(), p.end());
+          CHECK(tokens
+                == make_tokens({
+                  {  0, 1, token_kind::integer,       1ll },
+                  {  2, 5, token_kind::comment, " meow"sv },
+                  { 10, 1, token_kind::integer,       2ll }
+          }));
+        }
+      }
+
```
jeaye, line 1655: I think we can exercise the tests some more. For example:

* `##!foo`
* `#!!foo`
* `#! foo #!`
* any other ways this can fail

jeaye (changes requested):

compiler+runtime/test/cpp/jank/read/lex.cpp
bpiel: Gotcha.
Honestly, I did not know `#!` was a thing until this issue, so I was just copying the `;` comment tests.

bpiel: Well, I know about using them at the top of scripts... I didn't know they would be treated like a comment by clojure.

jeaye: Clojure supports them so that you can do things like:

```bash
#!/usr/bin/env bb
(println "hello")
```
And then
```bash
chmod +x hello.clj
./hello.clj
```

I suppose the easiest way for them to do that was to just treat `#!` as a comment.

bpiel: Yeah, I wasn't getting that when I was coding this. Thanks

jeaye (approved):

jeaye:
Great work, Bill!

Issue #238, opened 2025-02-06
Idempotent `alias` call fails


frenchy64:
I think the check in `ns::add_alias` doesn't use equality.

```
clojure.core=> (alias 'c 'clojure.core)
nil
clojure.core=> (alias 'c 'clojure.core)
Assertion failed: (o->type == T::obj_type), function expect_object, file rtti.hpp, line 91.
zsh: abort      ./build/jank repl
```

Issue #237, opened 2025-02-06
Syntax quote does not recognize namespace aliases


frenchy64:
```clojure
;;jank
clojure.core=> (require '[clojure.template :as temp])
nil
clojure.core=> `temp/do-template
temp/do-template

Clojure 1.12.0
user=> (require '[clojure.template :as temp])
nil
user=> `temp/do-template
clojure.template/do-template
```

Issue #195, opened 2025-01-29
Premature resolution of symbol namespace `foo/bar`


frenchy64:
```clojure
clojure.core=> #_foo/bar 1
Read error (2 - 2): unknown namespace: foo
clojure.core=> #?(:clj Long/MAX_VALUE :default 1)
Read error (8 - 8): unknown namespace: Long
```

PR #239, opened 2025-02-06
Resolve ns aliases in syntax quote and don't resolve syms unless needed

frenchy64:
Closes https://github.com/jank-lang/jank/issues/237
Closes https://github.com/jank-lang/jank/issues/238
Closes https://github.com/jank-lang/jank/issues/195

Most impactful change is `#?(:clj Long/MAX_VALUE :default 1)` now parses successfully (before, jank would complain that `Long` is not a namespace).

I included some script changes that I found useful.


compiler+runtime/bin/watch
```diff
@@ -2,7 +2,17 @@
...
+here="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+JANK_WATCH_SLEEP="${JANK_WATCH_SLEEP:-1}"
+
+if [ $# -eq 0 ]
+then
+  COMMAND="${here}/test"
+else
+  COMMAND="$*"
+fi
+
```
jeaye, line 14: Not sure of your use case here, but overall I don't want `watch` to default to running `test`. Using `^r` in your terminal should be able to bring you very quickly to a previous `./bin/watch ./bin/test`, if not a local alias or script you set up.

If you let me know the use case for removing the sleep, we can consider it. However, it will not use an environment variable; it'll use a CLI flag.

These changes also make me realize that we lost shellcheck linting when we moved to bb. These changes would not pass the linting, do to the lack of `${}` int he variable derefs. Side note that bash scripts should not have full-caps variables. Those are only used for environment variables, by convention, which is led a lot of people who don't write much bash to use them by default in bash. As you see with `here`, we want them to be lowercase.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1196,17 +1200,18 @@ namespace jank::read::parse
...
-        /* Normal symbols will have the ns resolved immediately. */
+        /* Normal symbols will have the ns resolved immediately if resolvable. */
         else
         {
           auto const resolved_ns(__rt_ctx->resolve_ns(make_box<obj::symbol>(ns_portion)));
           if(resolved_ns.is_none())
           {
-            return error::parse_unresolved_namespace(
-              fmt::format("Unresolved namespace '{}'", ns_portion),
-              { start_token.start, latest_token.end });
```
jeaye, line 1207: This error is no longer used and should be removed entirely (from the enum, its uses, and this fn).

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -480,7 +485,7 @@ namespace jank::runtime
     native_persistent_string resolved_ns{ ns };
     if(!resolved)
     {
-      /* The ns will be an ns alias. */
+      /* The ns will be an ns or alias. */
```
jeaye, line 488: If it's not resolved, it'll be an alias. There is no `or` here. `::foo/bar` is an unresolved keyword and `foo` *must* be an alias.

From a clj repl:

```
user=> ::clojure.core/foo
Syntax error reading source at (REPL:3:0).
Invalid token: ::clojure.core/foo

user=> :clojure.core/foo
:clojure.core/foo
```

compiler+runtime/src/cpp/jank/runtime/ns.cpp
```diff
@@ -43,6 +43,18 @@ namespace jank::runtime
     return new_var;
   }
 
+  result<void, native_persistent_string> ns::unmap(obj::symbol_ptr const &sym)
+  {
+    if(!sym->ns.empty())
+    {
+      return err("Can't unintern namespace-qualified symbol");
```
jeaye, line 50: Let's format the `sym` into this error.

compiler+runtime/src/cpp/jank/runtime/ns.cpp
```diff
@@ -60,19 +72,29 @@ namespace jank::runtime
...
+  result<void, native_persistent_string>
+  ns::add_alias(obj::symbol_ptr const &sym, ns_ptr const &nsp)
   {
     auto locked_aliases(aliases.wlock());
     auto const found((*locked_aliases)->data.find(sym));
-    if(found && expect_object<var>(*found) != ns)
+    if(found)
     {
-      return err(fmt::format("Alias already bound to a different ns: {}", sym->to_string()));
+      if(expect_object<ns>(*found) != nsp)
```
jeaye, line 82: I'm confused about this. Why were we expecting a `var` first, now we're expecting an `ns`?

jeaye (changes requested):
My mistake for not looking at this sooner. Feel free to ping me if I've not looked at something after a day. I get a _lot_ of Github emails and sometimes things get buried.

compiler+runtime/src/cpp/jank/runtime/ns.cpp
frenchy64: `nsp` is an `ns_ptr` here so we need to cast to match it.

I improved the error message here, basing it on a similar message from `ns::refer`.

```clojure
Bottom of clojure.core
clojure.core=> (create-ns 'foo)
foo

clojure.core=> (alias 'a 'clojure.core)
nil

clojure.core=> (alias 'a 'clojure.core)
nil

clojure.core=> (alias 'a 'foo)
Exception: a already aliases clojure.core in ns clojure.core, cannot change to foo
```

jeaye: I don't think that addresses my confusion. We went from this: `expect_object<var>(*found)` to this: `expect_object<ns>(*found)`

Has the type of `*found` changed? Was it wrong to cast it to `var` in the first place?

compiler+runtime/bin/watch
frenchy64: > Not sure of your use case here

Not important, I'll let you know if it comes up again.

frenchy64: Issue for shellcheck: https://github.com/jank-lang/jank/issues/279

compiler+runtime/src/cpp/jank/runtime/context.cpp
jeaye: Hm, the error is inside out here. The note should be the message, while the note should be "Found here". I'll fix that on my end, since it's not related to your PR.

compiler+runtime/src/cpp/jank/runtime/ns.cpp
frenchy64: My best guess is you're looking at code that uses `this->vars`. This function uses `this->alias`.

frenchy64: Oh, the diff. I can't explain that, I think it was wrong before.

jeaye: Switched to side-by-side view for this screenshot. Left side uses `expect_object<var>`. Right side uses `expect_object<ns>`.

image

frenchy64: Yes, it crashes on main.

```clojure
Bottom of clojure.core
clojure.core=> (alias 'a 'clojure.core)
nil

clojure.core=> (alias 'a 'clojure.core)
Assertion failed: (o->type == T::obj_type), function expect_object, file rtti.hpp, line 91.
zsh: abort      ./build/jank repl
```

frenchy64: > Switched to side-by-side view for this screenshot. Left side uses expect_object<var>. Right side uses expect_object<ns>.

Yes sorry. Left side contains a bug that I didn't notice.

jeaye: :joy: Alright. Agreed that it looks blatantly wrong on main.

jeaye (approved):
Great work, Ambrose!

frenchy64:
Thanks for your time Jeaye!

Issue #243, opened 2025-02-06
Sequence equality is broken


frenchy64:
```
clojure.core=> (= (range 5) (range 5))
false
clojure.core=> (= () (lazy-seq))
false


;; from clojure-test

FAIL in (nil) (nil:nil)
expected: (= (range 1) (butlast (range 2)))
  actual: (not (= (0) (0)))

FAIL in (nil) (nil:nil)
expected: (= (range 2) (butlast (range 3)))
  actual: (not (= (0 1) (0 1)))

FAIL in (nil) (nil:nil)
expected: (= (range 3) (butlast (range 4)))
  actual: (not (= (0 1 2) (0 1 2)))
```

jeaye:
I *suspect* this is related to #87, which is due to chunked seqs being borked.

PR #249, opened 2025-02-07
Sequence equality

frenchy64:
Closes https://github.com/jank-lang/jank/issues/243
Closes https://github.com/jank-lang/jank/issues/250
Closes https://github.com/jank-lang/jank/issues/251
Closes https://github.com/jank-lang/jank/issues/274
Closes https://github.com/jank-lang/jank/issues/283

It would be nice to replace `nullptr` returns with `nil` in `sequenceable` but it enables important optimizations for calls to `sequenceable` member functions. I did not investigate returning `option` types.


compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -245,7 +245,7 @@ namespace jank::runtime
     }
     else if(!rhs)
     {
-      return !lhs;
+      return false;
```
jeaye, line 248: Why change this? Right now, it would treat two null object ptrs as equal. This would also carry over to us removing `nullptr` support for boxes, such that two nils would be equal.

compiler+runtime/src/cpp/jank/runtime/obj/chunked_cons.cpp
```diff
@@ -167,7 +167,7 @@ namespace jank::runtime::obj
             return false;
           }
         }
-        return true;
+        return seq == nullptr;
```
jeaye, line 170: I don't think this is correct. In the case where `this` is a shorter seq than `o`, we would finish our `for` loop and then need to check if there is some `seq` left over.

compiler+runtime/src/cpp/jank/runtime/obj/cons.cpp
```diff
@@ -77,12 +77,12 @@ namespace jank::runtime::obj
         for(auto it(fresh_seq()); it != nullptr;
             it = runtime::next_in_place(it), seq = runtime::next_in_place(seq))
         {
-          if(seq == nullptr || !runtime::equal(it, seq->first()))
+          if(seq == nullptr || !runtime::equal(it->first(), seq->first()))
```
jeaye, line 80: This seems like a good fix. We'd need the same thing for `chunked_cons` above, though. I wonder where else this cropped up. I have a TODO for sequence equality having a lot of copy pasta.

compiler+runtime/src/cpp/jank/runtime/obj/detail/iterator_sequence.cpp
```diff
@@ -39,7 +39,7 @@ namespace jank::runtime::obj::detail
             return false;
           }
         }
-        return true;
+        return seq == nullptr;
```
jeaye, line 42: Very odd that you made the opposite change here that you did in other places. I think all of them should be `return seq == nullptr;`

compiler+runtime/test/cpp/jank/runtime/core/seq.cpp
```diff
@@ -0,0 +1,18 @@
...
+namespace jank::runtime::core
+{
+  TEST_SUITE("core runtime for seq")
+  {
+    TEST_CASE("sequence_equal")
+    {
+      CHECK(equal(nullptr, nullptr));
+      CHECK(!equal(nullptr, make_box(42)));
+      CHECK(!equal(make_box(42), nullptr));
+    }
```
jeaye, line 16: This is the same test as in the previous file. Doesn't do anything with sequences.

jeaye (changes requested):
Lots of good fixes in here. I think there are some introduced bugs though. I tried to pick them out.

compiler+runtime/src/cpp/jank/runtime/core.cpp
frenchy64: Thanks, I missed that box overloaded `!`.

compiler+runtime/src/cpp/jank/runtime/obj/cons.cpp
frenchy64: > We'd need the same thing for chunked_cons above, though.

Done.

> I wonder where else this cropped up.

Several places. I think I reverted some of them accidentally when experimenting with replace `nullptr` with `nil_const()`, I'll have another look.

> I have a TODO for sequence equality having a lot of copy pasta.

I wasn't sure how to tackle it, but I can at least write more tests so then the abstraction can be validated easily.

compiler+runtime/src/cpp/jank/runtime/obj/chunked_cons.cpp
frenchy64: > In the case where this is a shorter seq than o, we would finish our for loop and then need to check if there is some seq left over.

This is my attempt at adding that check. At this point, we've already called `next_in_place(seq)`, and `seq` will be `nullptr` if there's none left over.

compiler+runtime/src/cpp/jank/runtime/obj/cons.cpp
jeaye: Don't worry about removing support for `nullptr` yet. Certainly not in this PR, which already has a lot going on.

jeaye: I have some ideas for how to do it. The tests are very welcome, though.

compiler+runtime/src/cpp/jank/runtime/core.cpp
frenchy64: Actually, could you look at this again? At this point we already know `!lhs` is `false`, so I think returning either is equivalent.

Would this be another way of writing the test?

```c++
if(!lhs != !rhs) return false
```

compiler+runtime/test/cpp/jank/runtime/core/seq.cpp
frenchy64: Added tests for `sequence_equal`.

frenchy64:
Found and fixed a couple more issues while I scanned everything. Lmk if I should to pull them out.

- `apply_to` isn't compatible with seqs that return fresh seqs from `next_in_place`. Not sure if they exist, but `persistent_list` is implemented that way.
- removed transient cases to resolve https://github.com/jank-lang/jank/issues/274
  - extra: `conj` case for seqables didn't seem right to me compared to Clojure, so I removed it. For one, it changes the type of the seqable to a seq.
- added paranoid assertions for `nil` and `null` while the function/template ambiguity exists for `runtime::{next,seq,next_in_place,fresh_seq}`.

compiler+runtime/src/cpp/jank/runtime/core.cpp
frenchy64: Added assertions to point out my reasoning.

compiler+runtime/src/cpp/jank/runtime/obj/chunked_cons.cpp
frenchy64: We're talking about the same diff, right?

```
-       return true;
+       return seq == nullptr;
```

jeaye: I think I saw the opposite diff. :shrug: My mistake!

compiler+runtime/src/cpp/jank/runtime/obj/chunked_cons.cpp
```diff
@@ -162,12 +162,15 @@ namespace jank::runtime::obj
         for(auto it(fresh_seq()); it != nullptr;
             it = runtime::next_in_place(it), seq = runtime::next_in_place(seq))
         {
-          if(seq == nullptr || !runtime::equal(it, seq->first()))
+          assert(it != nil::nil_const());
+          assert(seq != nil::nil_const());
```
jeaye, line 166: It's impossible for these assertions to be false, proven by the type system. I think you're concerned about the `next_in_place(it)` call somehow returning an `object_ptr`. But it could never do that, since it would result in a compiler error.

Now, the history is that `next_in_place` has been in flux for a year or so and I was making some changes which would allow for it to be optional. That never happened, but it's why we're not using `it->next_in_place()` here. Ultimately, it's doing the same thing and all of these can be changed back. I just haven't done that yet, since it's a perf change and I haven't been focusing on perf.

Either way, your concerns about this being nil should not be keeping you up. `auto it(fresh_seq());` means that we'll have a fully typed `it` for the currently visited object. We cannot assign an `object_ptr` to it. C++ won't allow it.

compiler+runtime/src/cpp/jank/runtime/obj/repeat.cpp
```diff
@@ -16,6 +16,7 @@ namespace jank::runtime::obj
     : value{ value }
     , count{ count }
   {
+    assert(0 < to_int(count));
```
jeaye, line 19: Feels like this could be an exception, rather than assertion, since someone embedding jank may create it a repeat with `0` and we don't need to kill the whole program.

jeaye (changes requested):
Mostly a rehash of our slack discussion, but also called out some other concerns.

compiler+runtime/src/cpp/jank/runtime/behavior/callable.cpp
```diff
@@ -878,88 +878,88 @@ namespace jank::runtime
           case 1:
             return dynamic_call(source, s->first());
           case 2:
-            return dynamic_call(source, s->first(), s->next_in_place()->first());
+            return dynamic_call(source, s->first(), (s = s->next_in_place())->first());
```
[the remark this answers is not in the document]
frenchy64: > Either way, let's please undo this file's changes.

Sure.

> I don't want to impact its perf for this one case which needs to change anyway.

I don't think the bug is reachable right now.

`persistent_list::next_in_place` is the potential problem, but it's currently unreachable to `apply_to` since a fresh seq is taken and then `persistent_list_sequence::next_in_place` is called instead.

I'm not sure what would happen if `persistent_list_sequence` were removed and (presumably) replaced with `persistent_list` itself. I think the `apply_to` bug would be introduced if `persistent_list::next_in_place` continued to allocate a new list, as is implied in this comment.

```c++
  persistent_list_sequence_ptr persistent_list::next_in_place() const
  {
    /* In-place updates don't make sense for lists, since any call to fresh_seq would return
     * a list sequence. So we know, principally, that a list itself cannot be considered fresh. */
    return next();
  }
```

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -243,9 +243,10 @@ namespace jank::runtime
     {
       return !rhs;
     }
-    else if(!rhs)
+    if(!rhs)
```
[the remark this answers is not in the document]
frenchy64: I added back the `else`. Since the compiler pointed out the assertion was redundant, does that mean returning `false` is equivalent to returning `!lhs`?

compiler+runtime/src/cpp/jank/runtime/obj/repeat.cpp
frenchy64: Done.

It doesn't necessarily need to be an error either. We can made these constructors robust by adding checks in its members.

compiler+runtime/src/cpp/jank/runtime/obj/chunked_cons.cpp
frenchy64: Thanks for the explanation. What about something like this loop? Can this be `nil`?

https://github.com/jank-lang/jank/blob/bf75688aa2dca21f2db111393b20c5d047c68e8c/compiler%2Bruntime/src/cpp/clojure/core_native.cpp#L575-L581

frenchy64: Eep.

```clojure
clojure.core=> (== 1 1 1 1 1)
Exception: not a number: nil
```

compiler+runtime/src/cpp/jank/runtime/behavior/callable.cpp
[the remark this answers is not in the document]
jeaye: The comment would no longer apply, if we removed `persistent_list_sequence`.

compiler+runtime/src/cpp/jank/runtime/core.cpp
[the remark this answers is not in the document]
jeaye: Yep, equivalent. Was just like that for symmetry.

compiler+runtime/src/cpp/jank/runtime/obj/chunked_cons.cpp
jeaye: That one can be `nil`, yes, since it's using the `object_ptr` overloads. However, it cannot be `nullptr`, since those functions never return `nullptr`.

compiler+runtime/src/cpp/clojure/core_native.cpp
```diff
@@ -578,7 +578,8 @@ jank_object_ptr jank_load_clojure_core_native()
         return obj::boolean::false_const();
       }
 
-      for(auto it(fresh_seq(rest)); it != nullptr; it = next_in_place(it))
+      for(auto it(fresh_seq(rest)); it != nullptr && it != obj::nil::nil_const();
```
jeaye, line 581: So, here, `it` cannot be `nullptr`. That's not part of the contract. Everything returned from `runtime::` fns should be non-null.

The one exception is the `next_in_place` and `fresh_seq` templates. Ultimately, we can remove those (not in this PR); they're only there for a historical partial refactor which never panned out.

There should never be a case, right now, where we need to check for nullptr *and* nil.

jeaye (approved):
All good with your previous changes. The new bug fix has a comment, though. I'll fix it locally.

Thanks for the iteration, Ambrose!

compiler+runtime/src/cpp/clojure/core_native.cpp
frenchy64: > There should never be a case, right now, where we need to check for nullptr and nil.

There's a few more in `multi_function.cpp`.

PR #252, opened 2025-02-07
Return failing exit code if `jank` or `jank-test` catches exception in main

### LINKED ISSUE #181: `run-main` returns 0 exit code on exception
Probably should be non-zero.

```clojure
clojure-cli-test/lein-jank/test-project $ lein jank run
Bottom of clojure.core
Exception: invalid call to test-project.core/-main (jit_function@0x7ea926d0a2d8) with 0 args provided
clojure-cli-test/lein-jank/test-project $ echo $?
0
```

frenchy64:
Closes #181 

Also fixes `lein-jank` error handling.


frenchy64:
Something is very wrong in CI, reverting to draft.

frenchy64:
I think the culprit was a lack of `apt-get update`.

compiler+runtime/bin/jank/compiler+runtime/bash_test.clj
```diff
@@ -25,19 +26,25 @@
           (if skip?
             (util/log-warning "Skipped " relative-dirname)
             (util/with-elapsed-time duration
-              (let [res @(b.p/process {:out :string
-                                       :err :out
-                                       :dir dirname
-                                       :extra-env extra-env}
-                                      test-file)]
-                (when (or (and (zero? (:exit res)) expect-failure?)
```
jeaye, line 33: I don't understand why we need timeouts for this. None of your exit code stuff is blocking, looping, etc. What's the use case for being concerned about it not returning fast enough?

compiler+runtime/bin/jank/compiler+runtime/bash_test.clj
```diff
@@ -25,19 +26,25 @@
...
+                                    :extra-env extra-env}
+                                   test-file)
+                    res (deref p (* timeout-seconds-per-script 1000) :timeout)]
+                (when (or (= :timeout res)
+                          (and (zero? (:exit res)) expect-failure?)
                           (and (not (zero? (:exit res))) (not expect-failure?)))
-                  (vreset! unexpected-result res)))
+                  (case res
+                    :timeout (do (b.p/destroy-tree p)
+                                 (vreset! unexpected-result (assoc @p :timeout true)))
```
jeaye, line 40: Please always put a new line after `do` (below, on line 43, too). Also, please use `:timeout?` since we're talking about a bool.

compiler+runtime/test/bash/main-exit/pass-test
```diff
@@ -0,0 +1,75 @@
...
+(deftest jank-run-successful-script
+  (let [{:keys [err out exit]}
+        (proc/shell {:out :string :err :string :continue true}
+                    "jank run src/jank_test/successful_script.jank")]
+    (is (empty? err))
+    (is (= "Bottom of clojure.core\njank_test/successful_script.jank some expected output\n{:foo \"42\"}\n"
+           out))
+    (is (= 0 exit))))
+
+(defn -main [] (System/exit (if (t/successful? (t/run-tests this-ns)) 0 1)))
```
jeaye, line 72: Please always start fn/defn bodies on a new line.

compiler+runtime/test/bash/main-exit/pass-test
```diff
@@ -0,0 +1,75 @@
...
+(ns jank-test-main-exit
+  (:require [babashka.process :as proc]
+            [clojure.string :as str]
+            [clojure.test :as t :refer [deftest run-test is]]))
+
+(def this-ns (ns-name *ns*))
+
+(deftest jank-run-main-successful-exit
+  (let [{:keys [err out exit]}
+        (proc/shell {:out :string :err :string :continue true}
```
jeaye, line 12: I would break all of these map entries onto their own lines, too. There's nothing wrong with more lines!

compiler+runtime/test/bash/main-exit/pass-test
```diff
@@ -0,0 +1,75 @@
...
+                    "jank --module-path src run-main jank-test.successful-exit-code")]
+    (is (empty? err))
+    (is (= "Bottom of clojure.core\njank-test.successful-exit-code: some expected output\n"
```
jeaye, line 15: These tests are incredibly fragile, since they're based on the raw stdout of jank. So, not only will they all break when we remove the `(println "Bottom of clojure.core")`, which should not break any tests, but they'll also not work on the error-reporting branch since we're outputting source snippets, stack traces, etc. If this is a test file for exit codes, let's focus on exit codes.

bin/jank/check_everything.clj
```diff
@@ -36,6 +36,7 @@
       (util/log-warning "Skipping dependency install, since we don't have apt-get")
       (do
         ; Install deps required for running our tests.
+        (util/quiet-shell {} "sudo apt-get update -y")
```
jeaye, line 39: I had avoided this for perf reasons, in CI, since apt is so damn slow to update. It wasn't needed, in my testing. What were you seeing which led you to solve by adding this?

jeaye (changes requested):

bin/jank/check_everything.clj
frenchy64: https://github.com/jank-lang/jank/actions/runs/13209379786/job/36879822368#step:3:802

```
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

One idea could be to only run update if apt-get fails without it.

jeaye: Hm, that sounds too clever. If it's failing without an update, even sometimes, we should just always update, as you did.

compiler+runtime/bin/jank/compiler+runtime/bash_test.clj
jeaye: jank's CI *does not* fail fast. We specifically have this:

```yml
jobs:
  test:
    strategy:
      fail-fast: false
```

compiler+runtime/test/bash/main-exit/pass-test
frenchy64: Thanks, replaced with:

```clojure
(deftest jank-exit-code-test
  (is-shell-exit 0 "jank --module-path src run-main jank-test.successful-exit-code")
  (is-shell-exit 1 "jank --module-path src run-main jank-test.fail-throw-exit")
  (is-shell-exit 1 "jank --module-path src run-main jank-test.fail-uncatchable-exit")
  (is-shell-exit 106 "jank")
  (is-shell-exit 1 "jank run src/jank_test/throwing_script.jank")
  (is-shell-exit 1 "jank run src/jank_test/uncatchable_script.jank")
  (is-shell-exit 0 "jank run src/jank_test/successful_script.jank"))
```

bin/jank/check_everything.clj
frenchy64: Another idea: it seems it's usually the jvm that's failing to install (via lein?). Temurin LTS distributions are cached on Actions runners, we could install lein manually if java exists?

compiler+runtime/bin/jank/compiler+runtime/bash_test.clj
frenchy64: Yes, sorry. I was looking at the wrong build where I changing the `timeout-minutes` to be lower.

Here's a better example: https://github.com/jank-lang/jank/actions/runs/13209300118/job/36879582648#step:3:46

The raw logs show we were for 25m while testing `main-exit`. I still have no idea why it happened. This was the motivation for the timeout changes. Mysteriously the timeouts stopped around when I added the `apt-get update`.

```
2025-02-07T22:52:39.9143327Z 🛈  Tested module/cljc (542ms)
2025-02-07T23:16:25.4280703Z ##[error]The operation was canceled.
```

bin/jank/check_everything.clj
jeaye: Hm. Let's keep the apt update for now. I plan on caching apt deps to speed up that process anyway. We can address it then.

compiler+runtime/bin/jank/compiler+runtime/bash_test.clj
jeaye: I suspect it was waiting for user input. Maybe because it was trying to install something?

frenchy64: Yes. The timeout changes were to prevent silly mistakes like this from melting CI machines, which might be better handled higher up in the script. Leaving as-is for now.

jeaye: Our CI workflow file has this:

```yml
    timeout-minutes: 30
```

Analysis builds take ~25m, while everything other build finishes in ~7m, and linting takes ~3m (most of which is spent installing apt deps). We could perhaps configure this per-job, like we do the name, so that it's 30 for analysis builds and 10 for normal builds and 5 for linting.

compiler+runtime/test/bash/main-exit/pass-test
```diff
@@ -0,0 +1,35 @@
...
+                                      cmd)]
+    (is (= expected-code exit)
+        (pr-str (select-keys res [:exit :out :cmd])))
+    res))
+
+(deftest jank-exit-code-test
+  (is-shell-exit 0 "jank --module-path src run-main jank-test.successful-exit-code")
+  (is-shell-exit 1 "jank --module-path src run-main jank-test.fail-throw-exit")
+  (is-shell-exit 1 "jank --module-path src run-main jank-test.fail-uncatchable-exit")
+  (is-shell-exit 106 "jank")
```
jeaye, line 23: Why 106?

frenchy64: I'm guessing it's from here https://github.com/jank-lang/CLI11/blob/5a03ee58384ce18b6b3de5a2256e18f5ef103633/include/CLI/Error.hpp#L52

Issue #244, opened 2025-02-06
Cannot catch `std::runtime_error`


frenchy64:
```
clojure.core=> (try (count 1) (catch e (println "here")))
Exception: not seqable: 1
clojure.core=> (try (drop-last [] nil) (catch e (println "HERE")))
Exception: not a number: []
```

jeaye:
This is by design, which is not to say it's great. There is no base `Throwable` type in C++. *Anything* can be thrown. We can currently only catch jank objects though, and only if they're type-erased (i.e. `object_ptr`).

It would be possible to add a wrapper jank object for `std::exception`, which is the base exception type for the C++ stdlib, but we're never just going to be able to catch any old value thrown by an arbitrary C++ lib.

frenchy64:
> It would be possible to add a wrapper jank object for std::exception, which is the base exception type for the C++ stdlib

I guess you'd then throw the wrapper object from jank's cpp files instead of std::exception? I think that would help implement `(is (thrown?))` IIUC.

Or `thrown?` could be implemented in C++ to access the default exception handler.

> but we're never just going to be able to catch any old value thrown by an arbitrary C++ lib.

Are you referring to catching the exception or referring to the thrown value? Is something like this mapping possible?

```c++
try {}
catch(jank::runtime::object_ptr const o) {}
catch(...) {}
```

```jank
(try
  (catch e)
  (catch :default))
```

jeaye:
> > It would be possible to add a wrapper jank object for std::exception, which is the base exception type for the C++ stdlib
> 
> I guess you'd then throw the wrapper object from jank's cpp files instead of std::exception? I think that would help implement `(is (thrown?))` for at least the `clojure-test-suite` cases IIUC.

jank would catch the `std::exception`, then do `throw make_box<obj::native_exception_wrapper>(e);`. This could then be caught by a normal jank `catch`.

> 
> Or `thrown?` could be implemented in C++ to access the default exception handler.
> 
> > but we're never just going to be able to catch any old value thrown by an arbitrary C++ lib.
> 
> Are you referring to catching the exception or referring to the thrown value? Is something like this mapping possible?
> 
> try {}
> catch(jank::runtime::object_ptr const o) {}
> catch(...) {}
> 
> ```
> (try
>   (catch e)
>   (catch :default))
> ```

Something like this is possible, provided we don't try to get any values from the exception caught via `...`.

jeaye:
Closing in favor of #558.

Issue #240, opened 2025-02-06
`try` requires a `catch` clause


frenchy64:
```clojure
;; jank
clojure.core=> (try (finally *ns*))
Read error (0 - 0): each try must have a catch clause

Clojure 1.12.0
user=> (try (finally 1))
nil
user=> (try)
nil
```

jeaye:
This is intentional. In C++, a try does require a catch.

PR #253, opened 2025-02-08
Fix `subs`

### LINKED ISSUE #248: `subs` 3rd arg should be exclusive-end, not number of chars from start
```
;; jank
clojure.core=> (subs "abcde" 1 4)
"bcde"
clojure.core=> (subs "abcde" 1 1)
"b"
clojure.core=> (subs "asdf" 3 1)
"f"

;; clojure
user=> user=> (subs "abcde" 1 4)
"bcd"
user=> (subs "abcde" 1 1)
""
user=> (subs "asdf" 3 1)
Execution error (StringIndexOutOfBoundsException) at jdk.internal.util.Preconditions$1/apply (Preconditions.java:55).
Range [3, 1) out of bounds for length 4

;; from clojure-test

FAIL in (nil) (nil:nil)
expected: (= "bcd" (subs "abcde" 1 4))
  actual: (not (= "bcd" "bcde"))
```

frenchy64:
Closes #248 
Related #244


compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -600,4 +600,37 @@ namespace jank::runtime
     return o->type == object_type::tagged_literal;
   }
 
+  object_ptr catch_all(object_ptr const f,
+                       object_ptr const normal,
+                       object_ptr const caught,
+                       object_ptr const unknown)
+  {
+    try
+    {
```
jeaye, line 609: I don't think this is needed. You have effectively reproduced the `jank_try` function. I'm on with a short-term workaround, but:

1. We should extend the `jank_try` function instead (which is used by LLVM IR gen for `try`)
2. A workaround like this *needs* a comment explaining why and a TODO to fix it
3. There is no need for `f` and `normal` to be separate fns; they're both part of the `try` body. The macro should've just combined them, since you don't use `res#` anyway.

So, `jank_try` doesn't have support for a `catch_all` param yet. We can't overload `jank_try`, since that's a C API. We'd need to add a new symbol. I think we can do this:

1. Extend `jank_try` to also catch `std::exception` and convert it to a string, as you're doing here
2. Add a new `jank_try_catch_all` which is the same as `jank_try` except for an additional `catch_all` param after `catch`. Then add a `catch(...)` clause to call that guy.

The TODO we need to place is in the macro, since when we add IR support for `(catch :default`, we can use `jank_try_catch_all`. The macro will need to use it directly for now and the TODO will be to just use `try` once we support catch-all expressions.

compiler+runtime/test/bash/module/jar/skip-test
```diff
@@ -1,8 +0,0 @@
-#!/usr/bin/env bash
```
jeaye, line 1: Why was this deleted?

jeaye (changes requested):

compiler+runtime/test/bash/module/jar/skip-test
frenchy64: It's the same file as `pass-test`. I wonder if it's leftover from testing the bb test runner?

compiler+runtime/src/cpp/jank/runtime/core.cpp
frenchy64: Thanks, I'll take a shot at it.

compiler+runtime/test/bash/module/jar/skip-test
jeaye: I renamed `pass-test` to `skip-test` since it's failing. If `pass-test` is also there, a bad merge happened.

frenchy64: Oh. Happy to go either way but I haven't seen any problems on my machine or CI.

https://github.com/jank-lang/jank/actions/runs/13164148063/job/36740057437#step:3:42

```
──── Run bash test suite ────
⚠  Skipped module/jar
🛈  Tested module/jar (932ms)
🛈  Tested module/cljc (763ms)
```

compiler+runtime/src/cpp/jank/runtime/core.cpp
frenchy64: I started but I'm going to keep going and (try to) prototype the whole thing on a separate branch, IR and all. This seems like a good learning exercise and a gentle introduction to LLVM IR and jank's compiler.

I have a couple of ideas I want to try out related to https://github.com/jank-lang/jank/issues/240, in particular exploring the coupling of jank's `finally` and C++'s `try`.

Moving to https://github.com/frenchy64/jank/pull/4 and I'll resume here once that exercise has (hopefully!) bore fruit.

jeaye: Hm, we should talk more about this before you get too far into it. In jank, `try` is the most complicated AST node to IR gen. Because of that, I've actually skipped proper IR gen for it and instead rely on a hack to workaround it and rely on C++ to do the work.

jeaye: Not opposed to you tackling this. Just want to make sure you have all of the requisite knowledge, since `try` is a beast.

frenchy64: Yes for sure, let's talk about it tomorrow.

I think I'm making some progress. Here's POC of `(try)` and `(try (finally))` which I pulled out of my big branch.

https://github.com/frenchy64/jank/pull/5/files

frenchy64: One thing I'm curious about is whether a real LLVM IR implementation enables the ability to add types to `catch` from jank.

frenchy64: Pulled all improvements to `thrown?` from this branch and updated the `clojure-test-suite` comment to point to #244.

compiler+runtime/test/cpp/jank/runtime/core.cpp
```diff
@@ -8,6 +9,20 @@ namespace jank::runtime
...
+      CHECK(equal(subs(s, make_box(1)), make_box<obj::persistent_string>("oo bar")));
+      CHECK(equal(subs(s, make_box(2)), make_box<obj::persistent_string>("o bar")));
+      CHECK(equal(subs(s, make_box(0), make_box(7)), make_box<obj::persistent_string>("foo bar")));
+      CHECK(equal(subs(s, make_box(0), make_box(0)), make_box<obj::persistent_string>("")));
+      CHECK(equal(subs(s, make_box(1), make_box(1)), make_box<obj::persistent_string>("")));
+      CHECK(equal(subs(s, make_box(1), make_box(6)), make_box<obj::persistent_string>("oo ba")));
+      CHECK(equal(subs(s, make_box(1), make_box(7)), make_box<obj::persistent_string>("oo bar")));
+      CHECK(equal(subs(s, make_box(3), make_box(4)), make_box<obj::persistent_string>(" ")));
+    }
+
```
jeaye, line 25: This is duplicating tests coverage from clojure-test-suite. We don't want to do that.

I accepted the `equal` tests below because they use `nullptr`, which is not currently possible to represent from jank code itself, so this had to be done in C++.

I know we can't run the `subs` tests in the clojure-test-suite. We'll get there, but we don't want to duplicate test coverage along the way.

frenchy64: Removed the tests.

jeaye:
Thanks, Ambrose!

PR #259, opened 2025-02-11
letfn

### LINKED ISSUE #257: Implement `letfn*`
# Needed
* Add a new `letfn` expression (follow `let`)
  * Needs new header
  * Needs to be added to the variant in `expression.hpp`
  * Will fail to compile until there are overloads for eval and gen (the next step)
* Add eval and codegen support
  * For eval just wrap the expression and call it, like we do with `let` normally. This will require codegen by `letfn` by wrapping it in a `fn` and the evaluating that
* Ensure `letfn` macro is added and working
* Add tests under `test/jank/form/letfn`

# References
* https://www.nils-blum-oeste.net/letfn-drilldown-a-clojure-special-form-in-detail/

frenchy64:
Closes #257 

The main part I'm not sure about is the `deferred_init` struct. I think ideally it should contain a `local_reference_ptr` that we simply pass to `gen`. But I couldn't figure out the C++ needed to achieve a deep (enough) copy of one, so instead I passed the arguments needed to construct one.

There are some misc changes in here that I noticed or needed:
- I don't think the `operator<` case for symbols looks right, which currently compares hashes. I think it should look like `operator<` of strings.
- I flushed logging a little earlier in `jank-test` to print which test we're up to if we segfault before a test finishes.


frenchy64:
I'll put this back to draft as I remove the boost algorithms (and consider whether we need replacements at all).

frenchy64:
That's where I ended up.

I don't think rearranging `letfn` bindings has any affect. It doesn't change the closed over variables, just the order in which we set them.

My initial idea was to "compile `letfn` to `let`" in the spirit of individual nano-pass optimizations stacking up to a greater than its parts, but it compiles to the same code.

frenchy64:
I found a subtle difference in the native source generated with `letfn` deferred initialization: immediately initialized contexts seem to have attribute `nonnull` when being passed to `jank_closure_create`. Deferred contexts don't.

For example, in `(letfn [(a [] b) (b [] a)])`:

```
%1 = call ptr @GC_malloc(i64 8)
                                                  // <- storing b deferred to final line
%a = call ptr @jank_closure_create(..., ptr %1)   // <- missing nonnull
...
%4 = call ptr @GC_malloc(i64 8)
store ptr %a, ptr %4, align 8.                          // <- immediately stores a
%b = call ptr @jank_closure_create(..., ptr nonnull %4) // <- has nonnull
store ptr %b, ptr %1, align 8
```

It seems like the strongest lead so far in whether we should investigate rearranging letfn bindings to minimize deferred initialization to just mutually recursive bindings.

I removed the analysis pass to do this and wasn't planning on adding it back, but it was nice to find a tangible difference.

jeaye:
> I found a subtle difference in the native source generated with `letfn` deferred initialization: immediately initialized contexts seem to have attribute `nonnull` when being passed to `jank_closure_create`. Deferred contexts don't.
> 
> For example, in `(letfn [(a [] b) (b [] a)])`:
> 
> ```
> %1 = call ptr @GC_malloc(i64 8)
>                                                   // <- storing b deferred to final line
> %a = call ptr @jank_closure_create(..., ptr %1)   // <- missing nonnull
> ...
> %4 = call ptr @GC_malloc(i64 8)
> store ptr %a, ptr %4, align 8.                          // <- immediately stores a
> %b = call ptr @jank_closure_create(..., ptr nonnull %4) // <- has nonnull
> store ptr %b, ptr %1, align 8
> ```
> 
> It seems like the strongest lead so far in whether we should investigate rearranging letfn bindings to minimize deferred initialization to just mutually recursive bindings.
> 
> I removed the analysis pass to do this and wasn't planning on adding it back, but it was nice to find a tangible difference.

We can also just add `nonnull` as an attribute ourselves, if that becomes a problem. :) No need for reworking the binding order to get that. We have full control over the IR; it's *our* IR!

compiler+runtime/test/jank/form/letfn/binding/fail-non-symbol.jank
```diff
@@ -0,0 +1 @@
+(letfn* [1 (clojure.core/fn 1 [])])
```
jeaye, line 1: There's no need for `clojure.core/` in any of these tests. It is referred. That's where `assert`, `=`, etc come from.

Better yet, though, use `fn*`! You're not using destructuring here.

jeaye (changes requested):
:+1: Nice, dude. Some minor comments, but this looks great. I like the thought you put into the test cases, too.

compiler+runtime/test/cpp/jank/jit/processor.cpp
```diff
@@ -69,6 +69,7 @@ namespace jank::jit
         std::stringstream const captured_output;
 
         util::print("testing file {} => ", dir_entry.path().string());
+        std::cerr.flush();
```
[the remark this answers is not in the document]
frenchy64: Yeah, I thought it was a bit strange that this "worked". But the original problem was a segfault happened and this line never printed, so I couldn't figure out which test broke. When I added this line, it printed, clearly by coincidence.

compiler+runtime/test/jank/form/letfn/binding/fail-non-symbol.jank
jeaye: These are tests for the compiler feature, `leftfn*`. The tests for `clojure.core/letfn`, the library feature, will live in clojure-test-suite.

jeaye: So, we'll get all of that coverage. Just has to live in different places. I've gone back and forth on this, in the past, for a long time before finally making this distinction, but I think it helps to have a clear cut line.

frenchy64: Yeah it'll come out in the wash and I'll change it, but here's my take:

The semantics that `letfn*` macroexpands the rhs (or, that it's an expression position) are defined in the analyzer, hence the compiler. We lose test coverage of this property by using `fn*` everywhere.

Happy to concede the fully qualified `clojure.core/fn` is unnecessary, it was doing double duty with the much the more important test of `letfn*`'s evaluation order.

jeaye: `letfn*` is completely separate from `clojure.core/letfn`. Let's not conflate them. It's possible to never use `clojure.core/letfn` at all, while still taking advantage of the `letfn*`. They complement each other, but `clojure.core/letfn` depends on the special form `letfn*`, and not the other way around.

There may be other macros which expand to `letfn*`, in the future, and we won't know what shape they will take. We don't worry about it, since those would be library features, not compiler features.

In Clojure, we conflate the runtime and the compiler so often, but there are still useful (and, I think, necessary) distinctions to be made. I don't see that we'll lose test coverage at all, by using only `fn*` here.

1. Compiler tests
    * Coverage of the special form `letfn*`
3. Library tests (clojure-test-suite)
    * Redundant coverage of `letfn*`, since all roads lead there anyway
    * Coverage of the `letfn` macro, its logic, and its dependencies

By adding any macro assumptions into the compiler tests, we end up with redundancy on both sides.

jeaye:
Applied all feedback, resolved conflicts, and merged. Thanks, Ambrose! Great work figuring this one out.

Issue #246, opened 2025-02-06
`(keyword nil)` should be `nil` not `:nil`


frenchy64:
```
clojure.core=> (keyword nil)
:nil

;; via clojure-test

FAIL in (nil) (nil:nil)
expected: (nil? (keyword nil))
  actual: (not (nil? :nil))
```

jeaye:
This is a side effect of #241, I believe.

PR #264, opened 2025-02-13
fix: (str nil) => ""

### LINKED ISSUE #241: `(str nil)` differs from Clojure
```
;; jank
clojure.core=> (str nil)
"nil"

Clojure 1.12.0
user=> (str nil)
""
```

bpiel:
Closes #241 

Here are a some notes on the approach I took and why it differed from what was expected for this ticket.

**The original behavior:**

```
clojure.core=> 
nil
clojure.core=> (str nil)
"nil"
clojure.core=> (str nil "hello" nil)
"nilhellonil"
clojure.core=> (println nil)
nil
nil
clojure.core=> (println "hello" nil nil)
hello nil nil
nil
clojure.core=> (prn "hello" nil nil)
"hello" nil nil
nil
clojure.core=> nil
nil
```

**First-attempt changes:**

```
// nil.cpp
...
  native_persistent_string const &nil::to_string() const
  {
    // static native_persistent_string const s{ "nil" };
    static native_persistent_string const s{ "" };
    return s;
  }

  native_persistent_string const &nil::to_code_string() const
  {
    // return to_string();
    static native_persistent_string const s{ "nil" };
    return s;
  }

  void nil::to_string(util::string_builder &buff) const
  {
    //fmt::format_to(std::back_inserter(buff), "nil");
    fmt::format_to(std::back_inserter(buff), "");
  }
...
```

**Behavior after first attempt**
```
clojure.core=> 
nil    ;; should be empty line
clojure.core=> (str nil)
""     ;; good
clojure.core=> (str nil "hello" nil)
"hello" ;; good
clojure.core=> (println "hello" nil nil)
hello  ;; should be: hello nil nil
nil
clojure.core=> (prn "hello" nil nil)
"hello" nil nil  ;; good
nil
clojure.core=> nil
nil    ;; good
```

After some time spent trying and failing to correct the above, I eventually went to the Clojure source:

Clojure `str`:
https://github.com/clojure/clojure/blob/clojure-1.11.1/src/clj/clojure/core.clj#L546

```
(defn str
  ...
  (^String [] "")
  (^String [^Object x]
   (if (nil? x) "" (. x (toString))))  ;; <<==== NOTE
  (^String [x & ys]
     ((fn [^StringBuilder sb more]
          (if more
            (recur (. sb  (append (str (first more)))) (next more))
            (str sb)))
      (new StringBuilder (str x)) ys)))
```

Taking inspiration from Clojure, I decided to implement this `nil` behavior in jank's `str` functions. In the results below, all but the first case show correct behavior. The failing case is not a regression.

**New behavior**
```
clojure.core=> 
nil  ;; should be empty line
clojure.core=> (str nil)
""
clojure.core=> (str nil "hello" nil)
"hello"
clojure.core=> (println nil)
nil
nil
clojure.core=> (println "hello" nil nil)
hello nil nil
nil
clojure.core=> (prn "hello" nil nil)
"hello" nil nil
nil
clojure.core=> nil
nil
```

Also, this had no effect on: 
keyword differs from CLJ/S for non-strings
https://github.com/jank-lang/jank/issues/246

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -930,14 +930,14 @@ namespace jank::runtime
 
   native_persistent_string str(object_ptr const o, object_ptr const args)
   {
+    util::string_builder buff;
+    buff.reserve(16);
+    if(!is_nil(o))
+    {
+      runtime::to_string(o, buff);
+    }
     return visit_seqable(
-      [=](auto const typed_args) -> native_persistent_string {
-        util::string_builder buff;
-        buff.reserve(16);
-        if(!is_nil(o))
-        {
-          runtime::to_string(o, buff);
-        }
+      [&](auto const typed_args) -> native_persistent_string {
         for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
         {
           auto const fst(it->first());
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -937,7 +937,7 @@ namespace jank::runtime
       runtime::to_string(o, buff);
     }
     return visit_seqable(
-      [&](auto const typed_args) -> native_persistent_string {
+      [](auto const typed_args, auto &buff) -> native_persistent_string {
         for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
         {
           auto const fst(it->first());
@@ -949,7 +949,8 @@ namespace jank::runtime
         }
         return buff.release();
       },
-      args);
+      args,
+      buff);
   }
 
   obj::persistent_list_ptr list(object_ptr const s)
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -937,7 +937,7 @@ namespace jank::runtime
       runtime::to_string(o, buff);
     }
     return visit_seqable(
-      [](auto const typed_args, auto &buff) -> native_persistent_string {
+      [](auto const typed_args, util::string_builder &buff) -> native_persistent_string {
         for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
         {
           auto const fst(it->first());
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -934,14 +934,19 @@ namespace jank::runtime
       [=](auto const typed_args) -> native_persistent_string {
         util::string_builder buff;
         buff.reserve(16);
-        runtime::to_string(o, buff);
+        if(!is_nil(o))
+        {
+          runtime::to_string(o, buff);
+        }
         if(0 < sequence_length(typed_args))
         {
           auto const fresh(typed_args->fresh_seq());
-          runtime::to_string(fresh->first(), buff);
-          for(auto it(fresh->next_in_place()); it != nullptr; it = it->next_in_place())
+          for(auto it(fresh); it != nullptr; it = it->next_in_place())
           {
-            runtime::to_string(it->first(), buff);
+            if(!is_nil(it->first()))
+            {
+              runtime::to_string(it->first(), buff);
+            }
           }
         }
         return buff.release();
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -938,18 +938,16 @@ namespace jank::runtime
         {
           runtime::to_string(o, buff);
         }
-        if(0 < sequence_length(typed_args))
-        {
           for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
           {
             auto const fst(it->first());
-            if(!is_nil(fst))
+            if(is_nil(fst))
             {
-              runtime::to_string(fst, buff);
+	      continue;
             }
+	    runtime::to_string(fst, buff);
           }
-        }
-        return buff.release();
+          return buff.release();
       },
       args);
   }
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -938,16 +938,16 @@ namespace jank::runtime
         {
           runtime::to_string(o, buff);
         }
-          for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
+        for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
+        {
+          auto const fst(it->first());
+          if(is_nil(fst))
           {
-            auto const fst(it->first());
-            if(is_nil(fst))
-            {
-	      continue;
-            }
-	    runtime::to_string(fst, buff);
+            continue;
           }
-          return buff.release();
+          runtime::to_string(fst, buff);
+        }
+        return buff.release();
       },
       args);
   }
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -940,12 +940,12 @@ namespace jank::runtime
         }
         if(0 < sequence_length(typed_args))
         {
-          auto const fresh(typed_args->fresh_seq());
-          for(auto it(fresh); it != nullptr; it = it->next_in_place())
+          for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
           {
-            if(!is_nil(it->first()))
+            auto const fst(it->first());
+            if(!is_nil(fst))
             {
-              runtime::to_string(it->first(), buff);
+              runtime::to_string(fst, buff);
             }
           }
         }
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -252,7 +252,9 @@
     ([]
      "")
     ([o]
-     (clojure.core-native/to-string o))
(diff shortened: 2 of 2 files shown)


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -934,14 +934,19 @@ namespace jank::runtime
...
+        }
         if(0 < sequence_length(typed_args))
         {
           auto const fresh(typed_args->fresh_seq());
-          runtime::to_string(fresh->first(), buff);
-          for(auto it(fresh->next_in_place()); it != nullptr; it = it->next_in_place())
+          for(auto it(fresh); it != nullptr; it = it->next_in_place())
           {
-            runtime::to_string(it->first(), buff);
+            if(!is_nil(it->first()))
```
frenchy64, line 946: Could you pull this into a variable and reuse it?

```
auto const fst(it->first());
```

And create the fresh seq in the for loop to prevent accidents?

```
          for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
```

frenchy64 (changes requested):

frenchy64:
Thanks for tackling this Bill, could you fix these stylistic issues to help me see the approach without getting distracted?

frenchy64:
Also please add `Closes #241` to the top of the PR description.

bpiel:
@frenchy64 Ugh.. maybe this one test is stuck? Anyway to restart it?

jeaye:
> @frenchy64 Ugh.. maybe this one test is stuck? Any way to restart it?

What's stuck?

bpiel:
@jeaye Nothing now. It was running for 26 or so mins when I sent that. It probably resolved as soon as I did, of course.
thanks

jeaye:
> @jeaye Nothing now. It was running for 26 or so mins when I sent that. It probably resolved as soon as I did, of course. thanks

Static analysis is very slow. It's amazing what it finds, though. :slightly_smiling_face:

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -934,14 +934,19 @@ namespace jank::runtime
       [=](auto const typed_args) -> native_persistent_string {
         util::string_builder buff;
         buff.reserve(16);
-        runtime::to_string(o, buff);
+        if(!is_nil(o))
+        {
+          runtime::to_string(o, buff);
+        }
         if(0 < sequence_length(typed_args))
```
frenchy64, line 941: Question for @jeaye: can we remove this conditional? `persistent_list::fresh_seq()` will return `nullptr` and short-circuit the `for`, then we just need to check the count once.

jeaye: Yeah, that makes sense. The for loop with `fresh_seq` should handle this just fine.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -252,7 +252,9 @@
     ([]
      "")
     ([o]
-     (clojure.core-native/to-string o))
+     (if (nil? o)
+       ""
+       (clojure.core-native/to-string o)))
```
jeaye, line 257: We're already doing the `nil` check in the C++ code. The way I read it, this extra `nil?` check is not needed. Is that right?

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -934,14 +934,19 @@ namespace jank::runtime
...
-          runtime::to_string(fresh->first(), buff);
-          for(auto it(fresh->next_in_place()); it != nullptr; it = it->next_in_place())
+          for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
           {
-            runtime::to_string(it->first(), buff);
+            auto const fst(it->first());
+            if(!is_nil(fst))
+            {
+              runtime::to_string(fst, buff);
+            }
```
jeaye, line 949: Nit: In general, we try to keep the primary code path as unnested as possible. This means any checks we do are not for our happy path, but rather all of our early escapes. To model that here, we would check for nil and `continue`. Then, after that, we know we have non-nil.

In this case, it actually results in  one more line of code, but in general, it leads to a very clear pattern of checking your invariants first and then growing your understanding of the data as you progress. For a larger example of this, see: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/analyze/processor.cpp#L529

I do this so much that this code stands out to me, since it's the primary code path, but it's nested.

jeaye (changes requested):
Nice work finding your way when I led you down the wrong path, Bill! :slightly_smiling_face:

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
frenchy64: What's your take on using scopes to manage this kind of control flow where the variable isn't relevant outside the main branch? More like:

https://github.com/jank-lang/jank/blob/5d792669108fb89555e311f8542ced2522d4e430/compiler%2Bruntime/src/cpp/jank/runtime/ns.cpp#L90

For example, here we could use:

```c++
  object_ptr not_nil(object_ptr const o)
  {
    return is_nil(o) ? nullptr : o;
  }

....
        for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
        {
          if(auto const fst = not_nil(it->first()))
          {
            runtime::to_string(fst, buff);
          }
        }
```

or

```c++
          if(auto const fst(it->first()); !is_nil(fst))
```

jeaye: Not a fan. 

1. It doesn't allow us to use direct initialization
2. It nests the primary path of the code. This is the main thing we're trying to avoid.
3. It moves the non-primary path to the bottom, which may not be seen

Nesting in lisps is different. We need to nest even to introduce a local (via `let`). In C-like langs, though, nesting means even more cognitive complexity; if you have a big `if`, for the primary path, followed by an else at the bottom, we may miss that else. For some interfaces, it becomes necessary. In general, I would rather keep things flat and have all of my invariants checked directly above what I'm doing.

Keeping things flat and asserting invariants in short branches is one of the best patterns I've found. Another way to think about it is this: keep your `if` bodies as small as possible. How do we do that? We only use `if` for early returns, breaks, and continues. Everything else stays outside of an `if`.

frenchy64: Thanks. And are higher-order functions too expensive to use for this kind of thing? My impression was that the reason we use `is_some` + `unwrap` to deal with options is to avoid introducing closures.

jeaye: Yeah, this is about optimizing both compile-time and run-time perf. More fns means more to compile. I'd like to get jank's compilations DOWN. We just had to bump the CI analysis to 35 minutes, as you know. While normal builds are still under 10m in CI, we can still do much better. That means limiting our includes, using forward declarations, keeping headers as light as possible, only using templates where needed, avoiding object visiting where possible, and so on.

compiler+runtime/src/jank/clojure/core.jank
bpiel: The trick here is that this line is calling `to-string`, which does not have the `nil` check. I added the `nil` check to `str`. So, I think this is right.

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -934,15 +934,18 @@ namespace jank::runtime
...
-        runtime::to_string(o, buff);
-        if(0 < sequence_length(typed_args))
+        if(!is_nil(o))
         {
-          auto const fresh(typed_args->fresh_seq());
-          runtime::to_string(fresh->first(), buff);
-          for(auto it(fresh->next_in_place()); it != nullptr; it = it->next_in_place())
+          runtime::to_string(o, buff);
+        }
+        for(auto it(typed_args->fresh_seq()); it != nullptr; it = it->next_in_place())
```
frenchy64, line 941: Maybe not urgent, but I think everything above the `for` could be pulled out the visit to reduce the generated code. Then the `buff` could be passed as an argument.

frenchy64 (approved):

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
frenchy64: I have no idea if it makes a difference in practice, but I've seen other parts of the code pass in the argument explicitly rather than make a closure.

https://github.com/jank-lang/jank/blob/1a75d85ecd8e63cc8c13de06c97617eb6a1f8998/compiler%2Bruntime/src/cpp/jank/runtime/core.cpp#L27-L40

bpiel:
@frenchy64 You got your wish!

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
bpiel: oh, just saw this follow-up comment

frenchy64: Thanks @bpiel! I'm more worried about generated code size here but maybe @jeaye has a preference between the closure and explicit arg.

jeaye: There are two approaches here:

1. Use a closure (which Bill is currently doing with `[&]`)
2. Pass the explicit arg

The difference impacts runtime perf. The reason is that the closure generates to a struct when the C++ compiler does its thing. This struct has members for the captured objects. Calling the closure requires creating an instance of that struct, passing in the captured data, and then calling the call operator on that instance. This becomes harder to inline.

On the other hand, if we just use a `[](auto const typed_args, util::string_builder &buff)` then we can pass `buff` in as an arg. The C++ compiler will generate a free function which can be more easily inlined. We could use `auto &buff`, but that means implicitly adding a second type parameter, which then impacts compile-time perf more.

Overall, the difference between the two is marginal and the cases where I've used the second have been due to the profiler saying it'd help. Given that `str` is a very common function, I think it's a good bet to default to the faster option, since we're talking about it, but I probably wouldn't have put all of that info on Bill if it hadn't been brought up already.

PR #266, opened 2025-02-13
Add initial error reporting framework

jeaye:
This branch is getting very far from main, which used to work fine when it was just me, but now we have a lot of people changing main each week. I've decided to merge this in batches.

## Included in this PR
* New error namespace, types, helpers
* Lexer completely reworked to use new errors
* Parser completely reworked to use new errors
* Analysis partially reworked to use new errors
* New error report namespace
* Lexer-based highlighter
* CMake summary output on `./configure`

## Remaining work
* Add all necessary source info as meta during parsing
* Finish rework of analysis error handling
* Polish error reporting (jar support, horizontal centering, various edge cases, etc)
* Add a test suite for the error reporter
* Tackle the whole world of runtime errors


compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1215,7 +1206,9 @@ namespace jank::read::parse
           auto const resolved_ns(__rt_ctx->resolve_ns(make_box<obj::symbol>(ns_portion)));
           if(resolved_ns.is_none())
           {
-            return err(error{ token.pos, fmt::format("unknown namespace: {}", ns_portion) });
+            return error::parse_unresolved_namespace(
+              fmt::format("Unresolved namespace '{}'", ns_portion),
+              { start_token.start, latest_token.end });
```
frenchy64, line 1211: FYI this case might not be an error due to https://github.com/jank-lang/jank/issues/195, but I'm working on that.

I have this as

```
            ns = ns_portion;
```

in my WIP branch, but I don't have an example handy:

https://github.com/jank-lang/jank/pull/239/files#diff-bced15263f03bb66a379b6f9346fa47a66e15b168e39a710e7153d1c115090e2R1222

compiler+runtime/src/cpp/main.cpp
```diff
@@ -141,6 +144,12 @@ namespace jank
     le.setPrompt(get_prompt("=> "));
     native_transient_string input{};
 
+    /* We write every REPL expression to a temporary file, which then allows us
+     * to later review that for error reporting. We automatically clean it up
+     * and we reuse the same file over and over. */
+    auto const tmp{ boost::filesystem::temp_directory_path() };
+    auto const path{ tmp / boost::filesystem::unique_path("jank-repl-%%%%.jank") };
```
frenchy64, line 151: Thank you!

frenchy64 (commented):
Wow! First skim, I haven't looked at the tests or tried it yet.

compiler+runtime/include/cpp/jank/util/type_name.hpp
```diff
@@ -0,0 +1,76 @@
...
+   * special handling, depending on the compiler, to know how to pull out the type from
+   * the string.
+   *
+   * With that done, we just do some template work to turn the string literal into a
+   * constexpr static variable to which we can have a string_view. Entirely at compile-time.
+   * However, we only want to keep the portion we care about in the resulting binary, not
+   * the whole pretty function string, so we copy it into an array which contains only
+   * the type name.
+   *
+   * Just do type_name<T>() and there's your string_view. */
```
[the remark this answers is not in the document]
jeaye: I said that already, in the second to last paragraph.

compiler+runtime/src/cpp/jank/error/report.cpp
```diff
@@ -0,0 +1,495 @@
...
+      }
+    }
+
+    size_t last_number{};
+    for(size_t i{}; i < lines.size(); ++i)
+    {
+      /* Remove ellipsis if needed. */
+      if(lines[i].kind == line::kind::ellipsis)
+      {
+        /* We can be confident there's no note right after an ellipsis. */
```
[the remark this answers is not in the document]
jeaye: The top margin guarantees this.

compiler+runtime/src/cpp/jank/read/parse.cpp
jeaye: Gotcha. Leaving as-is for now. Can be fixed in your PR.

compiler+runtime/src/cpp/main.cpp
jeaye: Was this bothering you somehow? :grin:

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -630,13 +622,13 @@ namespace jank::read::parse
     }
     else if(list_result.expect_ok().is_none())
     {
-      return err(
-        error{ start_token.pos, native_persistent_string{ "value after #( must be present" } });
+      return error::parse_invalid_shorthand_function({ start_token.start, latest_token.end },
+                                                     "Value after #( must be present");
```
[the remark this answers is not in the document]
jeaye: I haven't combed through every parse error yet, to verify they're all reachable. I think this one is not reachable and so should be either removed or changed to an internal parse error instead.

jeaye:
> Wow! First skim, I haven't looked at the tests or tried it yet.

Thanks for the review. No requirement to try it locally. You're welcome to, though.

frenchy64:
@jeaye just tried it, it's great. I can look at the tests another time.

frenchy64 (approved):

compiler+runtime/src/cpp/main.cpp
frenchy64: I found this file littered around my directories at some point.

jeaye: Hmmm. We weren't writing to a file before. Must've been from something else.

frenchy64: Oh, probably `.jank-repl-history`.

jeaye: Ah, yes. You'll see that. We don't clean them up. Leiningen does the same with `.lein-repl-history`.

frenchy64: Never been bothered by Leiningen's behavior here. Looks like it uses the project-root if project.clj is present, otherwise uses a config directory

https://github.com/technomancy/leiningen/blob/24fb93936133bd7fc30c393c127e9e69bb5f2392/src/leiningen/repl.clj#L152-L153

jeaye: Yeah, I'd prefer if we did something like that, too. Not sure how we'd want to do it, though. We could just use jank's cache dir, which is in `~/.cache/jank` by default. That would be a global history for all projects.

frenchy64: Yes maybe jank should take a path to save configuration, that can be set by e.g., lein-jank.

PR #270, opened 2025-02-16
feat(compilation): compilation directory depends on more cli flags

Samy-33:
on optimization level, include dirs and the macro defines

Related to https://github.com/jank-lang/jank/issues/139

```clojure
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ rm -rf ~/.cache/jank/*
zsh: sure you want to delete all 4 files in /home/saket/.cache/jank [yn]? y
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ ls ~/.cache/jank
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ cat dev/src/a.jank
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: dev/src/a.jank
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ (ns a
   2   │    (:require b))
   3   │
   4   │  (println "Loading a")
   5   │
   6   │  (defmacro load-with-delay [& body]
   7   │    (println "Waiting for 2000ms")
   8   │    (clojure.core-native/sleep 2000)
   9   │    `(do ~@body))
  10   │
  11   │  (load-with-delay
  12   │    (println "Loading with dealy")
  13   │    (b/hello))
  14   │
  15   │  (println "Loaded a")
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ cat dev/src/b.jank
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: dev/src/b.jank
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ (ns b)
   2   │
   3   │ (defn hello [] (println "Hello, world!"))
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ time ./build/jank -O 2 --module-path ./dev/src compile a
Bottom of clojure.core
Loading a
Waiting for 2000ms
Loading with dealy
Hello, world!
Loaded a
./build/jank -O 2 --module-path ./dev/src compile a  0.24s user 0.04s system 11% cpu 2.274 total
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ time ./build/jank -O 2 --module-path ./dev/src compile a
Bottom of clojure.core
Loading a
Loading with dealy
Hello, world!
Loaded a
./build/jank -O 2 --module-path ./dev/src compile a  0.49s user 0.04s system 99% cpu 0.530 total
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ ls ~/.cache/jank
x86_64-pc-linux-gnu-7447f407ffb4bafe59ff60f5a258966c3c737e30c91c43ce817d4d17fad2a15f
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ ls ~/.cache/jank/x86_64-pc-linux-gnu-7447f407ffb4bafe59ff60f5a258966c3c737e30c91c43ce817d4d17fad2a15f
classes
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ ls ~/.cache/jank/x86_64-pc-linux-gnu-7447f407ffb4bafe59ff60f5a258966c3c737e30c91c43ce817d4d17fad2a15f/classes
a.o  b.o
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ time ./build/jank -O 1 --module-path ./dev/src compile a
Bottom of clojure.core
Loading a
Waiting for 2000ms
Loading with dealy
Hello, world!
Loaded a
./build/jank -O 1 --module-path ./dev/src compile a  0.38s user 0.04s system 17% cpu 2.418 total
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ tree ~/.cache/jank
/home/saket/.cache/jank
├── x86_64-pc-linux-gnu-7447f407ffb4bafe59ff60f5a258966c3c737e30c91c43ce817d4d17fad2a15f
│   └── classes
│       ├── a.o
│       └── b.o
└── x86_64-pc-linux-gnu-cbbfd6ecd45e15eb3c16b9c59f1c5ebed2d457381af01a6a2f08271b114084a8
    └── classes
        ├── a.o
        └── b.o

5 directories, 4 files
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ time ./build/jank -O 1 --module-path ./dev/src compile a
Bottom of clojure.core
Loading a
Loading with dealy
Hello, world!
Loaded a
./build/jank -O 1 --module-path ./dev/src compile a  0.39s user 0.03s system 99% cpu 0.430 total
➜  compiler+runtime git:(feat/binary-cache-dir) ✗ time ./build/jank -O 1 --module-path ./dev/src compile a
Bottom of clojure.core
Loading a
Loading with dealy
Hello, world!
Loaded a
./build/jank -O 1 --module-path ./dev/src compile a  0.28s user 0.04s system 99% cpu 0.317 total
```


jeaye (changes requested):
I can see why you were asking me questions about these two directories. Everything we have in there is confusing. :joy: 

This is what we want, going forward:

1. Compiled object files ALWAYS are written to the binary cache dir; there is no option to change that
2. The CLI option is for changing where the final executable is stored; I think we should just rename this to `-o <file>` for the CLI command which compiles binaries. We don't need to specify a dir or anything.

I think that means we can rename `output_dir` in RT ctx to `binary_cache_dir`, to be super clear of the intention. Then we can remove the `--output-dir` flag altogether. Later, when we support compiling executables, we can add in the `-o <file>` flag for that command.

compiler+runtime/src/cpp/jank/util/dir.cpp
```diff
@@ -61,15 +62,20 @@ namespace jank::util
...
     static native_persistent_string res;
     if(!res.empty())
     {
       return res;
     }
 
-    return res = fmt::format("{}/{}", user_cache_dir(), binary_version());
+    return res = fmt::format("{}/{}",
+                             user_cache_dir(),
+                             binary_version(optimization_level, includes, defines));
```
jeaye, line 78: Hm, I was wrong, Saket. This is not going to work. If two separate projects both have ns `a`, with different code, then compiling one project first, then going to compile the other project could result in loading the first project's `a` when you go to compile the second.

This could work fine for dependencies from maven, but we don't have a good way to distinguish them from project sources right now. To mitigate this, let's put the binary cache dir in `target/<binary version>`, relative to the current directory. This is what leiningen does. So we should just be able to replace `user_cache_dir` here with `"target"`.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/util/dir.cpp
Samy-33: Uhuh. That is a good observation. Thanks for catching.

Dependency and project modules could also conflict. But I think it can be the developer's responsibility to choose unique prefixes for their namespaces.

Not as important to handle this within language than other features, at the moment.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -41,7 +41,10 @@ namespace jank::runtime
 
   context::context(util::cli::options const &opts)
     : jit_prc{ opts }
-    , output_dir{ opts.compilation_path }
+    , binary_cache_dir{ fmt::format(
+        "{}/{}",
+        util::binary_cache_dir(opts.optimization_level, opts.include_dirs, opts.define_macros),
+        "classes") }
```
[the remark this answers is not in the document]
Samy-33: Done.

jeaye:
Thanks, Saket!

PR #278, opened 2025-02-18
fix(lex): fixes error while lexing strings with unicode chars

### LINKED ISSUE #277: unicode string does not work
```
❯ build/jank repl
Bottom of clojure.core
clojure.core=> "你好"
─ lex/unterminated-string ───────────────────────────────────────────────────────
error: Unterminated string

╭ /var/folders/qd/sl94n12j2g560g4c4my84mdr00v4yd/T/jank-repl-44cd.jank ──────────╮
│1│"你好"                                                                         │
│ │^^ Unterminated string                                                        │
╰─┴──────────────────────────────────────────────────────────────────────────────╯

clojure.core=> "nihao"
"nihao"


clojure.core=>
```

Samy-33:
fixes #277

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -1079,7 +1079,7 @@ namespace jank::read::lex
             {
               escaped = contains_escape = true;
             }
-            ++pos;
+            pos += oc.expect_ok().len;
           }
           require_space = true;
           ++pos;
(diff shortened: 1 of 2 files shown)


jianlingzhong:
can you also add a jank test file?

jeaye:
> can you also add a jank test file?

`.jank` files aren't used for the lexer. The test which is included is correct.

Is there something else you wanted captured by adding a `.jank` file?

jeaye (approved):

jeaye:
Thanks for the quick fix!

jianlingzhong:
jank file can be the final check if unicode actually works or not.

jeaye:
> jank file can be the final check if unicode actually works or not.

I hear what you're saying. There's no difference between doing it in C++ versus in a .jank file, though. We want to minimize test duplication unless it's providing additional coverage in the code.

PR #284, opened 2025-02-25
Support conj'ing nil onto maps and merging any combination of maps

### LINKED ISSUE #261: invalid map entry: nil
```
;; jank
clojure.core=> (conj {} nil)
Exception: invalid map entry: nil

;; clojure
user=> (conj {} nil)
{}
```

frenchy64:
Closes #261 

- [ ] figure out how to call `runtime::assoc` template in `runtime::merge`

```
clojure.core=> (into {} [nil {1 2} [2 3] (sorted-map 3 4)])
{1 2, 3 4, 2 3}

clojure.core=> (conj {} nil)
{}
```


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -707,44 +707,34 @@ namespace jank::runtime
...
   {
-    return visit_object(
-      [&](auto const typed_m) -> object_ptr {
-        using T = typename decltype(typed_m)::value_type;
-
-        if constexpr(behavior::associatively_writable<T>)
+    return visit_map_like(
+      [&](auto const typed_other, auto const zero, auto const one) {
+        object_ptr ret{ m };
+        for(auto seq{ typed_other->fresh_seq() }; seq != nullptr; seq = seq->next_in_place())
```
jeaye, line 713: We've lost the ability to merge nil with this, I believe.

compiler+runtime/src/cpp/jank/runtime/obj/detail/base_persistent_map.cpp
```diff
@@ -1,7 +1,12 @@
+#include <fmt/format.h>
+#include <jank/native_persistent_string/fmt.hpp>
```
jeaye, line 2: I see the reasoning for grouping these, but that's not how we roll. Please keep the jank includes together, with fmt above them, separated by a new line.

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -707,44 +707,34 @@ namespace jank::runtime
...
+  {
+    return visit_map_like(
+      [&](auto const typed_other) {
+        object_ptr ret{ m };
+        for(auto seq{ typed_other->fresh_seq() }; seq != nullptr; seq = seq->next_in_place())
         {
-          throw std::runtime_error{ fmt::format("not associatively writable: {}",
-                                                typed_m->to_string()) };
+          auto const e{ seq->first() };
+          ret = assoc_in_place(ret, runtime::nth(e, make_box(0)), runtime::nth(e, make_box(1)));
```
jeaye, line 733: `e` should be a `obj::persistent_vector_ptr` here. We can do `e->data[0]`.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -241,7 +241,7 @@ namespace jank::analyze
     auto const keys_exprs{ visit_map_like(
       [&](auto const typed_imap_obj) -> string_result<keys_and_exprs> {
         keys_and_exprs ret{};
-        for(auto seq{ typed_imap_obj->seq() }; seq != nullptr; seq = seq->next())
+        for(auto seq{ typed_imap_obj->fresh_seq() }; seq != nullptr; seq = seq->next_in_place())
         {
           auto const e{ seq->first() };
           auto const k_obj{ runtime::nth(e, make_box(0)) };
```
jeaye, line 247: `e` should be a `obj::persistent_vector_ptr` here. We can do `e->data[0]`.

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -707,44 +707,34 @@ namespace jank::runtime
...
-              }
-              else
-              {
-                throw std::runtime_error{ fmt::format("not associatively readable: {}",
-                                                      typed_m->to_string()) };
-              }
-            },
-            other);
+          auto const e{ seq->first() };
+          ret = assoc(ret, runtime::nth(e, zero), runtime::nth(e, one));
```
jeaye, line 716: `e` should be a `obj::persistent_vector_ptr` here. We can do `e->data[0]`.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
frenchy64: I think previously you'd get an error if `m == nil` (not `associatively_writeable`) and an error if `other == nil` (not `associatively_readable`).

It works to not handle nil because `clojure.core/merge` actually calls `typed_s->conj`, which then calls `runtime::merge` on `this` only if the argument is a map.

Similar for `merge_in_place` via `typed_m->conj_in_place`.

frenchy64: `nil` is `associatively_readable` but the previous logic hardcoded the maps instead of checking for the concept.

frenchy64: Ah, but it is `associatively_writable`. I will add some checks, but I'm not sure if there's a code path to triggering them.

jeaye:
Is there any more work you wanted to do on this PR? Looks like you addressed my latest feedback, but I hadn't taken a look since another review wasn't requested. If you're done with this, I can resolve the conflicts.

frenchy64:
I think the only thing left is a FIXME about using `auto` in one of the template calls to improve the C++ type information. If you have the time, please take it across the line.

PR #286, opened 2025-02-25
feat: aot compile cpp

### LINKED ISSUE #139: Make module system more robust

Samy-33:
closes #139 
[asciicast](https://asciinema.org/a/rnczCULaw2txXMfw7VI8iSSfD)

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/context.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/context.cpp
@@ -207,9 +207,12 @@ namespace jank::runtime
   void context::eval_cpp_string(native_persistent_string_view const &code) const
   {
     profile::timer const timer{ "rt eval_cpp_string" };
+
+    /* TODO: Handle all the errors here to avoid exceptions. Also, return a message that
+     * is valuable to the user. */
     auto &partial_tu{ jit_prc.interpreter->Parse({ code.data(), code.size() }).get() };
 
-    /* Writing the module before executing it because `llvm::Interpreter#Execute`
+    /* Writing the module before executing it because `llvm::Interpreter::Execute`
      * moves the `llvm::Module` held in the `PartialTranslationUnit`. */
     if(truthy(compile_files_var->deref()))
     {
@@ -218,7 +221,6 @@ namespace jank::runtime
     }
 
     auto err(jit_prc.interpreter->Execute(partial_tu));
-    llvm::logAllUnhandledErrors(std::move(err), llvm::errs(), "error: ");
   }
 
   object_ptr context::read_string(native_persistent_string_view const &code)
--- a/compiler+runtime/src/cpp/jank/runtime/context.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/context.cpp
@@ -315,7 +315,7 @@ namespace jank::runtime
                                             std::unique_ptr<llvm::Module> const &module) const
   {
     profile::timer const timer{ fmt::format("write_module {}", module_name) };
-    boost::filesystem::path const module_path{
+    std::filesystem::path const module_path{
       fmt::format("{}/{}.o", binary_cache_dir, module::module_to_path(module_name))
     };
     std::filesystem::create_directories(module_path.parent_path());
--- a/compiler+runtime/src/cpp/jank/runtime/context.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/context.cpp
@@ -197,11 +197,28 @@ namespace jank::runtime
       fn->unique_name = fn->name;
       codegen::llvm_processor cg_prc{ wrapped_exprs, module, codegen::compilation_target::module };
       cg_prc.gen().expect_ok();
-      write_module(std::move(cg_prc.ctx)).expect_ok();
+      write_module(cg_prc.ctx->module_name, cg_prc.ctx->module).expect_ok();
     }
 
     assert(ret);
     return ret;
+  }
+
+  void context::eval_cpp_string(native_persistent_string_view const &code) const
+  {
+    profile::timer const timer{ "rt eval_cpp_string" };
+    auto &partial_tu{ jit_prc.interpreter->Parse({ code.data(), code.size() }).get() };
+
+    /* Writing the module before executing it because `llvm::Interpreter#Execute`
+     * moves the `llvm::Module` held in the `PartialTranslationUnit`. */
+    if(truthy(compile_files_var->deref()))
+    {
+      auto module_name{ runtime::to_string(current_module_var->deref()) };
+      write_module(module_name, partial_tu.TheModule).expect_ok();
+    }
+
+    auto err(jit_prc.interpreter->Execute(partial_tu));
+    llvm::logAllUnhandledErrors(std::move(err), llvm::errs(), "error: ");
   }
 
   object_ptr context::read_string(native_persistent_string_view const &code)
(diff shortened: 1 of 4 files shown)


compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -197,13 +197,30 @@ namespace jank::runtime
...
+    /* Writing the module before executing it because `llvm::Interpreter#Execute`
+     * moves the `llvm::Module` held in the `PartialTranslationUnit`. */
+    if(truthy(compile_files_var->deref()))
+    {
+      auto module_name{ runtime::to_string(current_module_var->deref()) };
+      write_module(module_name, partial_tu.TheModule).expect_ok();
+    }
+
+    auto err(jit_prc.interpreter->Execute(partial_tu));
+    llvm::logAllUnhandledErrors(std::move(err), llvm::errs(), "error: ");
```
jeaye, line 221: So here, we just swallow the error and write to stderr. I think it's important to return this error, rather than just make noise about it. If we fail to parse, this will throw, but if we fail to execute, this should behave the same. Either they both return errors or they both throw.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -561,7 +567,7 @@ namespace jank::runtime::module
         return err(
           fmt::format("unable to map file {} due to error: {}", entry.path, file.expect_err()));
       }
-      rt_ctx.jit_prc.eval_string({ file.expect_ok().head, file.expect_ok().size });
+      rt_ctx.eval_cpp_string({ file.expect_ok().head, file.expect_ok().size });
```
jeaye, line 570: When we were talking in Slack, we noted how the `load_cpp` fn needs to call the `jank_load_` fn. Let's include that in this PR, please, since `.cpp` modules won't work properly without it (until they're AOT compiled).

jeaye (changes requested):
Exciting to see this! Moves us another step toward full AOT jank programs. :muscle:

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
Samy-33: `extern "C"`s in the cpp module open the same hole that we covered by introducing the symbol counters in jank sources.

Suppose user has two modules `foo.cpp` and `bar.cpp` and they both define a function `hello_world`

```cpp
// foo.cpp
extern "C" void hello_world() {
    cout<<"Hello from foo\n";
}
```

```cpp
// bar.cpp
extern "C" void hello_world() {
    cout<<"Hello from bar\n";
}
```

Trying to load both modules causes error because of duplicate definition of the `hello_world` function. Reloading a, already loaded C++ module, causes the same error.

Not sure how big of a problem it is. People may have conflicting extern defines that conflict with a library defines?
This PR leaves that problem open as of now.

Samy-33:
Pending issues:
- `(require 'cpp-module :reload)` doesn't seem to work.
- Error handling during JIT parsing or execution of CPP source.
- Handling conflicting defines in modules backed by CPP sources.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
jeaye: Yep, this would be a problem with *any* C++ program, so it's not going to be a surprise to any C++ dev.

jeaye (approved):
Nice job, Saket!

PR #289, opened 2025-02-28
Replace `boost::variant` for AST expressions with custom type dispatch

jeaye:
Growing pains. We ran up against the limit of `boost::variant` for our AST nodes (expressions). To replace that, we now have a very similar mechanism to RT objects, with a similar visiting pattern. Rather than duplicating the `expect_object`, `dyn_cast`, etc, I hooked into LLVM's RTTI system by specializing `llvm::isa_impl`. This enables [LLVM's custom RTTI implementation](https://llvm.org/docs/ProgrammersManual.html#isa) while using jank's internal representation. If this goes well, we can do the same for RT objects easily.

This new design doesn't rely on templates for expressions, so a great deal of code could be moved into `.cpp` files. We can rely much more on forward declarations now, so that has been done. We also use expression pointers pretty much everywhere, rather than values. This makes (shallow) copying expressions faster, but it also removes the illusion that any deep copying was happening at all, since it never was.


compiler+runtime/src/cpp/jank/analyze/expr/do.cpp
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+
+  object_ptr do_::to_runtime_data() const
+  {
+    object_ptr body_maps{ make_box<obj::persistent_vector>() };
+    for(auto const &e : values)
+    {
+      visit_expr(
+        [&](auto const typed_e) { body_maps = conj(body_maps, typed_e->to_runtime_data()); },
```
frenchy64, line 45: I think we could use `body_maps->conj` directly here if we preserved its type. Ditto the other `to_runtime_data()` functions.

compiler+runtime/src/cpp/jank/analyze/expr/if.cpp
```diff
@@ -0,0 +1,44 @@
...
+    , condition{ condition }
+    , then{ then }
+    , else_{ else_ }
+  {
+  }
+
+  void if_::propagate_position(expression_position const pos)
+  {
+    position = pos;
+    if(then)
```
frenchy64, line 24: Redundant test?

compiler+runtime/src/cpp/jank/analyze/expr/if.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+  object_ptr if_::to_runtime_data() const
+  {
+    return merge(expression::to_runtime_data(),
+                 obj::persistent_array_map::create_unique(make_box("condition"),
+                                                          jank::detail::to_runtime_data(*condition),
+                                                          make_box("then"),
+                                                          jank::detail::to_runtime_data(*then),
+                                                          make_box("else"),
+                                                          jank::detail::to_runtime_data(else_)));
```
frenchy64, line 42: Do we still want to assoc if `else_` is none? Ditto all the other `option<expression_ptr>` cases for other expressions.

compiler+runtime/src/cpp/jank/evaluate.cpp
```diff
@@ -14,32 +14,54 @@
...
+#include <jank/analyze/expr/recursion_reference.hpp>
+#include <jank/analyze/expr/named_recursion.hpp>
+#include <jank/analyze/expr/local_reference.hpp>
+#include <jank/analyze/expr/let.hpp>
+#include <jank/analyze/expr/do.hpp>
+#include <jank/analyze/expr/if.hpp>
+#include <jank/analyze/expr/throw.hpp>
+#include <jank/analyze/expr/try.hpp>
+#include <jank/analyze/expr/case.hpp>
+#include <jank/analyze/visit.hpp>
```
frenchy64, line 37: IIUC I think this include makes the `/expr/*.hpp` includes redundant, but maybe they're  here for namespacing reasons.

frenchy64 (approved):
Much nicer!

compiler+runtime/src/cpp/jank/analyze/expr/if.cpp
jeaye: Yep, it'll render as `"none"`.

frenchy64: Ah, does that mean none and an option containing the string `"none"` render the same?

frenchy64 (approved):

compiler+runtime/src/cpp/jank/analyze/expr/if.cpp
jeaye:
```cpp
  template <typename T>
  object_ptr to_runtime_data(option<T> const &m)
  {
    return obj::persistent_array_map::create_unique(
      make_box("__type"),
      make_box("option"),
      make_box("data"),
      (m.is_none() ? make_box("none") : to_runtime_data(m.unwrap())));
  }
```

frenchy64: Looks like it? Might be an issue when converting data back to expressions.

jeaye: We don't convert data back to expressions. The whole `to_runtime_data` mechanism is for debugging AST expressions, used only by me. :joy: 

Once we can have a good Portal integration, hopefully turning jank's compiler data into Clojure data will prove much more useful.

PR #290, opened 2025-02-28
Enable ccache in CI

jeaye:
This should drastically cut our CI job times, hopefully allowing us to run CI on PR creation and synchronization, as we were before.


Samy-33:
We are caching a particular version of each apt dependency. Does the "included action" update the cache with the new versions when they are available?

My hunch is that it will be rare for an apt dependency to break building Jank. But we might want to periodically evict the cache to get new versions. This will keep us informed if a new version of a dependency breaks Jank.

jeaye:
> We are caching a particular version of each apt dependency. Does the "included action" update the cache with the new versions when they are available?
> 
> My hunch is that it will be rare for an apt dependency to break building Jank. But we might want to periodically evict the cache to get new versions. This will keep us informed if a new version of a dependency breaks Jank.

You're correct. Github has its own cache eviction mechanism, but we may still end up in a situation where an upstream dependency will break jank but CI doesn't get it. I'm open to solutions for that, going forward, but not trying to solve everything here.

PR #291, opened 2025-03-03
fix: nth should not respond to unordered collections

bpiel:
Closes https://github.com/jank-lang/jank/issues/211

```
clojure.core=> (nth {1 2} 0)
Exception: not ordered: persistent_array_map

clojure.core=> (nth #{1 2} 0)
Exception: not ordered: persistent_hash_set

clojure.core=> (nth [1] 0)
1

clojure.core=> (nth [] 0 123)
123

clojure.core=> (nth '(1 2) 0)
1

clojure.core=> (nth "abc" 1)
\b
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -790,6 +790,10 @@ namespace jank::runtime
         }
         else if constexpr(behavior::seqable<T>)
         {
+          if constexpr(!behavior::sequential<T> && !std::is_same_v<T, obj::persistent_string>)
+          {
+            throw std::runtime_error{ fmt::format("not ordered: {}", object_type_str(o->type)) };
+          }
           native_integer i{};
           for(auto it(typed_o->fresh_seq()); it != nullptr; it = next_in_place(it), ++i)
           {
@@ -830,6 +834,10 @@ namespace jank::runtime
         }
         else if constexpr(behavior::seqable<T>)
         {
+          if constexpr(!behavior::sequential<T> && !std::is_same_v<T, obj::persistent_string>)
+          {
+            throw std::runtime_error{ fmt::format("not ordered: {}", object_type_str(o->type)) };
+          }
           native_integer i{};
           for(auto it(typed_o->fresh_seq()); it != nullptr; it = next_in_place(it), ++i)
           {
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -788,12 +788,8 @@ namespace jank::runtime
         {
           return typed_o->nth(idx);
         }
-        else if constexpr(behavior::seqable<T>)
+        else if constexpr(behavior::seqable<T> && behavior::sequential<T>)
         {
-          if constexpr(!behavior::sequential<T> && !std::is_same_v<T, obj::persistent_string>)
-          {
-            throw std::runtime_error{ fmt::format("not ordered: {}", object_type_str(o->type)) };
-          }
           native_integer i{};
           for(auto it(typed_o->fresh_seq()); it != nullptr; it = next_in_place(it), ++i)
           {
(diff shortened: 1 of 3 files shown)


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -790,6 +790,10 @@ namespace jank::runtime
         }
         else if constexpr(behavior::seqable<T>)
         {
+          if constexpr(!behavior::sequential<T> && !std::is_same_v<T, obj::persistent_string>)
```
frenchy64, line 793: Since `persistent_string` implements `associatively_readable` I'm going to go out on a limb and say it should also implement `indexable` (even though it doesn't in Clojure/script).

I think then we can make the condition here `(behavior::seqable<T> && behavior::sequential<T>)` and then remove the extra throws.

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -790,6 +790,10 @@ namespace jank::runtime
...
+          {
+            throw std::runtime_error{ fmt::format("not ordered: {}", object_type_str(o->type)) };
+          }
           native_integer i{};
```
frenchy64, line 797: I wonder if this should be `size_t`?

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
jeaye: Seems reasonable to me.

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
jeaye: This is a `native_integer` because `index` is. Comparing a `size_t` to an `native_integer` would lead to a compiler error. Since the incoming index is a `native_integer`, it's impossible for it to represent beyond the range of the type, so the `size_t` doesn't buy us anything aside from some consistency, at the cost of requiring us to cast `index` to a local `size_t` after verifying it's `>= 0`.

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
bpiel: Yeah, as you noted, I took the approach that felt most similar to Clojure, but it does make a lot of sense that `persistent_string` would have `nth` implemented. I'll go that way.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
```diff
@@ -99,6 +99,16 @@ namespace jank::runtime::obj
     throw std::runtime_error{ fmt::format("get_entry not supported on string") };
   }
 
+  object_ptr persistent_string::nth(object_ptr const index) const
+  {
+    return get(index);
```
frenchy64, line 104: This needs to throw if out-of-bounds.

frenchy64 (changes requested):

compiler+runtime/src/cpp/jank/runtime/obj/persistent_string.cpp
jeaye: Note for Bill: `get` doesn't throw, for strings. It just returns `nil`. But `nth` does.

```
user=> (nth "foo" 5)
Execution error (StringIndexOutOfBoundsException) at jdk.internal.util.Preconditions$1/apply (Preconditions.java:55).
Index 5 out of bounds for length 3
user=> (get "foo" 5)
nil
```

We can just check the result of `get` and throw if it's `nil`. Let's format a message which helpfully provides the index and also the string's size. We do this already in `substring`.

frenchy64:
Looks like the Clojure equivalent requires a Number for the default `nth` method.

```clojure
Clojure 1.12.0
user=> (nth "a" :a nil)
Execution error (ClassCastException) at user/eval1 (REPL:1).
class clojure.lang.Keyword cannot be cast to class java.lang.Number (clojure.lang.Keyword is in unnamed module of loader 'app'; java.lang.Number is in module java.base of loader 'bootstrap')
user=> (get "a" :a nil)
nil
```

jank has this check for vectors, so I think it makes sense to add it here too.

```
clojure.core=> (nth [] :a nil)
Exception: not a number: :a
```

jeaye:
Nice work, Bill!

Issue #245, opened 2025-02-06
`vswap!` is uncallable


frenchy64:
I'm guessing the varargs is being interpreted as fixed.
```
clojure.core=> (let [v (volatile! 1)] (vswap! v inc))
Exception: invalid call to clojure.core/vswap! with 2 args provided
clojure.core=> (let [v (volatile! 1)] (vswap! v + 1))
Exception: invalid sequence: 1
clojure.core=> (let [v (volatile! 1)] (vswap! v + 1 2))
Exception: invalid call to clojure.core/vswap! with 4 args provided
```

PR #292, opened 2025-03-03
fix: make `vswap!` callable

bpiel:
Closes https://github.com/jank-lang/jank/issues/245

```
clojure.core=> (def v (volatile! 1))
#'clojure.core/v

clojure.core=> (vswap! v inc)
2

clojure.core=> (vswap! v + 1)
3

clojure.core=> (vswap! v + 1 2)
6
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core.cpp
@@ -556,6 +556,12 @@ namespace jank::runtime
     return o->type == object_type::volatile_;
   }
 
+  object_ptr vswap(object_ptr const v, object_ptr const fn)
+  {
+    auto const v_obj(expect_object<obj::volatile_>(v));
+    return v_obj->reset(dynamic_call(fn, v_obj->deref()));
+  }
+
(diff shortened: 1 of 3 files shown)


compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -556,6 +556,12 @@ namespace jank::runtime
     return o->type == object_type::volatile_;
   }
 
+  object_ptr vswap(object_ptr const v, object_ptr const fn)
+  {
+    auto const v_obj(expect_object<obj::volatile_>(v));
```
jeaye, line 561: I know you're just following what's there, but this makes me realize that `vswap` (both overloads) and `vreset` need `try_object` instead of `expect_object`. Otherwise, giving it anything else will not throw. It'll just be UB.

jeaye (changes requested):
Your changes look good. :+1: Just left a note about an issue which I think makes sense to fix while we're here.

jeaye (approved):
Nice work, man!

PR #295, opened 2025-03-07
Improved error reporting (part 2)

jeaye:
* Added source meta to more parsed objects, to improve error reporting later on
* Added reparsing support (indirectly finding object sources)
* Added macro expansion stack tracing
* Combed through all analysis errors to ensure they're reachable and improve messaging
* Removed some dead module loading code
* Moved meta fns into their own core header to minimize the cost of including them
* Reordered some clojure.core fns to aid in earlier syntax quoting support
* Improved the grammar and wording of existing error messages
* Removed some old reference usages from before we were passing around ptrs
* Updated tests

image


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -149,6 +150,37 @@
 (def seqable?
   "Return true if the seq function is supported for x"
   clojure.core-native/seqable?)
+
+(def ^:private reverse*
+  (fn* reverse* [coll]
+    (loop* [acc '()
+            coll coll]
```
frenchy64, line 157: IMO should be initialized to `(seq coll)` so we're not reasoning about a mutable collection.

compiler+runtime/test/jank/metadata/pass-meta.jank
```diff
@@ -1,11 +1,9 @@
...
-(assert (= nil (meta '())))
-(assert (= nil (meta 'foo)))
-
-; TODO: Set literals
-;(assert (= nil (meta #{})))
+(assert (map? (meta {})))
+(assert (map? (meta [])))
+(assert (map? (meta '())))
+(assert (map? (meta 'foo)))
+(assert (map? (meta #{})))
```
frenchy64, line 6: Do the meta maps contain source information?

compiler+runtime/test/jank/syntax-quote/pass-namespace-resolution.jank
```diff
@@ -1,5 +1,4 @@
 (assert (= 'clojure.core/meow `clojure.core/meow `meow))
-(assert (= '(quote clojure.core/meow) ``meow))
```
frenchy64, line 2: These two test the `syntax_quoted` variable in the reader (really, `old_syntax_quoted`).

frenchy64 (approved):
Neat and tidy, a great improvement.

compiler+runtime/test/jank/metadata/pass-meta.jank
jeaye: Yep!

compiler+runtime/src/jank/clojure/core.jank
jeaye: What would that change?

compiler+runtime/test/jank/syntax-quote/pass-namespace-resolution.jank
jeaye: They're not equal anymore, since the meta is syntax quoted.

```clj
clojure.core=> ``meow
(clojure.core/with-meta (quote clojure.core/meow) (clojure.core/apply* clojure.core/hash-map (clojure.core/seq (clojure.core/concat* (clojure.core/list :jank/source) (clojure.core/list (clojure.core/apply* clojure.core/hash-map (clojure.core/seq (clojure.core/concat* (clojure.core/list :file) (clojure.core/list "/tmp/nix-shell.nBU4ho/jank-repl-XAmZno") (clojure.core/list :start) (clojure.core/list (clojure.core/apply* clojure.core/hash-map (clojure.core/seq (clojure.core/concat* (clojure.core/list :offset) (clojure.core/list 2) (clojure.core/list :line) (clojure.core/list 1) (clojure.core/list :col) (clojure.core/list 3))))) (clojure.core/list :end) (clojure.core/list (clojure.core/apply* clojure.core/hash-map (clojure.core/seq (clojure.core/concat* (clojure.core/list :offset) (clojure.core/list 6) (clojure.core/list :line) (clojure.core/list 1) (clojure.core/list :col) (clojure.core/list 7)))))))))))))
```

jeaye: I could've kept the test and dumped the mega form in there, but I wasn't convinced the test was needed. Do you think it is?

compiler+runtime/src/jank/clojure/core.jank
frenchy64: There's a couple of subtle possibilities.

Some collections like `eduction` don't cache their values at all. So testing `empty?` on an eduction doesn't necessarily mean it has a `first`. Both calls might, say, hit the network to rebuild the seq.

And there's a race condition for mutable collections. A mutable collection could change its size between the `empty?` call and the `first` call. If it changes to empty, then the output will be `(cons nil ...)`.

compiler+runtime/test/jank/syntax-quote/pass-namespace-resolution.jank
frenchy64: (I can live without it but sad to see it go)

Could you say a bit about the overhead of all this meta info? Is this stuff you'd strip out (eventually) in a production build?

compiler+runtime/src/jank/clojure/core.jank
jeaye: I think the first case is far-fetched, given what jank supports, and the second case is a user error. But I see your argument more as a common principle for iteration, rather than for this specific *private* var. So, taking the argument that way, we should be consistent in iterating over seqs to avoid these classes of problems. That's fair.

compiler+runtime/test/jank/syntax-quote/pass-namespace-resolution.jank
jeaye: Yep, there are multiple TODOs in this PR to support a flag for such a thing. I also documented the implications here: https://github.com/jank-lang/jank/pull/295/files#diff-bced15263f03bb66a379b6f9346fa47a66e15b168e39a710e7153d1c115090e2R1100-R1104

jeaye: I think it's impractical to maintain the tests here, since their source info will differ by just adding a line above the test. Also, we'd need to ensure relative pathing for the source file, but even that could break depending on how the tests are executed.

frenchy64: Yeah, definitely. Passing the right hand side to `eval` might be a way to preserve the original test. But at that point, maybe it should live in the clojure-test-suite repo.

compiler+runtime/src/jank/clojure/core.jank
frenchy64: Yes. I think of it in terms of the input type of `reverse`, which is `seqable`. All it does is return a `seq`, with no guarantees that it's cached or idempotent.

PR #298, opened 2025-03-12
macOS Intel compat

frenchy64:
This fixes the build on my macOS Intel machine. I couldn't get the automatic architecture detection to work.


compiler+runtime/CMakeLists.txt
```diff
@@ -91,6 +91,10 @@ set(
   jank_common_compiler_flags
   -std=gnu++20
   -DIMMER_HAS_LIBGC=1 -DIMMER_TAGGED_NODE=0 -DHAVE_CXX14=1
+  # After removing boost in https://github.com/jank-lang/jank/commit/beafe61ddb1ac3ee28d88650e4caed10f171ac4d
+  # macOS x86_64 requires using fmt in header-only mode.
```
jeaye, line 95: We should link to the fmt ticket which talks about this: https://github.com/fmtlib/fmt/issues/3247

jeaye:
Thanks, Ambrose. I'll update this to target x86_64 only, by EOW.

Issue #273, opened 2025-02-17
`get` corner cases differ from Clojure


frenchy64:
We should add some more cases to clojure-test-suite combining non-associative input and defaults. `get-in` may also have some subtle differences.

```clojure
;; jank
clojure.core=> (get 1 :a :default)
nil

;; clojure
user=> (get 1 :a :default)
:default
```

PR #299, opened 2025-03-13
get: return fallback for non-associative arg

bpiel:
Closes https://github.com/jank-lang/jank/issues/273

```
;; jank BEFORE
clojure.core=> (get :a :b :c)
nil

clojure.core=> (get nil nil :def)
nil

clojure.core=> (get 1 :a :def)
nil

;; jank AFTER
clojure.core=> (get :a :b :default)
:default

clojure.core=> (get nil nil :default)
:default

clojure.core=> (get 1 :a :default)
:default

;; Clojure
user=> (get :a :b :default)
:default
user=> (get nil nil :default)
:default
user=> (get 1 :a :default)
:default

```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -577,7 +577,7 @@ namespace jank::runtime
         }
         else
         {
-          return obj::nil::nil_const();
+          return fallback;
         }
       },
       m);


jeaye:
> This seems like a case where I might want to run the clojure-test-suite against the change. Can that be done? How?

jank can't run much of that suite yet, due to some missing pieces. Don't worry about it for now!

jeaye (approved):
Great find and fix, Bill!

Issue #281, opened 2025-02-19
Add support for big_decimal and big_integer objects


jeaye:
## Overview
Clojure supports [BigInt](https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/BigInt.java) and [BigDecimal](https://docs.oracle.com/en/java/javase/21/docs/api//java.base/java/math/BigDecimal.html) as numbers which can grow arbitrarily. There is also lexing support for both of these, such that the `N` suffix on an integer creates a `BigInt` and the `M` suffix creates a `BigDecimal`.

## Steps
* Determine if we're implementing these ourselves or bringing in a library
    * Explore libraries and determine best options based on run-time and compile-time performance
    * Add the lib as a submodule and hook it into cmake, if we are adding one
* Add `big_integer` and `big_decimal` objects
* Update C API with `jank_create_big_integer` and `jank_create_big_decimal` to aid in LLVM IR gen
* Add operator overloads for them (follow ratio)
* Don't worry about all of the math overloads for them for now; those aren't being used with LLVM IR gen right now anyway
* Add lexing support for `N` and `M` suffixes, with tests
* Update numeric lexing for integers to automatically use a `big_integer` if the number is too large for a `native_integer`, with tests
* Update `clojure.core/integer?` and `runtime::is_integer`
* Add `clojure.core/bigint`, `clojure.core/biginteger`, `clojure.core/bigdec`, `clojure.core/decimal?`
* Update ratio object to use `big_integer` for its parts

Jake-Moss:
Hi, I've been (excitedly) following following Janks development for a while now and just noticed this issue and thought I might be able to add something to the first two points.

> * Determine if we're implementing these ourselves or bringing in a library
> 
>    * Explore libraries and determine best options based on run-time and compile-time performance

Last year I did some work on the [python-flint](https://github.com/flintlib/python-flint) (as my honours thesis) to bring better support for FLINTs multivariate polynomials to the Python wrapper. As part of that I did a heap of benchmarking on Pythons, GMPs, and FLINTs arbitrarily sized integers, I've copied some of that work here with the relevant benchmark.

I think it may be worth while to consider Python vs GMP vs FLINT as a sort of case study for what Jank should do.

> * Determine if we're implementing these ourselves or bringing in a library

[...]

PR #303, opened 2025-03-18
WIP big integers

Jake-Moss:
Here's the code I mentioned in #281, I was able to read my SSD and get the important things off. I believe in it's current state it can parse, lex, print, and store big integers but that's about it. I'm also not experienced which CMake to gauge if I have added the GMP dependency correctly or not.

You're welcome to use this code and/or take over this PR, unfortunately I won't be able to work on this until I have my PC back up and running which might be another week or so.


jeaye:
Thanks, @Jake-Moss! These changes look like a reasonable approach. This ticket is currently assigned to @jianlingzhong, and I know he's working on it, so I'll leave it to you two to coordinate. Best of luck with the PC. :slightly_smiling_face:

jeaye:
Merged as #328.

PR #304, opened 2025-03-19
Add final compiler error reporting improvements

jeaye:
* Rework terminal reporting to be more table-based
* Add reporting test suite
* Update macro expansion notes to only note the latest expansion
* Add support for reporting errors with synthetic macros (macros which return data that has no source info)
* Added error note sorting, for more predictable logic, which sets us up to implement note bridges later on
* Add usage tracking to errors, which allow us to refer to a macro where things failed, even if the originating data is not within the macro
* Add JAR support to error reporter, so we can refer to sources installed from maven
* Add relative pathing and path truncation to reported file names
* File mapping has been moved into the module loader and abstracted into a `file_view` which works for normal files and JAR files
* Update some `clojure.core` to use more syntax quoting, which aids in error messages

image


compiler+runtime/test/bash/error-reporting/src/analyze/invalid-throw/input.jank
```diff
@@ -0,0 +1,4 @@
+(ns input)
+
+(defn doworkson [a b]
+  (throw a b 3))
```
frenchy64, line 4: GitHub thinks that a file named `output` is a binary format, so it's difficult to review. I can't even put this comment there :)

frenchy64 (approved):
Awesome! I didn't look at all the test outputs, see my comment. But the ones I did were very nice.

compiler+runtime/test/bash/error-reporting/src/analyze/invalid-throw/input.jank
frenchy64: Hmm, `output.txt` is still displays the same. Maybe it's a git thing? Couple of things to try:

https://stackoverflow.com/a/41240899
- `git rm` + `git add` or squash

https://stackoverflow.com/a/53287723
- see if git thinks it's binary `git check-attr --all -- src/my_file.txt`

jeaye: We'll see what happens after this is squashed and merged.

PR #309, opened 2025-03-24
Unify jank's exception reporting 

jeaye:
* Add custom `JANK_TRY`/`JANK_CATCH` macros to unify our exception handling
* Add specific support for macroexpansion exceptions, to treat them as both compiler and runtime errors
* Add support for showing source info for evaluated call expressions which throw
* Gracefully handle a lack of source when reparsing; this is possible to hit when calling `(eval 'some-data)`
* Bind `*source-file*` to `nil` when in `read-string` and handle that case when reading source info from meta
* Add stack frame filtering, to remove some unneeded details (needs to be expanded for other OSs)
* Strip out a lot of redundant error creation code by using a template


compiler+runtime/include/cpp/jank/util/try.hpp
```diff
@@ -0,0 +1,74 @@
...
+ * combine that with `JANK_CATCH(jank::util::print_exception)`.
+ * However, you can specify any function you want there; it just
+ * needs to have all of the correct overloads. You'll know if it
+ * doesn't. The `JANK_CATCH_THEN` is a generalization of this
+ * which also takes some arbitrary code to then after handling
+ * the exception. We can use this to early return, for example.
+ *
+ * One implication of this design is that we can provide a single
+ * function to handle all expected exception types, with full type
+ * info. For example: `JANK_CATCH([&](auto const &e){  })` */
```
frenchy64, line 47: Impressive! Thanks for explaining.

compiler+runtime/src/cpp/jank/evaluate.cpp
```diff
@@ -618,6 +628,10 @@ namespace jank::evaluate
 
   object_ptr eval(expr::throw_ptr const expr)
   {
+    /* XXX: Clojure wraps throw expressions. I _suspect_ it does this because
+     * clojure.main uses the stack trace to provide source info by stripping out
+     * Clojure frames until the first non-Clojure frame is found. If we throw
+     * from an eval, maybe that doesn't happen? For now, we support eval, however. */
```
frenchy64, line 634: Maybe also related to working around around checked exceptions?

compiler+runtime/src/cpp/main.cpp
```diff
@@ -385,34 +305,5 @@ int main(int const argc, char const **argv)
...
-    jank::error::report(e);
-    return 1;
-  }
-  catch(...)
-  {
-    jank::util::println("Unknown exception thrown");
-    jank::print_exception_stack_trace();
-    return 1;
-  }
+  JANK_CATCH_THEN(jank::util::print_exception, return 1)
```
frenchy64, line 308: WIN

frenchy64 (approved):

Issue #310, opened 2025-03-29
Can't Compile on Linux Mint 21.3 with 6.8 kernel


veqqq:
On Linux Mint 21.3 with 6.8 kernel and a fresh brew installation:

With Brew:

```
[...]

PR #315, opened 2025-04-01
Add Distrobox Build Instructions, solve (#310)

veqqq:
Hopefully temporary, but convenient and easy.

Just adding visibility to [this](https://github.com/jank-lang/jank/issues/310#issuecomment-2767133364), which simplified [this](https://github.com/jeremy-rifkin/cpptrace/issues/226#issuecomment-2745913508).


compiler+runtime/doc/install.md
```diff
@@ -14,6 +14,27 @@ git lfs install
...
+./bin/configure -GNinja -DCMAKE_BUILD_TYPE=Debug
+./bin/compile
+```
+
+Start ( `jank/compiler+runtime/`:
+
+```
+build/jank repl
+```
+N.b. it will continue to work even if you [remove](https://distrobox.it/#uninstallation) distrobox or the `jank-ubuntu` image: `distrobox stop jank-ubuntu;distrobox rm jank-ubuntu`
```
jeaye, line 36: This last bit depends *heavily* on the host system. jank's binary is dynamically linked, so a build from Ubuntu 24.10, for example, may not run properly on Arch. They may have different versions of glibc, libz, llvm, or any of the other shared lib dependencies jank has. I don't think we can reliably have this line here.

compiler+runtime/doc/install.md
```diff
@@ -14,6 +14,27 @@ git lfs install
...
+```
+git clone --recurse-submodules https://github.com/jank-lang/jank.git
+cd jank/compiler+runtime/
+curl -s https://raw.githubusercontent.com/89luca89/distrobox/main/install | sudo sh
+distrobox create jank-ubuntu --image ubuntu:24.10
+distrobox enter jank-ubuntu
+sudo apt-get install -y curl git git-lfs zip build-essential entr libssl-dev libdouble-conversion-dev pkg-config ninja-build cmake zlib1g-dev libffi-dev clang libclang-dev llvm llvm-dev libzip-dev libbz2-dev doctest-dev gcc g++ libgc-dev
+export CC=clang; export CXX=clang++
+./bin/configure -GNinja -DCMAKE_BUILD_TYPE=Debug
+./bin/compile
```
jeaye, line 28: We have a dedicated `build.md` which has build instructions, including everything in here except for the distrobox bit. This `install.md` is meant for installing jank via package manager, whereas `build.md` is meant for manually building jank from source.

I think we can add a section to `build.md` for distrobox, if that helps, but it should just contain the three distrobox lines. The rest is already covered by `build.md`.

jeaye (changes requested):
Thanks! I've left some notes, since we have most of this documented already.

compiler+runtime/doc/install.md
veqqq: Perhaps I've misunderstood how distrobox works. Isn't it dynamically linked within the main OS' filesystem (and not a container specific system)? It worked on fresh Arch and Ubuntu 24.10, (besides Mint 21.3) and even Alpine, with dependencies only installed when entered in distrobox. (Dist didn't work on Nix.) I don't know how to check where the jank binaries are linked to, but I will make less assumption in the future.

jeaye: Distrobox is just a wrapper around docker. Each distrobox image is just a docker image and each distrobox is a docker container. Distrobox handles automatically mounting volumes and devices, to make things feel seamless, but it has its own root file system. The only shared files among the two systems, in this case, would be the jank sources.

compiler+runtime/doc/install.md
veqqq: The added prominence of the easy solution would help those unfamiliar with the C++ ecosystem. Either way, since Distrobox builds, there is a clear path to package repo inclusion, so it'd only be a temporary clutch.

(Not having used llvm, cmake, ninja etc. (beyond copypasting), it took perhaps 10 hours reading through documentation over a week to wade through some errors and arrive at the broken Debian-based builds in these issues. E.g. realizing `CXX` relates to c++ and should be set, I learned from your Distrobox code!)

jeaye: Gotcha. So, I think the changes we want to come out of this are (both in `build.md`):

1. Add the `export CC=clang; export CXX=clang++` bit to the snippets here: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/doc/build.md#compiling-jank
2. Add a bit under Dependencies which suggests that people can use distrobox, with three relevant (install, create, end) lines: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/doc/build.md#dependencies

Everything else in this PR is already in that doc.

PR #316, opened 2025-04-01
[wip] leaner syntax-quote

frenchy64:


jeaye:
I'd love to get some explanation of what you're doing here, ideally in the code itself, as a helpful comment. Is there more you're planning to do on this PR?

If we're pushing this forward, it'd be great to get some numbers as to how these changes affect doing something like compiling `clojure.core`. I can help with that.

frenchy64:
The main insight here is that syntax quote always expands to code that assumes its form contains unquote-splicing, but if unquote-splicing is never used then the code can be made significantly more direct.

For example (Clojure JVM), this should really just expand to `[]`:

```clojure
'`[]
;=> (clojure.core/apply clojure.core/vector (clojure.core/seq (clojure.core/concat)))
```

However, we need the more general expansion if unquote-splicing:

```clojure
'`[~@local-binding]
;=> (clojure.core/apply clojure.core/vector (clojure.core/seq (clojure.core/concat local-binding)))
```

We don't need it for unquoting into collection literals whose children are in evaluation contexts. This includes vector literals, so this should just be `[local-binding]`:

```clojure
'`[~local-binding]
;=> (clojure.core/apply clojure.core/vector (clojure.core/seq (clojure.core/concat (clojure.core/list local-binding))))
```

But doesn't include lists, the best we can do here is `(clojure.core/list local-binding)`:

```clojure
'`(~local-binding)
;=> (clojure.core/seq (clojure.core/concat (clojure.core/list local-binding)))
```

Maybe I'll investigate this further by forking [backtick](https://github.com/brandonbloom/backtick) to prove the ideas there first before adding directly to jank.

jeaye:
This is great insight, Ambrose! Thanks for sharing. I'm very interested in it, since the impact this has on compile times is palpable.

frenchy64:
Started here https://github.com/frenchy64/backtick/pull/1/files

PR #319, opened 2025-04-04
Feature: unbox c strings ir gen

### LINKED ISSUE #301: Don't require boxing strings for reading globals
In `llvm_processor.cpp`, when we generate our call to `jank_read_string`, we box the argument into a global:

```cpp
      llvm::SmallVector<llvm::Value *, 1> const args{ gen_global(make_box(runtime::to_code_string(o))) };
```

Instead, let's just use a C string and call a new `jank_read_string_c` (using the `_c` suffix now, so we can be consistent later). This will exist in `c-api.{h,cpp}`, right next to `jank_read_string`. They'll both do effectively the same thing, which won't require any changes to `runtime::context::read_string`, since it already takes a string view.

stmonty:
Fixes #301 

I originally made a `gen_global_from_read_string_c` but after discussion this looks like a better solution.

Adds a new c api function `jank_read_string_c`.

### DIFF
--- a/compiler+runtime/include/cpp/jank/c_api.h
+++ b/compiler+runtime/include/cpp/jank/c_api.h
@@ -25,7 +25,6 @@ extern "C"
   /* TODO: Add jank_read_c_string to avoid boxing so much. */
   jank_object_ptr jank_read_string(jank_object_ptr s);
   jank_object_ptr jank_read_string_c(char const *s);
-  /* jank_object_ptr jank_read_string_c(jank_object_ptr s); */
 
   void jank_ns_set_symbol_counter(char const * const ns, uint64_t const count);
 
--- a/compiler+runtime/include/cpp/jank/c_api.h
+++ b/compiler+runtime/include/cpp/jank/c_api.h
@@ -24,6 +24,8 @@ extern "C"
   jank_object_ptr jank_eval(jank_object_ptr s);
   /* TODO: Add jank_read_c_string to avoid boxing so much. */
   jank_object_ptr jank_read_string(jank_object_ptr s);
+  jank_object_ptr jank_read_string_c(char const *s);
+  /* jank_object_ptr jank_read_string_c(jank_object_ptr s); */
 
   void jank_ns_set_symbol_counter(char const * const ns, uint64_t const count);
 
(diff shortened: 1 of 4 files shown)


stmonty:
When I made the `gen_global_from_read_string_c` which was very similar to the non-c version I didn't include this part: https://github.com/jank-lang/jank/blob/5eb45b0b6a2ddc4349d1ac7bee02a76507e040b1/compiler%2Bruntime/src/cpp/jank/codegen/llvm_processor.cpp#L1409-L1433 

I saw the TODO and that it was creating a new global recursively. Do we still need that part?

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -75,6 +75,11 @@ extern "C"
     return __rt_ctx->read_string(s_obj->data);
   }
 
+  jank_object_ptr jank_read_string_c(char const *s)
```
jeaye, line 78: Let's make the pointer here const. `char const * const s`

compiler+runtime/include/cpp/jank/c_api.h
```diff
@@ -24,6 +24,7 @@ extern "C"
   jank_object_ptr jank_eval(jank_object_ptr s);
   /* TODO: Add jank_read_c_string to avoid boxing so much. */
```
jeaye, line 25: Please remove this TODO. That's what your PR is tackling.

jeaye (changes requested):
Nice work, Monty. Two small requests. Otherwise, looks great!

jeaye:
> When I made the `gen_global_from_read_string_c` which was very similar to the non-c version I didn't include this part:
> 
> https://github.com/jank-lang/jank/blob/5eb45b0b6a2ddc4349d1ac7bee02a76507e040b1/compiler%2Bruntime/src/cpp/jank/codegen/llvm_processor.cpp#L1409-L1433
> 
> I saw the TODO and that it was creating a new global recursively. Do we still need that part?

Let's not worry about this for now. We can tackle it outside of this PR.

jeaye:
The failing job should pass if you merge in main.

jeaye (approved):

PR #320, opened 2025-04-05
feature: Unbox var interning in the IR

stmonty:
This PR creates a new c_api function `jank_var_intern_c` used for interning vars just with c strings.

```shell
[monty@nixos:~/repos/jank/compiler+runtime]$ ./build/jank compile clojure.core | wc -l
162070
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/c_api.cpp
+++ b/compiler+runtime/src/cpp/jank/c_api.cpp
@@ -95,6 +95,14 @@ extern "C"
     return erase(__rt_ctx->intern_var(ns_obj->data, name_obj->data).expect_ok());
   }
 
+  jank_object_ptr jank_var_intern_c(char const * const ns, char const * const name)
+  {
+    auto const ns_string{ native_persistent_string(ns) };
+    __rt_ctx->intern_ns(ns_string);
+    auto const name_string{ native_persistent_string(name) };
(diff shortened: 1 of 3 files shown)


compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -95,6 +95,14 @@ extern "C"
     return erase(__rt_ctx->intern_var(ns_obj->data, name_obj->data).expect_ok());
   }
 
+  jank_object_ptr jank_var_intern_c(char const * const ns, char const * const name)
+  {
+    auto const ns_string{ native_persistent_string(ns) };
+    __rt_ctx->intern_ns(ns_string);
+    auto const name_string{ native_persistent_string(name) };
```
jeaye, line 102: No need for lines 100 or 102. They're costing us two extra allocations!

jeaye (changes requested):

compiler+runtime/src/cpp/jank/c_api.cpp
stmonty: Are you saying that it will implicitly call the constructor to turn the c-string into a persistent string when we pass the parameter? If so, isn't that the same amount of allocations as it is passed by reference into `intern_ns` and `intern_var`?

jeaye:
Nice work, Monty!

PR #321, opened 2025-04-05
Fix Folly build issues related to memory allocators on MacOS

jianlingzhong:
This commit addresses several build failures encountered when compiling and linking against the selectively built Folly library, particularly concerning its integration with jemalloc and tcmalloc since these problem still happens on MacOS.

Problem:
- Linker errors due to undefined symbols like `_nallocx`, `_xallocx`, `_mallctl`, etc., referenced by Folly code.
- Compilation errors (conflicting preprocessor defines, missing headers) when attempting to disable allocator usage via flags.
- Compilation error when adding C stub file due to C++ standard flag being applied.

Solution:
- Added preprocessor flags (`-DFOLLY_HAVE_JEMALLOC=0`, `-DFOLLY_ASSUME_NO_JEMALLOC=1`, and equivalents for tcmalloc) to `jank_common_compiler_flags` to correctly signal Folly headers during compilation that these allocators are unavailable.
- Introduced `src/cpp/jank/folly_stubs.c` with minimal stub implementations for `_nallocx` and `_xallocx` to satisfy the linker.
- Created a `folly_stubs` static library for these stubs.
- Linked both `jank_exe` and `jank_test_exe` against `folly_stubs`.
- Refactored C/C++ standard setting to use CMake variables (`CMAKE_CXX_STANDARD`, `CMAKE_C_STANDARD`, etc.) instead of putting `-std=gnu++20` in common flags, fixing the C compilation error.

These changes allow the project and its tests to compile and link successfully without requiring jemalloc or tcmalloc.

### DIFF
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -27,6 +27,12 @@ project(
     # https://stackoverflow.com/questions/71740678/cmake-error-in-findterminfo-with-clang-15-on-macos
     LANGUAGES C CXX
 )
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS ON) # To enable GNU extensions, equivalent to gnu++20
+set(CMAKE_C_STANDARD 11) # Or C17, C99 - C11 is a safe default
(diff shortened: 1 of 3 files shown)


jianlingzhong:
closing this in favor of #322

PR #327, opened 2025-04-27
fix: pass module name while evaluating cpp modules

Samy-33:
Required to compile cpp backed module into object file

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/context.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/context.cpp
@@ -1,3 +1,4 @@
+#include "jtl/immutable_string.hpp"
 #include <exception>
 
 #include <llvm/ExecutionEngine/Orc/LLJIT.h>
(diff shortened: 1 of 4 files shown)


compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -212,8 +214,7 @@ namespace jank::runtime
      * moves the `llvm::Module` held in the `PartialTranslationUnit`. */
     if(truthy(compile_files_var->deref()))
     {
-      auto module_name{ runtime::to_string(current_module_var->deref()) };
```
jeaye, line 215: So, aside from moving the `set_is_loaded` and `is_loaded` to RT context, which isn't a behavior change, this is actually the logical change.

We have a var to track the current module. Why is it not sufficient? You're using a parameter instead. This is suspicious to me, since we'll have other places in the code relying on the `current_module_var` holding the current module. Why are we not setting it in the loader when load `.cpp` modules. In fact, *where are we setting it*?

These are questions we need to understand, answer, and (if we're keeping this code) ultimately document. My gut tells me there's no reason we can't use the var for this.

jeaye (changes requested):
Thanks for the PR, Saket. Aside from minor notes, I don't understand the reasoning of the primary change. If you've considered all of the questions I posed, please just explain with a comment in the code. Otherwise, please consider my questions and revise accordingly.

PR #328, opened 2025-05-09
Support big integers

jianlingzhong:
Related to #281.


compiler+runtime/src/cpp/jank/runtime/obj/big_integer.cpp
```diff
@@ -0,0 +1,344 @@
...
+    else if(radix == 16)
+    {
+      init("0x" + std::string(s));
+    }
+    else
+    {
+      /* For all other radix, we need to manually parse the number. */
+      native_big_integer temp_val = 0;
+      native_big_integer current_radix_power = 1;
+      for(auto it = s.rbegin(); it != s.rend(); ++it)
```
jeaye, line 183: Please use direct initialization for all values.

compiler+runtime/src/cpp/jank/runtime/obj/big_integer.cpp
```diff
@@ -0,0 +1,344 @@
...
+      {
+        int digit_val;
```
jeaye, line 185: Always default-initialize primitives. `int digit_val{};`

compiler+runtime/src/cpp/jank/runtime/obj/big_integer.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    if(is_negative)
+    {
+      data *= -1;
+    }
+  }
+
+  object_ptr big_integer::create(native_persistent_string_view const &s,
+                                 native_integer radix,
+                                 native_bool is_negative)
```
jeaye, line 209: These can be const.

compiler+runtime/src/cpp/jank/runtime/obj/big_integer.cpp
```diff
@@ -0,0 +1,344 @@
...
+    seed ^= hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+  }
+
+  native_hash big_integer::to_hash(native_big_integer const &data)
+  {
+    auto const &backend = data.backend();
+
+    auto const *limbs = backend.limbs();
+    auto const size = backend.size();
+    auto const sign = backend.sign();
```
jeaye, line 268: Please always use direct initialization.

compiler+runtime/src/cpp/jank/runtime/obj/big_integer.cpp
```diff
@@ -0,0 +1,344 @@
...
+  native_real big_integer::to_native_real(native_big_integer const &data)
+  {
+    try
+    {
+      return data.convert_to<native_real>();
+    }
+    catch(std::overflow_error const &)
+    {
+      return data < 0 ? -std::numeric_limits<native_real>::infinity()
+                      : std::numeric_limits<native_real>::infinity();
```
jeaye, line 330: Is this the same behavior Clojure has?

jeaye (changes requested):
Nice work, Jianling! This is a big effort to get everything supported. Overall, the structure makes sense. I had a couple of questions around the logic, and a lot of comments around style consistency.

I think the `N` suffix support is missing, but that may've been intentional.

Again, don't worry about the merge conflicts. If we can get the PR ready to merge, I'll handle it from there.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -761,12 +819,35 @@ namespace jank::read::lex
...
+                if(denominator.expect_ok().kind == token_kind::big_integer)
+                {
+                  return ok(token(
+                    token_start,
+                    pos,
+                    token_kind::ratio,
+                    { .numerator = numerator,
+                      .denominator = native_big_integer(
+                        std::get<lex::big_integer>(denominator_token.data).number_literal) }));
+                }
```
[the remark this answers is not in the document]
jianlingzhong: currently we don't need the `N` suffix to indicate a big int. if a number is large enough, it will be a big int automatically, otherwise it's always a regular int. i suppose we want the `N` suffix support to explicitly indicate a big int? @jeaye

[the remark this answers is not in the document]
jeaye: Correct! `1N` is a big integer. `1M` is a big decimal.

jianlingzhong:
OK, we can add `N` suffix support. `M` for big decimals needs to be added
later since this PR is for big int only.

compiler+runtime/src/cpp/jank/read/lex.cpp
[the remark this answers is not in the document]
jianlingzhong: OK, we can add `N` suffix support. `M` for big decimals needs to be added
later since this PR is for big int only.

compiler+runtime/src/cpp/jank/runtime/obj/big_integer.cpp
jianlingzhong: yes:

```
user=> (double 1e10000)
##Inf
user=> (double -1e10000)
##-Inf
```

jianlingzhong:
@jeaye i've removed some unit tests that are not necessary but kept many. let me know if you think any is still redundant.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -15,6 +16,10 @@ namespace jank::runtime
     {
       return t.to_real();
     }
+    else if constexpr(std::same_as<T, native_big_integer>)
+    {
+      return t.template convert_to<native_real>();
```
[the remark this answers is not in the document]
jianlingzhong: to_number is actually to_real and i realize there is already a to_real so i deleted to_number instead.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1164,7 +1164,7 @@ namespace jank::runtime
     return visit_number_like(
       [](auto const typed_l, auto const r) -> object_ptr {
         return visit_number_like(
-          [](auto const typed_r, auto const typed_l) -> object_ptr {
+          [](auto const typed_r, auto const &typed_l) -> object_ptr {
```
jeaye, line 1167: Were all of these refs on `typed_l` added because of clang-tidy errors?

jeaye:
Nice work, @jianlingzhong! I resolved all conflicts, did some clean up, re-formatted all files, and merged.

PR #331, opened 2025-05-11
feat: aot compilation into executable

Samy-33:
Requires `clojure.core` to be compiled into the project's `target` directory for now.


compiler+runtime/src/cpp/jank/aot.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+    llvm::IntrusiveRefCntPtr<clang::DiagnosticIDs> DiagID(new clang::DiagnosticIDs());
+
+    llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem> VFS = llvm::vfs::getRealFileSystem();
+
+    clang::DiagnosticsEngine Diags(DiagID, &*DiagOpts, DiagClient, /*ShouldOwnClient=*/true);
+
+
+    std::string const TargetTripleStr = llvm::sys::getDefaultTargetTriple();
+    llvm::errs() << "Using target triple: " << TargetTripleStr << "\n";
```
jeaye, line 150: Let's remove this log.

compiler+runtime/src/cpp/jank/aot.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+    /* The Driver needs the path to the executable (used for finding related tools/resources)
+     * For this example, we'll use argv[0] or a placeholder. This might need adjustment
+     * depending on how/where your program is run relative to clang resources.
+     * TODO: infer clang++ path. Will be OS dependent and will require to infer correct version's path */
+    auto const ClangExecutablePath{ "/usr/bin/clang++" };
```
jeaye, line 156: The best way to infer clang's path is to just use `clang++` and rely on the user having set up their `PATH` the way they want. Guarantee you that `clang++` on my machine is not in `/usr/bin`. For Linux, it's best to never assume that something is in `/usr/bin`.

compiler+runtime/src/cpp/jank/aot.cpp
```diff
@@ -0,0 +1,281 @@
...
+                                    "jank_aot_compilation",
+                                    VFS);
+    TheDriver.setCheckInputsExist(false);
+
+    std::vector<char const *> Args = { strdup(ClangExecutablePath) };
+
+    for(auto const &module : __rt_ctx->loaded_modules_in_order)
+    {
+      auto const &module_path{ module::module_to_path(module) };
+      util::string_builder sb;
```
jeaye, line 169: Looks like this is unused.

compiler+runtime/src/cpp/jank/aot.cpp
```diff
@@ -0,0 +1,281 @@
...
+      Args.push_back(strdup(util::format("-I{}", include_dir).c_str()));
+    }
+
+    for(auto const &library_dir : library_dirs)
+    {
+      Args.push_back(strdup(util::format("-L{}", library_dir).c_str()));
+    }
+
+    for(auto const &lib : { "-ljank",
+                            // Default libs that Jank depends on
```
jeaye, line 187: jank is always lowercase. Also, please use `/* */` comments and use correct punctuation. Comments are documentation.

compiler+runtime/src/cpp/jank/aot.cpp
```diff
@@ -0,0 +1,281 @@
...
+    Args.push_back(strdup("-o"));
+    auto const output_filepath{ relative_to_cache_dir(output_filename) };
+    Args.push_back(strdup(output_filepath.c_str()));
+
+    /* Required because of `strdup` usage and need to manually free the memory.
+     * Clang expects C strings that we own. */
+    util::scope_exit const cleanup{ [&]() {
+      for(auto const s : Args)
+      {
+        // NOLINTNEXTLINE(cppcoreguidelines-no-malloc): Needs to free these strings becuase we strdupped them
```
jeaye, line 222: The justification here is duplicated from three lines above. We don't need both. Please use `/* */` comments.

compiler+runtime/src/cpp/jank/aot.cpp
```diff
@@ -0,0 +1,281 @@
...
+      // Diagnostics engine already printed errors.
+      return;
+    }
+
+    // Execute the compilation jobs (preprocess, compile, assemble)
+    // This actually runs the commands determined by BuildCompilation.
+    int Result = 0;
+    if(C && !C->containsError())
+    {
+      llvm::SmallVector<std::pair<int, clang::driver::Command const *>, 4> FailingCommands;
```
jeaye, line 244: Why 4?

compiler+runtime/src/cpp/jank/aot.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+    if(!C || C->containsError())
+    {
+      llvm::errs() << "Failed to build compilation steps.\n";
+      // Diagnostics engine already printed errors.
+      return;
+    }
+
+    // Execute the compilation jobs (preprocess, compile, assemble)
+    // This actually runs the commands determined by BuildCompilation.
```
jeaye, line 240: Please use `/* */` comments.

compiler+runtime/src/cpp/jank/init.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <llvm/Support/CommandLine.h>
+#include <llvm/Support/ManagedStatic.h>
+#include <llvm/Support/TargetSelect.h>
+
+#include <clojure/core_native.hpp>
+#include <clojure/string_native.hpp>
+
+#include <jank/compiler_native.hpp>
+#include <jank/perf_native.hpp>
+#include <jank/init.hpp>
```
jeaye, line 11: Including this doesn't get us anything.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -282,7 +280,11 @@ namespace jank::runtime
       absolute_module = module::nest_module(ns->to_string(), module);
     }
 
-    binding_scope const preserve{ *this };
+    binding_scope const preserve{ *this,
+                                  obj::persistent_hash_map::create_unique(
+                                    std::make_pair(current_ns_var, ns),
+                                    std::make_pair(current_module_var,
+                                                   make_box(absolute_module))) };
```
jeaye, line 287: Let's explain what's going on here. Remember how this wasn't clear to you when you saw it in Clojure's code.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -305,8 +307,7 @@ namespace jank::runtime
 
     binding_scope const preserve{ *this,
                                   obj::persistent_hash_map::create_unique(
-                                    std::make_pair(compile_files_var, jank_true),
-                                    std::make_pair(current_module_var, make_box(module))) };
+                                    std::make_pair(compile_files_var, jank_true)) };
```
jeaye, line 310: What's the purpose of this change?

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -71,11 +71,20 @@ namespace jank::util::cli
...
-    /* Run subcommand. */
+    /* run-main subcommand. */
     auto &cli_run_main(*cli.add_subcommand("run-main", "Load and execute -main."));
     cli_run_main.fallthrough();
     cli_run_main.add_option("module", opts.target_module, "The entrypoint module.")->required();
 
+    /* compile-main subcommand. */
+    auto &cli_compile_main(
+      *cli.add_subcommand("compile-main",
+                          "Load and compile project with entrypoint module containing -main."));
```
jeaye, line 82: What do you think about:

1. Renaming `compile-main` to just `compile`
2. Renaming the existing `compile` to `compile-module`

jeaye (changes requested):
Very exciting! This is so cool to see coming together, Saket. I think it needs cleanup, but I'm cool with the overall structure of this as a first pass. I know we talked about getting something working and merged first and then building on it, so we're right on track for that.

The key thing we're missing is tests, which will need to go in `test/bash/aot`. We can chat this week about how that test suite works and the types of things we'll want to test for. No need to wait until Thursday, if you want to meet sooner.

compiler+runtime/src/cpp/jank/runtime/context.cpp
Samy-33: Here, I removed the `current_module_var` because `load_module` will set the `current_module_var`.

compiler+runtime/src/cpp/jank/util/cli.cpp
Samy-33: I was thinking the same. I will make that change.

compiler+runtime/include/cpp/jank/runtime/context.hpp
```diff
@@ -140,6 +140,7 @@ namespace jank::runtime
     /* TODO: This needs to be a dynamic var. */
     native_unordered_map<jtl::immutable_string, native_vector<jtl::immutable_string>>
       module_dependencies;
+    native_deque<jtl::immutable_string> loaded_modules_in_order;
```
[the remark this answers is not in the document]
Samy-33: Why does it need to be thread-local and need sync?

When we'll have full AOT support that considers multiple threads + runtime dependency ordering, this variable will go away. We will need more sophisticated handling i.e. keep a direct acyclic graph of module dependencies.

compiler+runtime/include/cpp/jank/aot.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <jank/util/cli.hpp>
```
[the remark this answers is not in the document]
jeaye: Everything that *can* be forward declared should be. Some things cannot be. If you use a type by value, you need to know its size. However, if you only use it by reference or by pointer then you don't, since the only size you'll need is the size of a pointer, which is known.

compiler+runtime/include/cpp/jank/runtime/context.hpp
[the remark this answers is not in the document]
jeaye: If we have multiple threads loading modules, and thus updating `loaded_modules_in_order`, we will have undefined behavior, since `native_deque` is not thread-safe.

compiler+runtime/src/cpp/jank/init.cpp
Samy-33: Without it the linter recommends we make the `init` function static

jeaye: Ah, silly linter. Ok, we can keep it then.

compiler+runtime/src/cpp/jank/aot.cpp
Samy-33: Just using `clang++` doesn't work. I'm not sure if this way of doing compilation looks for the `clang++` binary. Instead, it expects the executable as the first arg.

The error I got was: `clang++: command not found`

Samy-33: We can accept a CXX variable from command line for clang's executable path?

Although, ideally we should infer the path ourselves. We can try figuring it out by looking into known system installation directories and clang's version to make sure that we get the correct clang++ executable. But this won't always yield results. You talked about AppImage that will include the clang's executable. Before that, I think there's no deterministic way to get clang executable path.

Samy-33:
Pending tasks:
* Tests
* clang executable path inference
* changing compile-main to compile, and compile to compile-module
* Converting `loaded_modules_in_order` into a dynamic var

compiler+runtime/src/cpp/jank/aot.cpp
jeaye: > Just using `clang++` doesn't work. I'm not sure if this way of doing compilation looks for the `clang++` binary. Instead, it expects the executable as the first arg.
> 
> The error I got was: `clang++: command not found`

Is `clang++` on your `PATH`? i.e. can you type it into a shell? Or do you only put it into `CXX`? We can look at `CXX`, but we can't rely on it, since an AOT compiled jank program shipped to some non-devs machine will not likely have `CXX` set. We need them to have clang installed, though, and we need to be able to find it. Relying on `PATH` is the most portable way, but we may also need more beyond that, such as looking into common places if we don't find it in `CXX` or `PATH`.

Samy-33:
No, I'm not using CXX yet. Just using `clang++` doesn't work.

I found `llvm::sys::Program::FindProgramByName("clang++")`. I'll see if
this works.

jeaye:
> No, I'm not using CXX yet. Just using `clang++` doesn't work. I found `llvm::sys::Program::FindProgramByName("clang++")`. I'll see if this works.

Sounds like Clang isn't searching for the binary. It expects an absolute path. Using `FindProgramByName` sounds like a good way to do that. If not, we can use `getenv_s` on `PATH` to look for it manually. Much better to rely on LLVM to do this for us.

compiler+runtime/include/cpp/jank/aot/processor.hpp
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include <jtl/immutable_string.hpp>
+#include <jtl/result.hpp>
+
+/* Forward declared */
```
jeaye: There is no need for this. It's 100% a *what* comment.

compiler+runtime/include/cpp/jank/aot/processor.hpp
```diff
@@ -0,0 +1,40 @@
...
+namespace jank::util::cli
+{
+  struct options;
+}
+
+namespace jank::aot
+{
+  struct compiler_err
```
jeaye: Most importantly, this type should not exist. We have an existing error facility in `error.hpp`.

As a side note, no need to shorten the name of public types. If we were to have this type, we would spell our `error` in the name.

compiler+runtime/include/cpp/jank/aot/processor.hpp
```diff
@@ -0,0 +1,40 @@
...
+  {
+    compiler_err(int const &return_code, jtl::immutable_string const &err_message)
+      : return_code{ return_code }
+      , err_message{ err_message }
+    {
+    }
```
jeaye: Also a side note, since we're not keeping this struct, but this is 4 lines of code that does nothing. Without this code, you can still construct the type by specifying two values and the compiler will use aggregate initialization. Do not write ctors unless you need them!

Also side note, be mindful not to take `int const &`, since `int` is small enough to fit in a register. If you take a reference to it, you disallow that optimization.

compiler+runtime/include/cpp/jank/aot/processor.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+    int return_code;
```
jeaye: Also a side note, since we're not keeping this struct, but *every primitive type needs to be default initialized.* No exceptions. i.e. this should be `int return_code{};`

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+extern "C" jank_object_ref jank_load_jank_compiler_native();
+extern "C" jank_object_ref jank_load_clojure_core();
+extern "C" jank_object_ref jank_var_intern_c(char const *, char const *);
+extern "C" jank_object_ref jank_deref(jank_object_ref);
+extern "C" jank_object_ref jank_call2(jank_object_ref, jank_object_ref, jank_object_ref);
+extern "C" jank_object_ref jank_parse_command_line_args(int, char const **);
+)");
+
+    /* TODO: Is there a better way to do this then copying? */
+    auto const ordered_modules{ __rt_ctx->loaded_modules_in_order.copy() };
```
jeaye: Use a read lock. i.e. `__rt_ctx->loaded_modules_in_order.rlock()`

Multiple readers can read simultaneously.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+    std::string main_file_path{ tmp_dir / "jank-main-XXXXXX" };
+
+    auto const fd{ mkstemp(main_file_path.data()) };
+    close(fd);
+
+    std::ofstream out(main_file_path);
+    out << sb.release();
+
+    out.flush();
+    out.close();
```
jeaye: In truth, neither of these lines are needed. The dtor will do everything.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+      return compiler_err{ 1, "clang++ executable not found. Ensure it exists on the path!" };
+    }
+    clang::driver::Driver driver(clang_inferred_path.get(),
+                                 target_triple,
+                                 diags,
+                                 "jank_aot_compilation",
+                                 vfs);
+    driver.setCheckInputsExist(false);
+
+    std::vector<char const *> Args = { strdup(clang_inferred_path.get().c_str()) };
```
jeaye: Please use snake_case and direct initialization.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -246,48 +264,14 @@ int main(int const argc, char const **argv)
...
     if(parse_result.is_err())
     {
       return parse_result.expect_err();
     }
     auto const &opts(parse_result.expect_ok());
 
-    if(opts.gc_incremental)
-    {
-      GC_enable_incremental();
-    }
```
jeaye: Looks like this was removed here, but not added anywhere else. Why?

jeaye (changes requested):

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+    };
+    llvm::IntrusiveRefCntPtr<clang::DiagnosticIDs> const diag_id{ new clang::DiagnosticIDs() };
+    clang::DiagnosticsEngine diags{ diag_id, &*diag_opts, diag_client, /*ShouldOwnClient=*/true };
+
+    llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem> const vfs{ llvm::vfs::getRealFileSystem() };
+
+    std::string const target_triple{ llvm::sys::getDefaultTargetTriple() };
+
+    /* The Driver needs the path to the executable (used for finding related tools/resources)
+     * For this example, we'll use argv[0] or a placeholder. This might need adjustment
```
jeaye: This comment doesn't seem accurate. We're not using `argv[0]` or a placeholder.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+      for(auto const s : Args)
+      {
+        /* NOLINTNEXTLINE(cppcoreguidelines-no-malloc) */
+        free(reinterpret_cast<void *>(const_cast<char *>(s)));
+      }
+    } };
+
+    llvm::ArrayRef<char const *> const Argv(Args);
+
+    std::unique_ptr<clang::driver::Compilation> C(driver.BuildCompilation(Argv));
```
jeaye: Clang/LLVM commonly uses single letter, uppercase variable names. Please use something snake_case and more descriptive in jank code. Also, we can just use `auto` here to clean things up. For example:

```cpp
auto const compilation_result{ driver.BuildCompilation(argv) };
```

Because of the `auto`, we still end up saving space even with a longer name.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+    if(!clang_inferred_path)
+    {
+      return compiler_err{ 1, "clang++ executable not found. Ensure it exists on the path!" };
+    }
+    clang::driver::Driver driver(clang_inferred_path.get(),
+                                 target_triple,
+                                 diags,
+                                 "jank_aot_compilation",
+                                 vfs);
+    driver.setCheckInputsExist(false);
```
jeaye: What's the reasoning for this being false?

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+    {
+      llvm::SmallVector<std::pair<int, clang::driver::Command const *>> FailingCommands;
+      Result = driver.ExecuteCompilation(*C, FailingCommands);
+
+      for(auto const &P : FailingCommands)
+      {
+        /* Check if command signaled an error */
+        if(P.first < 0)
+        {
+          llvm::errs() << "Error executing command: " << P.second->getExecutable() << "\n";
```
jeaye: Let's use `util::println` for printing in jank. You can pass `stderr` to print there.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+          /* TODO: We can print more details about the failing command P.second */
+        }
+      }
+    }
+    else
+    {
+      Result = 1;
+    }
```
jeaye: We can initialize this int to 1 and get rid of 4 lines of code. Also, please use snake_case.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,305 @@
...
+    if(Result == 0)
+    {
+      llvm::outs() << util::format("Compilation successful. Find the executable at: `{}`",
+                                   output_filepath)
+                        .c_str();
+    }
+    else
+    {
+      return compiler_err{ Result, "Compilation command execution failed.\n" };
+    }
```
jeaye: Rather than having an if for positive, an else for negative, and then a positive return, let's minimize the branching. Check for your early returns. Everything else doesn't get nested.

So, if we check that the result != 0, we can early return with the error. Otherwise, we can print before we return `ok()` (the `jtl::` should not be needed).

compiler+runtime/src/cpp/main.cpp
Samy-33: Instead of taking the value from `opts`, we now get the value out of a define macro. This was done when using a common `jank_init` fn, which should not rely on the CLI.

Let me know if you have better ideas.

jeaye: This is a flag for the *user* to set when invoking jank. This needs to remain a flag.

I don't understand why this can't still be in `opts`.

Samy-33:
When do we use -DVARS vs cli args?

If we want to use CLI args, this will be passed again as a parameter to
`jank_init`. It already has 4 args as of now. Increasing number of args
didn't feel like a good idea for an API fn.

WDYT?

jeaye:
> When do we use -DVARS vs cli args?

When you install jank on your machine, you need to be able to customize, with each invocation, if you want incremental GC or not. Defines happen at compile time. If we use a define, you can't choose, since the jank binary you install will only have the define it was compiled with.

You wouldn't use a `-DJANK_MODULE_PATH`, right? The module path is specific to the user's computer, their project, etc. It's not something we can know when building the jank compiler. Same with the type of GC the user wants to use.

> If we want to use CLI args, this will be passed again as a parameter to `jank_init`. It already has 4 args as of now. Increasing number of args didn't feel like a good idea for an API fn. WDYT?

I don't get it, dude. We have `opts` literally right above the code that was removed. Just put the code back. :joy: 

image

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,298 @@
...
+    auto compilation(driver.BuildCompilation(Argv));
+
+    if(!compilation || compilation->containsError())
+    {
+      return error::aot_compilation_failure();
+    }
+
+    /* Execute the compilation jobs (preprocess, compile, assemble)
+     * This actually runs the commands determined by BuildCompilation. */
+    int exit_code = 1;
```
jeaye: Please always use direct initialization.

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -939,4 +945,63 @@ extern "C"
   {
     profile::report(label);
   }
+
+  int jank_init(int const argc,
+                char const **argv,
+                jank_bool init_default_ctx,
```
jeaye: Both `argv` and `initi_default_ctx` can be `const`.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -238,6 +235,28 @@ namespace jank
...
+
+  static void compile(util::cli::options const &opts)
+  {
+    using namespace jank;
+    using namespace jank::runtime;
+
+    if(opts.target_ns != "clojure.core")
+    {
+      __rt_ctx->compile_module("clojure.core").expect_ok();
+    }
```
jeaye: `target_ns` will never be set, right?

compiler+runtime/include/cpp/jank/error.hpp
```diff
@@ -78,6 +78,9 @@ namespace jank::error
     analyze_unresolved_var,
     analyze_unresolved_symbol,
     analyze_macro_expansion_exception,
+    aot_compilation_failure,
+    aot_clang_executable_not_found,
+    aot_module_not_found,
```
jeaye: 1. These were added inside the `analyze_` block here
2. Then they were added right after the `analyze_` block in the `switch` below
3. Then they were added in a new section in the `switch` in `error.cpp`

The correct place is a new section. On top of that, the usage of `internal_failure` here is not as intended. Note, in this enum, that every section has its own internal failure type. i.e. `internal_lex_failure`, `internal_parse_failure`, `internal_analyze_failure`, `internal_codegen_failure`. But for the AOT errors you're instead using the generic fallback `internal_failure` value. This is not what we want. Instead, we should have a new section for these AOT errors. I'd be inclined to put it after the codegen section (i.e. `internal_codegen_failure`), since each of these sections is also laid out in the order in which things actually happen. Here, AOT errors are put before codegen, which is not how it would play out. We'd want the following:

* aot_compilation_failure
* aot_clang_executable_not_found
* aot_module_not_found
* internal_aot_failure

Ther internal failure function belongs with the rest of the AOT error functions, in `error/aot.hpp`.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -0,0 +1,298 @@
...
+        /* NOLINTNEXTLINE(cppcoreguidelines-no-malloc) */
+        free(reinterpret_cast<void *>(const_cast<char *>(s)));
+      }
+    } };
+
+    // prints the command used for compilation
+    // for(auto const st : compiler_args)
+    // {
+    //   util::print("{} ", st);
+    // }
```
jeaye: `util::println` supports vectors of printable things. i.e. you can just `util::println("args: {}", compiler_args);`

Side note that you don't have a newline anywhere here, so the next (unrelated) log line will not be separated.

If we keep this log (as a single `println`), we should remove the comment on line 252.

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -939,4 +945,63 @@ extern "C"
...
+      llvm::llvm_shutdown_obj const Y{};
+
+      llvm::InitializeNativeTarget();
+      llvm::InitializeNativeTargetAsmParser();
+      llvm::InitializeNativeTargetAsmPrinter();
+
+      if(init_default_ctx)
+      {
+        runtime::__rt_ctx = new(GC) runtime::context{};
+      }
```
jeaye, line 986: What's the use case for this code path? From what I can tell, it's only for tests. Why not have the tests directly construct a context?

I imagine you've heard the various sayings about introducing a bool into an API. Let's be very mindful about this one.

If you're also imagining it would be used by people embedding jank, I'm not so sure, since they'd likely want to be able to configure optimizations levels and such as well. Better to not assume right now.

compiler+runtime/test/bash/ahead-of-time/expected_outputs/jank-and-cpp-modules/core
```diff
@@ -0,0 +1,6 @@
+Loaded 'my-lib
```
jeaye: Directory names are always singular. i.e. `test`, `src`, `include`, `target`. Also, we use kebab-case whenever it's not on the module path. So `compiler+runtime/test/bash/ahead-of-time/expected_outputs/jank-and-cpp-modules/core` should instead be something like `compiler+runtime/test/bash/ahead-of-time/expected-output/jank-and-cpp/core`.

compiler+runtime/test/bash/ahead-of-time/pass-test
```diff
@@ -0,0 +1,159 @@
...
+  (:require
+   [babashka.fs :as fs]
+   [babashka.process :as proc]
+   [clojure.string :as str]
+   [clojure.test :as t :refer [deftest is testing use-fixtures]]))
+
+(def this-nsym (ns-name *ns*))
+
+(def jank-path
+  (-> (proc/sh "which jank") :out str/trim))
```
jeaye, line 15: If `jank` is on the path, which is necessary for `which jank` to find it, we can trust that we can invoke it by just using `jank`. For the same reason that you didn't need to do `which clojure` to invoke `clojure` below.

compiler+runtime/test/bash/ahead-of-time/pass-test
```diff
@@ -0,0 +1,159 @@
...
+  (let [libs-relative-paths ["/" "/third-party/libzippp"
+                             "/third-party/ftxui"
+                             "/third-party/cpptrace"
+                             "/_deps/libdwarf-build/src/lib/libdwarf"]]
+    (->> libs-relative-paths
+         (map (fn [lib] (str "-L" jank-build-dir lib)))
+         (str/join " "))))
+
+(def jank_exe
+  (str jank-path " -DIMMER_HAS_LIBGC " lib-paths))
```
jeaye: Very odd that this is using snake_case.

compiler+runtime/test/bash/ahead-of-time/pass-test
```diff
@@ -0,0 +1,159 @@
...
+
+(defn- module-path [alias]
+  (-> (proc/sh "clojure" (str "-A:" alias) "-Spath") :out str/trim))
+
+(defn- find-cli []
+  (-> (proc/sh "find ./target -name cli") :out str/trim))
+
+(defn- get-headers []
```
jeaye: There is no need to make these private.

compiler+runtime/src/cpp/main.cpp
Samy-33: Thanks for pointing this out.

Related to this, we have for different commands used `target_module` and `target_ns`. We should streamline their usages.

**compile-module**: We have used `target_ns`
**repl**: `target_module`
**cpp-repl**: `target_module`
**compile**: `target_module`
**run-main**: `target_module`

I think we should have
**compile-module**: `target_module`
**repl**: `target_ns`
**cpp-repl**: `target_module`
**compile**: `target_ns`
**run-main**: `target_ns`

what do you think?

compiler+runtime/src/cpp/jank/c_api.cpp
Samy-33: We are using `jank_init` in the `entrypoint` cpp file generated for AOTd project and in jank's main entrypoint.

This codepath was chosen because of the difference in how AOT entrypoint needs to create the context (using default constructor) and how jank's main entrypoint creates it (constructor accepting cli options).)

If we don't want this path here, we'd want to initialize the context in the `entrypoint` cpp file similar to how we're doing in the jank's main.

I wanted `entrypoint` file to just use the jank's c_api. So chose to have this path.

We can have a c_api function to create context using default constructor. It's such a small function, should it even be their in the api?

What do you think?

compiler+runtime/test/bash/ahead-of-time/pass-test
Samy-33: Need the `jank-path` to infer the `build` directory for compile command. We need the library directories to be included in the command.

This will be removed when your changes that incorporate the library directories in cmake file, I believe.

Added a todo to remove it once the branches are merged

compiler+runtime/test/bash/ahead-of-time/pass-test
Samy-33: Is it that we don't need privates in the tests or it's being inconsistently used in this test file?

compiler+runtime/src/cpp/jank/aot/processor.cpp
Samy-33: Removed. Copy paste miss

compiler+runtime/src/cpp/jank/c_api.cpp
Samy-33: Accept a struct instead of a bool, for now. Later think of a better design.

jeaye:
Great work, Saket! This is so nice to have merged.

PR #332, opened 2025-05-11
fix renaming jank_tests to jank_test

markx:
Fix building jank-test binary. 

The CMakeLists.txt file was broken in 222d0f6


jeaye:
Looks like this is also outdated, which is why CI is failing here: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/bin/jank/compiler%2Bruntime/build%2Btest.clj#L28

jeaye:
Nice! That's all sorted now. Looks like main continued to build due to CMake caching. It'd be good to have our CI script delete the CMake cache before configuring!

Thanks for the PR. :slightly_smiling_face:

PR #333, opened 2025-06-01
docs: update clang, llvm version and adds libdwarf

Samy-33:
By default ubuntu install v18 for clang and llvm
And can't build without libdwarf

### DIFF
--- a/compiler+runtime/doc/build.md
+++ b/compiler+runtime/doc/build.md
@@ -6,13 +6,13 @@ scripts to compile your own. See the section on compiling Clang/LLVM below.
 For Debian-based distros, this should be all you need:
 
 ```bash
-sudo apt-get install -y curl git git-lfs zip build-essential entr libssl-dev libdouble-conversion-dev pkg-config ninja-build cmake zlib1g-dev libffi-dev clang libclang-dev llvm llvm-dev libzip-dev libbz2-dev doctest-dev gcc g++ libgc-dev
+sudo apt-get install -y curl git git-lfs zip build-essential entr libssl-dev libdouble-conversion-dev pkg-config ninja-build cmake zlib1g-dev libffi-dev clang-19 libclang-19-dev llvm-19 llvm-19-dev libzip-dev libbz2-dev doctest-dev gcc g++ libgc-dev libdwarf-dev
 ```
 
 For Arch:
 
 ```bash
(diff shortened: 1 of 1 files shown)


jeaye:
Thanks, Saket. Which version of Ubuntu did you try this with? If you're using Ubuntu 24.04, since that's what our CI is using, it's ooooold. 24.10 and 25.04 have both been released already. That's why our CI bb scripts manually install Clang/LLVM, outside of apt.

So, I think we want to be careful about locking things in version 19. jank should work fine with version 20, based on my testing, and if a distro moves on to not have 19 in their repos, we don't want our install commands to not work. I think keeping them unversioned is the best move, combined with the fact the CMake will verify the Clang version when it builds jank and jank will verify the Clang version (next Q) when it tries to use Clang for AOT compilation. This grants us the freedom of dealing with supported ranges, rather than just one version at a time.

However, adding libdwarf makes total sense, if we need that.

What do you think?

Samy-33:
Yes, I am trying to replicate CI on Ubuntu 24.04. Didn't realise CI builds
clang from source.

The intension behind locking clang-19 on debian based distros to even
support old distro versions that may not yet be on clang-19+. Do distros
remove previous versions quite frequently from their repos? Specially
debian?

Another issue is clang and llvm known to introduce breaking changes in
newer versions. Jank might work with clang-20 but we can't say the same
about clang-21 or higher versions. So locking to clang19 makes sense to me.

I agree if distros remove older versions from their repos often, it's a
problem. We might just want to rely on the unversioned clang.

jeaye:
> Yes, I am trying to replicate CI on Ubuntu 24.04. Didn't realise CI builds clang from source.

They don't build from source. They install precompiled clang/llvm builds using the official install script.

> Jank might work with clang-20 but we can't say the same about clang-21 or higher versions. So locking to clang19 makes sense to me. 

This is why jank's build system will check which clang version you have installed. It's up to us to support whatever range we can, ideally based on what's present from Debian unstable to Homebre to the latest Arch. We cannot lock into one version, since Debian, Homebrew, and Arch are *not* always going to be on the same version. We need to support a range.

Case in point, as of today:

* Debian stable: Clang 16
* Debian unstable: Clang 19
* Arch: Clang 19 [^1]
* Homebrew: Clang 20


[^1]: It's 19, but was flagged out of date 3 months ago and Clang 20 is available in the staging repos.

RyanP1978 (approved):

RyanP1978 (approved):

RyanP1978 (approved):

Issue #209, opened 2025-01-30
Set and map literals should check for duplicate keys


frenchy64:
```clojure
;;jank
clojure.core=> #{1 1}
#{1}
clojure.core=> {1 1 1 1}
{1 1}

;;clj
user=> #{1 1}
Syntax error reading source at (REPL:1:7).
Duplicate key: 1
user=> {1 1 1 1}
Syntax error reading source at (REPL:2:10).
Duplicate key: 1
```

jeaye:
The TODO for this is here: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/analyze/processor.cpp#L1245

shantanu-sardesai:
I'm interested in working on this!

shantanu-sardesai:
> The TODO for this is here: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/analyze/processor.cpp#L1245

Updated [permalink](https://github.com/jank-lang/jank/blob/4239ee52e60e290df6656cf52e4a3d00609a6eb5/compiler%2Bruntime/src/cpp/jank/analyze/processor.cpp#L1245).

PR #334, opened 2025-06-07
GH ISSUE 209 | Verifying duplicate keys

shantanu-sardesai:
Fixes for #209.

**Note to reviewer**: Due to the LLVM@19 issues I was not able to run the entire test suite.

Preview for duplicate map literal errors:

image

Preview for duplicate set literal errors:

image (1)

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -375,12 +375,12 @@ namespace jank::read::parse
       {
         return err(it.latest.unwrap().expect_err());
       }
-      auto const key(it.latest.unwrap().expect_ok());
+      auto const key(it.latest.unwrap().expect_ok().unwrap());
 
       if(++it == end())
       {
         return error::parse_odd_entries_in_map({ start_token.start, latest_token.end },
-                                               { key.unwrap().start.start, key.unwrap().end.end });
+                                               { key.start.start, key.end.end });
       }
 
       if(it.latest.unwrap().is_err())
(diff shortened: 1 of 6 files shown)


compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -367,6 +366,7 @@ namespace jank::read::parse
       .expect_ok();
     util::scope_exit const finally{ [] { __rt_ctx->pop_thread_bindings().expect_ok(); } };
 
+    native_unordered_map<runtime::object_ref, jank::read::parse::object_source_info> parsed_keys{};
```
jeaye: You're already in `jank::read::parse`, so there's no need for that here.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -389,6 +389,19 @@ namespace jank::read::parse
       }
       auto const value(it.latest.unwrap().expect_ok());
 
+      if(auto parsed_key = parsed_keys.find(key.unwrap().ptr); parsed_key != parsed_keys.end())
```
jeaye: `parsed_key` can be `const`. The general rule is, if it can be `const`, it should be.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -589,14 +602,31 @@ namespace jank::read::parse
       .expect_ok();
     util::scope_exit const finally{ [] { __rt_ctx->pop_thread_bindings().expect_ok(); } };
 
+    native_unordered_map<runtime::object_ref, jank::read::parse::object_source_info> parsed_items{};
```
jeaye: Same as above, there's extraneous namespacing here.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -589,14 +602,31 @@ namespace jank::read::parse
...
     {
       if(it.latest.unwrap().is_err())
       {
         return err(it.latest.unwrap().expect_err());
       }
-      ret.insert(it.latest.unwrap().expect_ok().unwrap().ptr);
+
+      auto const item(it.latest.unwrap().expect_ok());
+
+      if(auto parsed_item = parsed_items.find(item.unwrap().ptr); parsed_item != parsed_items.end())
```
jeaye: Same as above, this can be `const`.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -589,14 +602,31 @@ namespace jank::read::parse
...
+            item.unwrap().start.start,
+            item.unwrap().end.end
+        },
+          { "Original item.",
+            { parsed_item->second.start.start, parsed_item->second.end.end },
+            error::note::kind::info });
+      }
+
+      parsed_items.insert({ item.unwrap().ptr, item.unwrap() });
+      ret.insert(item.unwrap().ptr);
```
jeaye: We do `item.unwrap()` SIX times in the span of 15 lines. I think we should put it into its own `unwrapped_item` local instead. Same with the map keys.

Unwrapping isn't free, but this is also just about minimizing duplication.

shantanu-sardesai: Makes sense, I've reduced it in the sets parse logic and made a separate commit for the `parse_map` function in case we don't wish to do the same there, should be easier to revert in that case 😅.

RyanP1978 (approved):

RyanP1978 (approved):

RyanP1978 (approved):

RyanP1978 (approved):

RyanP1978 (approved):

PR #335, opened 2025-06-13
Support promotion to `persistent_hash_map`

shantanu-sardesai:


compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -365,56 +359,100 @@ namespace jank::read::parse
...
-      {
-        return error::parse_odd_entries_in_map({ start_token.start, latest_token.end },
-                                               { key.start.start, key.end.end });
-      }
+    expected_closer = prev_expected_closer;
 
-      if(it.latest.unwrap().is_err())
+    native_unordered_map<runtime::object_ref, object_source_info> parsed_keys{};
+
+    auto const build_map([&](auto &ret) -> processor::object_result {
```
jeaye: An `object_result` has two cases:

* Success case: `jtl::option<object_source_info>`
* Error case: `error_ref`

For `build_map`, we don't care about the success case. So, let's just return a `jtl::result<void, error_ref>` instead. On success, we return `ok()`. Right now, it's possible to return an `object_source_info` from `build_map` and it'll never be used! That's just asking for bugs to happen.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -365,56 +359,100 @@ namespace jank::read::parse
...
 
-      if(auto const parsed_key = parsed_keys.find(key.ptr); parsed_key != parsed_keys.end())
+      return jtl::ok(none);
+    });
+
+    /* TODO: use transients to build up maps. */
+    if(items.size() <= runtime::detail::native_persistent_array_map::max_size)
+    {
+      runtime::detail::native_persistent_array_map ret;
+      auto res = build_map(ret);
```
jeaye: Please always use direct initialization. Same with below.

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -821,6 +821,14 @@ namespace jank::read::parse
         CHECK(r1.is_err());
       }
 
+      SUBCASE("Promoted maps")
+      {
+        lex::processor lp{ "{:k1 1 :k2 2 :k3 3 :k4 4 :k5 5 :k6 6 :k7 7 :k8 8 :k9 9 :k10 10}" };
+        processor p{ lp.begin(), lp.end() };
+        auto const r1(p.next());
+        CHECK(r1.is_ok());
```
jeaye, line 834: Let's also check that the size of the parsed hash map is 10.

compiler+runtime/src/cpp/jank/read/parse.cpp
jeaye: `auto &ret` isn't actually being returned, so perhaps `auto &map` is a better name.

jeaye (changes requested):
Functionally, looks good! Just a few cleanup and testing comments. :muscle:

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -421,7 +421,7 @@ namespace jank::read::parse
     });
 
     /* TODO: use transients to build up maps. */
-    if(items.size() <= runtime::detail::native_persistent_array_map::max_size)
+    if((items.size() / 2) <= runtime::detail::native_persistent_array_map::max_size)
```
shantanu-sardesai, line 424: Should I bother using `std::ceil(item.size() / 2.0)` here?

Since it could so happen that we have 17 items, currently I'll end up using the `native_persistent_array_map` instead of `persistent_hash_map`. Since this will anyway be an "Odd elements" error, I'm assuming we don't need to care about this?

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -821,6 +821,40 @@ namespace jank::read::parse
...
+
+        CHECK(r2.is_ok());
+
+        auto const t2(r2.expect_ok().unwrap().ptr);
+
+        CHECK(t2.data->type == object_type::persistent_hash_map);
+        CHECK(r2.expect_ok().unwrap().start
+              == lex::token{ 0, 1, lex::token_kind::open_curly_bracket });
+        CHECK(r2.expect_ok().unwrap().end
+              == lex::token{ hash_map_source.size() - 1, 1, lex::token_kind::close_curly_bracket });
```
shantanu-sardesai: The hack that I've implemented for checking for the size. The two _proper_ approaches I can think off would be to:
1. `expect_object<obj::persistent_hash_map>(t2)->data.size()` but the type constraint of `object_like` is not being satisfied by the `obj::persistent_hash_map`. Interestingly it is callable with the same type parameter from other parts of the project just not in this case here.
2. I manually construct an equivalent map using `make_box<obj::persistent_hash_map>` and just `equal` it with `t2`.

compiler+runtime/src/cpp/jank/read/parse.cpp
jeaye: I don't think we need extra logic for that here.

compiler+runtime/test/cpp/jank/read/parse.cpp
jeaye: Option 1 is correct. I explained in Slack what you're missing.

PR #337, opened 2025-06-18
Improve `native_array_map`

shantanu-sardesai:
- Rename the incorrectly named `native_persistent_array_map` to `native_array_map`.
- Add capability to preallocate the backing array.
- Use `reserve` in `parse_map`.

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -424,7 +424,7 @@ namespace jank::read::parse
     {
       runtime::detail::native_array_map map{};
       map.reserve(items.size() / 2);
-      auto res = build_map(map);
+      auto const res{ build_map(map) };
 
       if(res.is_err())
       {
@@ -441,7 +441,7 @@ namespace jank::read::parse
     {
       /* TODO: use transients to build up maps. */
       runtime::detail::native_persistent_hash_map map{};
-      auto res = build_map(map);
+      auto const res{ build_map(map) };
 
       if(res.is_err())
       {
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -407,7 +407,7 @@ namespace jank::read::parse
 
         parsed_keys.insert({ key.ptr, key });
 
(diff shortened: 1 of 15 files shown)


compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -420,10 +420,10 @@ namespace jank::read::parse
...
     });
 
-    /* TODO: use transients to build up maps. */
-    if((items.size() / 2) <= runtime::detail::native_persistent_array_map::max_size)
+    if((items.size() / 2) <= runtime::detail::native_array_map::max_size)
     {
-      runtime::detail::native_persistent_array_map map{};
+      runtime::detail::native_array_map map{};
+      map.reserve(items.size() / 2);
       auto res = build_map(map);
```
jeaye: Please always use direct initialization. Also, `res` can be `const`.

compiler+runtime/src/cpp/jank/runtime/detail/native_array_map.cpp
```diff
@@ -1,15 +1,23 @@
...
+#include <jank/runtime/detail/native_array_map.hpp>
 #include <jank/runtime/core/equal.hpp>
 #include <jank/util/fmt.hpp>
 
 namespace jank::runtime::detail
 {
   /* TODO: Int sequence to clean this up? */
-  static object_ref *make_next_array(object_ref const * const prev,
+  static object_ref *make_next_array(object_ref * const prev,
+                                     usize const reserved_length,
```
jeaye: We called this `cap` in the type, so we should mirror the same name here.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -439,6 +439,7 @@ namespace jank::read::parse
     }
     else
     {
+      /* TODO: use transients to build up maps. */
       runtime::detail::native_persistent_hash_map map{};
       auto res = build_map(map);
```
jeaye: Please always use direct initialization. Also, `res` can be `const`. (I pointed these out in the last PR, but they were missed)

jeaye (changes requested):

jeaye (approved):
Nice work!

PR #338, opened 2025-06-25
Add support for `transient_array_map`

shantanu-sardesai:


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -219,6 +219,10 @@ namespace jank::runtime
     {
       return expect_object<obj::transient_hash_set>(coll)->disjoin_in_place(o);
     }
+    if(coll->type == object_type::transient_array_map)
```
jeaye: This should be an `else if`, like what comes after it.

PR #339, opened 2025-06-27
Remove redundant `persistent_list_sequence`

shantanu-sardesai:
Since the `persistent_list_sequence` and `persistent_list` data structures are basically the same we can just get rid of one of them.


compiler+runtime/include/cpp/jank/read/reparse.hpp
```diff
@@ -6,7 +6,7 @@
 namespace jank::runtime::obj
 {
   using persistent_list_ref = oref<struct persistent_list>;
-  using persistent_list_sequence_ref = oref<struct persistent_list_sequence>;
+  using persistent_list_sequence_ref = oref<struct persistent_list>;
```
jeaye, line 9: We should get rid of any references to `persistent_list_sequence_ref`. No point forward declaring the same type with two different aliases here.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_list.cpp
```diff
@@ -137,7 +137,7 @@ namespace jank::runtime::obj
     {
       return {};
     }
-    return make_box<persistent_list_sequence>(this, ++data.begin(), data.end(), data.size() - 1);
+    return make_box<persistent_list>(data.rest());
   }
 
   persistent_list_sequence_ref persistent_list::next_in_place() const
```
jeaye: The comment in here no longer applies. We should be able to fill this in with a more efficient implementation.

jeaye (changes requested):

compiler+runtime/include/cpp/jank/read/reparse.hpp
shantanu-sardesai: Indeed, I was just thinking of keeping it around in case we ever go back on this decision... but I guess we can just bring this back I guess.

Issue #330, opened 2025-05-11
divide by 0 is wrong


jianlingzhong:
on current main branch:

```clojure
clojure.core=> (/ 1 0)
0

clojure.core=> (/ 0 1)
0

clojure.core=> (/ 1 0.0)
##Inf

```

jeaye:
In C++, division by an integer zero is undefined behavior. Division by floating point zero is well defined. We could conditionally have this check in each of our 12+ `div` functions, but then we're adding a branch to every division.

Clojure JVM throws an exception on division by zero, but I think it makes sense to follow jank's host here, which is C++. So, we should leave it up to the user to handle division safely and instead put our focus on doing division quickly.

Decided here: #342.

PR #342, opened 2025-06-30
Fix division by zero to throw exception like Clojure JVM

luizhespanha:
- Add division by zero check in div(object_ref, object_ref) function
- Throw 'Divide by zero' exception for integer division by zero
- Behavior now matches Clojure JVM: (/ 1 0) throws exception
- Float division by zero still returns infinity: (/ 1 0.0) = ##Inf
- Add required <limits> header for the implementation

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
@@ -1,4 +1,5 @@
 #include <random>
+#include <limits>
 
 #include <jank/runtime/core/math.hpp>
 #include <jank/runtime/behavior/number_like.hpp>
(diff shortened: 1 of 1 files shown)


luizhespanha:
https://github.com/jank-lang/jank/issues/330

jeaye:
Hi Luiz! Thanks for the PR. I've reviewed the code and I've spent some time thinking about this.

In C++, division by an integer zero is undefined behavior. Division by floating point zero is well defined. We could conditionally have this check in each of our 12+ `div` functions, but then we're adding a branch to every division.

Clojure JVM throws an exception on division by zero, but I think it makes sense to follow jank's host here, which is C++. So, we should leave it up to the user to handle division safely and instead put our focus on doing division quickly. I will close the ticket accordingly.

PR #345, opened 2025-07-10
Nativetry2

jianlingzhong:


compiler+runtime/include/cpp/jank/codegen/llvm_processor.hpp
```diff
@@ -177,5 +177,13 @@ namespace jank::codegen
     native_unordered_map<obj::symbol_ref, llvm::Value *> locals;
     /* TODO: Use gc allocator to avoid leaks. */
     std::list<deferred_init> deferred_inits{};
+
+    struct lpad_and_catch_bb
+    {
+      llvm::BasicBlock *lpad_bb;
+      llvm::BasicBlock *catch_bb;
```
jeaye: These pointers will have undefined values by default. Please make sure they're default initialized with `{}`.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -362,7 +367,22 @@ namespace jank::codegen
...
+
+    llvm::Value *call{};
+
+    if(lpad_and_catch_body_stack.empty())
+    {
+      call = ctx->builder->CreateCall(fn, arg_handles);
+    }
+    else
+    {
+      auto const normal_dest = llvm::BasicBlock::Create(*ctx->llvm_ctx, "invoke.normal", this->fn);
```
jeaye: 1. Please use direct initialization.
2. What do you think about using `call_fn_name` with `"invoke.{}.normal"` to improve IR readability?

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +961,251 @@ namespace jank::codegen
...
+      };
+      ctx->builder->CreateInvoke(fn,
+                                 unreachable_dest,
+                                 lpad_and_catch_body_stack.back().lpad_bb,
+                                 { value });
+      ctx->builder->SetInsertPoint(unreachable_dest);
+    }
+    else
+    {
+      /* If not in a try block, a simple call is sufficient. The program will terminate. */
```
jeaye: The program may not terminate. We just don't have a catch for the exception in IR. For example, when running in the terminal REPL client, exceptions are just printed and we continue to prompt.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +961,251 @@ namespace jank::codegen
...
-                                                    finally.unwrap_or(gen_global(jank_nil)) };
-    auto const call(ctx->builder->CreateCall(fn, args));
+    auto const current_fn(ctx->builder->GetInsertBlock()->getParent());
 
-    if(expr->position == expression_position::tail)
+    /* A function must have a personality function set to use landing pads. */
+    auto personality_fn_type
+      = llvm::FunctionType::get(ctx->builder->getInt32Ty(), /*isVarArg=*/true);
+    auto personality_fn
+      = ctx->module->getOrInsertFunction("__gxx_personality_v0", personality_fn_type);
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +961,251 @@ namespace jank::codegen
...
+    }
 
-    llvm::SmallVector<llvm::Value *, 3> const args{ body,
-                                                    catch_.unwrap_or(gen_global(jank_nil)),
-                                                    finally.unwrap_or(gen_global(jank_nil)) };
-    auto const call(ctx->builder->CreateCall(fn, args));
+    auto const current_fn(ctx->builder->GetInsertBlock()->getParent());
 
-    if(expr->position == expression_position::tail)
+    /* A function must have a personality function set to use landing pads. */
```
jeaye: It would be lovely to explain more about personality functions here. If not, perhaps link to the best docs for them.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +961,251 @@ namespace jank::codegen
...
+
+    /* --- Block Setup --- */
+    auto const original_insert_bb{ ctx->builder->GetInsertBlock() };
+    auto const finally_bb{ llvm::BasicBlock::Create(*ctx->llvm_ctx, "finally") };
+    auto const cont_bb{ llvm::BasicBlock::Create(*ctx->llvm_ctx, "try.cont") };
+    auto const lpad_bb{ llvm::BasicBlock::Create(*ctx->llvm_ctx, "lpad", current_fn) };
+
+    llvm::BasicBlock *catch_body_bb{};
+
+    bool const has_catch = expr->catch_body.is_some();
```
jeaye: Please use direct initialization. Also, we can use `auto` here.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +961,251 @@ namespace jank::codegen
...
+    if(has_catch)
     {
-      return ctx->builder->CreateRet(call);
+      catch_body_bb = llvm::BasicBlock::Create(*ctx->llvm_ctx, "catch.body");
     }
-    return call;
+
+    lpad_and_catch_body_stack.emplace_back(lpad_and_catch_bb(lpad_bb, catch_body_bb));
```
jeaye: `emplace_back` takes the arguments for construction, so we should just pass `lpad_bb, catch_body_bb`, rather than creating a second `lpad_and_catch_bb` and using the move ctor.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +961,251 @@ namespace jank::codegen
...
+    util::scope_exit const pop_landing_pad{ [this]() { lpad_and_catch_body_stack.pop_back(); } };
+
+    /* --- Try Block --- */
+    ctx->builder->SetInsertPoint(original_insert_bb);
+    auto const original_try_pos = expr->body->position;
+    expr->body->propagate_position(expression_position::value);
+    auto const try_val = gen(expr->body, arity);
```
jeaye: Please use direct initialization. Same on line 1042.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +961,251 @@ namespace jank::codegen
...
+    if(ctx->builder->GetInsertBlock()->getTerminator() == nullptr)
+    {
+      ctx->builder->CreateBr(cont_bb);
+    }
+
+    /* We pushed the landing pad for our `try` block. It has now been popped, before
+     * generating catch/finally. We must push it back on, so that the scope_exit
+     * guard at the top can correctly pop it later, restoring the stack for the rest
+     * of this function's codegen. */
+    lpad_and_catch_body_stack.emplace_back(lpad_and_catch_bb(lpad_bb, catch_body_bb));
```
jeaye: Same comment from above about `emplace_back` usage.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -14,11 +13,14 @@
 #include <cpptrace/gdb_jit.hpp>
 
 #include <jank/util/process_location.hpp>
-#include <jank/util/make_array.hpp>
 #include <jank/util/dir.hpp>
 #include <jank/util/fmt.hpp>
 #include <jank/jit/processor.hpp>
 #include <jank/profile/time.hpp>
+#include <llvm/ExecutionEngine/Orc/ExecutionUtils.h>
+#include <llvm/ExecutionEngine/Orc/Mangling.h>
```
jeaye: Please put LLVM includes up with the LLVM section after line 11/12.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -14,11 +13,14 @@
...
+#include <jank/runtime/object.hpp>
+#include <jank/runtime/oref.hpp>
```
jeaye: There is no need to include `oref.hpp`, since it's part of `object.hpp`.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -138,9 +140,33 @@ namespace jank::jit
     compiler_instance->LoadRequestedPlugins();
 
     interpreter = llvm::cantFail(clang::Interpreter::create(std::move(compiler_instance)));
-
-    auto const &load_result{ load_dynamic_libs(opts.libs) };
-    if(load_result.is_err())
+    /* Add a search generator that can resolve symbols from the main executable itself.
+     * This is essential for the JIT to find runtime functions and RTTI data
+     * (like typeinfo for exceptions) that live in the host process. */
```
jeaye: I'm unsure about this. The JIT is already able to find runtime functions, so I suspect it already has what we're adding. Are you sure that we don't just need to define the symbols?

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -138,9 +140,33 @@ namespace jank::jit
...
+    auto &lljit = interpreter->getExecutionEngine().get();
+    auto &jd = lljit.getMainJITDylib();
+    auto const &dl = lljit.getDataLayout();
```
jeaye: This whole code block uses indirect initialization, as well as some LLVM style PascalCase naming (i.e. `Mangle`).

jeaye (changes requested):
Excellent work, Jianling. This was a huge effort.

I've done a first pass of review, but I'd like to understand the meat of the IR gen for try some more. We can discuss it in our upcoming call. From there, I would like to have us expand the comments for the try IR gen to give an idea of the overall structure before hand.

Also, I'll want to review the tests we have for `try` to ensure that we have good coverage. Please do the same.

compiler+runtime/src/cpp/jank/jit/processor.cpp
jianlingzhong: we need this. if i revert changes in this file, i got errors:

```
❯ bin/test
[4/5] Generating classes/core-libraries
FAILED: classes/core-libraries /Users/jianlingzhong/personal/jank/compiler+runtime/build/classes/core-libraries 
cd /Users/jianlingzhong/personal/jank/compiler+runtime/build && /Users/jianlingzhong/personal/jank/compiler+runtime/build/jank compile-module clojure.core && touch /Users/jianlingzhong/personal/jank/compiler+runtime/build/classes/core-libraries
JIT session error: Symbols not found: [ _N4jank7runtime4orefINS0_6objectEEE ]
Uncaught exception: Failed for find symbol: 'jank_load_clojure_core'

#0  in <jank jit frame -- not yet supported> at libc++abi.dylib
#1  in <jank jit frame -- not yet supported> at libunwind.dylib
#2  in <jank jit frame -- not yet supported> at libc++abi.dylib
#3  in jtl::result<jank::runtime::object*  at jank
#4  in jtl::result<jank::runtime::object*  at jank
#5  in jank::runtime::module::loader::load_o at jank
#6  in jank::runtime::module::loader::load at jank
#7  in jank::runtime::context::load_module at jank
#8  in jank::runtime::context::compile_module at jank
#9  in jank::compile_module at main.cpp:107
#10 in main::$_0::operator at main.cpp:295
#11 in main::$_0::__invoke at main.cpp:266
#12 in _jank_init at jank
#13 in main at main.cpp:266
#14 in start + 6075 at dyld
```

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -192,21 +191,27 @@ namespace jank::codegen
...
       {
-        ctx->builder->CreateRet(gen_global(jank_nil));
+        continue; /* Start the next iteration of the outer loop */
       }
-    }
 
-    if(target == compilation_target::eval)
-    {
-      //to_string();
+      if(!ctx->builder->GetInsertBlock()->getTerminator())
```
jeaye: Do we need to check this again, or is the `block_terminated` check above sufficient to unconditionally do this?

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
+    util::scope_exit const pop_landing_pad{ [this]() { lpad_and_catch_body_stack.pop_back(); } };
+
+    /* --- Try Block --- */
+    ctx->builder->SetInsertPoint(original_insert_bb);
+    auto const original_try_pos{ expr->body->position };
+    expr->body->propagate_position(expression_position::value);
+    auto const try_val{ gen(expr->body, arity) };
+    expr->body->propagate_position(original_try_pos);
+
+    llvm::BasicBlock *try_end_bb = nullptr;
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
-      return ctx->builder->CreateRet(call);
+      catch_body_bb = llvm::BasicBlock::Create(*ctx->llvm_ctx, "catch.body");
     }
-    return call;
+
+    lpad_and_catch_body_stack.emplace_back(lpad_bb, catch_body_bb);
+    util::scope_exit const pop_landing_pad{ [this]() { lpad_and_catch_body_stack.pop_back(); } };
+
+    /* --- Try Block --- */
+    ctx->builder->SetInsertPoint(original_insert_bb);
```
jeaye: Do we need to set this here? I don't see it being changed anywhere above, unless creating a new BB does that.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
+    auto const original_try_pos{ expr->body->position };
+    expr->body->propagate_position(expression_position::value);
```
jeaye, line 1760: I think it's worth commenting that we put the try into value position so that no returns are created because we need to ensure we make it to the finally.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
+        llvm::PointerType::get(ctx->builder->getInt8Ty(), 0)) };
+      landing_pad->addClause(object_ptr_type_info);
+    }
+
+    llvm::Value *catch_val{};
+    llvm::BasicBlock *catch_end_bb{};
+
+    if(has_catch)
+    {
+      auto const cleanup_lpad_bb
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
+      landing_pad->setCleanup(true);
+    }
+
+    if(has_catch)
+    {
+      auto const object_ptr_type_info{ ctx->module->getOrInsertGlobal(
+        typeid(object_ref).name(),
+        llvm::PointerType::get(ctx->builder->getInt8Ty(), 0)) };
+      landing_pad->addClause(object_ptr_type_info);
+    }
```
jeaye: Seems like this could be merged into the `has_catch` below.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
+      cleanup_lpad->setCleanup(true);
+      if(expr->finally_body.is_some())
+      {
+        gen(expr->finally_body.unwrap(), arity);
+      }
+      ctx->builder->CreateResume(cleanup_lpad);
+    }
+    else
+    {
+      if(expr->finally_body.is_some())
```
jeaye: We know we have a finally body, since the check above handles the case of no catch and no finally. This else means we have no catch, so thus we must have a finally.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
+      {
+        gen(expr->finally_body.unwrap(), arity);
+      }
+
+      if(!lpad_and_catch_body_stack.empty())
+      {
+        /* rethrow the exception to the outer catch if one is available. */
```
jeaye: Let's explain more of what's going on here and *then* lead up to what we're going to do about it (rethrow).

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
+    if(ctx->builder->GetInsertBlock()->getTerminator() == nullptr)
+    {
+      ctx->builder->CreateBr(cont_bb);
+    }
+
+    /* We pushed the landing pad for our `try` block. It has now been popped, before
+     * generating catch/finally. We must push it back on, so that the scope_exit
+     * guard at the top can correctly pop it later, restoring the stack for the rest
+     * of this function's codegen. */
+    lpad_and_catch_body_stack.emplace_back(lpad_bb, catch_body_bb);
```
jeaye, line 2068: This is a great comment.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -926,54 +964,267 @@ namespace jank::codegen
...
+        ctx->builder->CreateRet(final_val);
+      }
+      else
+      {
+        ctx->builder->CreateUnreachable();
+      }
+      return nullptr;
+    }
+
+    return final_val ? final_val : gen_global(jank_nil);
```
jeaye: You can use this: `return final_val ?: gen_global(jank_nil);`

jeaye:
The other note that I have is that I don't fully understand why `nullptr` is needed to convey that a terminator has been created. Outside of a `try`, we can rely 100% on a terminator being created simply by using tail position.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
jianlingzhong: removed the check.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
jianlingzhong: added more comments to explain it.

jeaye:
Nice. The main comment I have remaining is just for the returning of nullptr to indicate terminators. Can we rely on checking if the block has a terminator instead and never return nullptr?

jianlingzhong:
> Great work, Jianling. I've given this a thorough read throughout the past few days. Overall, I think this is well done. Most of the code is straightforward to follow, until we get into the weeds of `try` generation. The spot where it becomes most dense is `if(has_catch)`, where we deal with rethrowing and such. I've left some comments along the way.
> 
> Aside from my comments, I think that the biggest chunk of code which looks like it could be cleaned up is the `if(has_catch)` and the corresponding `else`. Is there any way we could merge those two branches, or do they need to stay as they are? If we can't flatten this and deduplicate it, then we should try to provide a clearer narrative of what we're doing with comments.

I've made some major changes to the code so the IR gen related to the catch and finally bodies are much easier to follow. Unfortunately that does mean you'll have to review the code logic again (but hopefully much easier this time). I've also added extensive and verbose comments to explain the code. Let me know if this reads better.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1767,7 +1767,7 @@ namespace jank::codegen
      *    - If there's no 'finally' block, it branches directly to the continuation
      *      block 'cont_bb', as the try-catch-finally construct is complete.
      */
-    if(try_val && ctx->builder->GetInsertBlock()->getTerminator() == nullptr)
```
jeaye: This is not what I was talking about when we were discussing returning `nullptr`. This change doesn't do anything; it's just different syntax for the same logic. The problem is that this PR is now *returning* `nullptr` from `gen` functions. We don't want to do this.

Example: https://github.com/jank-lang/jank/pull/345/files#diff-5007b8edf88a4766c304b097dd6f86a792d35bcef1cce60693d0be86c26e866dR1515-R1522

jeaye:
@jianlingzhong We'll need to make sure CI is passing before this can be merged. CI is green on main, so this branch is introducing some failures we need to investigate.

jeaye:
Great job getting this huge feature done, Jianling!

PR #347, opened 2025-07-12
feat: introduce jank lib with static runtime

Samy-33:
- Also allows user to choose type of runtime while ahead of compiling their projects

### DIFF
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -153,8 +153,7 @@ endfunction()
 # remains a static lib, since these symbols need to be accessible in the
 # compiler's runtime by the JIT compiler.
 
-add_library(
-  jank_lib STATIC
+set(JANK_LIBRARY_COMMON_SOURCES
   src/cpp/jtl/panic.cpp
   src/cpp/jtl/assert.cpp
   src/cpp/jank/c_api.cpp
@@ -273,16 +272,24 @@ add_library(
   src/cpp/jank/analyze/expr/case.cpp
   src/cpp/jank/analyze/local_frame.cpp
   src/cpp/jank/analyze/step/force_boxed.cpp
+
+  # Native module sources.
+  src/cpp/clojure/core_native.cpp
+  src/cpp/clojure/string_native.cpp
+  src/cpp/jank/perf_native.cpp
+)
+
+add_library(
+  jank_lib STATIC
+  ${JANK_LIBRARY_COMMON_SOURCES}
+
   src/cpp/jank/evaluate.cpp
   src/cpp/jank/codegen/llvm_processor.cpp
   src/cpp/jank/jit/processor.cpp
   src/cpp/jank/aot/processor.cpp
 
   # Native module sources.
-  src/cpp/clojure/core_native.cpp
-  src/cpp/clojure/string_native.cpp
   src/cpp/jank/compiler_native.cpp
-  src/cpp/jank/perf_native.cpp
 )
 
 set_property(TARGET jank_lib PROPERTY OUTPUT_NAME jank)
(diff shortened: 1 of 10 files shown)


compiler+runtime/CMakeLists.txt
```diff
@@ -153,8 +153,7 @@ endfunction()
 # remains a static lib, since these symbols need to be accessible in the
 # compiler's runtime by the JIT compiler.
 
-add_library(
-  jank_lib STATIC
+set(JANK_LIBRARY_COMMON_SOURCES
```
jeaye, line 156: No need to uppercase all of this. All of our CMake variables are lowercase.

compiler+runtime/CMakeLists.txt
```diff
@@ -273,16 +272,24 @@ add_library(
...
+
   src/cpp/jank/evaluate.cpp
   src/cpp/jank/codegen/llvm_processor.cpp
   src/cpp/jank/jit/processor.cpp
   src/cpp/jank/aot/processor.cpp
 
   # Native module sources.
-  src/cpp/clojure/core_native.cpp
-  src/cpp/clojure/string_native.cpp
   src/cpp/jank/compiler_native.cpp
```
jeaye, line 292: Why keep this one in the static sources? It requires codegen for the only function in there.

compiler+runtime/CMakeLists.txt
```diff
@@ -363,6 +370,75 @@ target_link_options(jank_lib PRIVATE ${jank_linker_flags})
...
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/folly>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/bpptree/include>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/immer>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/magic_enum/include/magic_enum>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/cli11/include>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/ftxui/include>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/libzippp/src>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/cpptrace/include>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/boost-preprocessor/include>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third-party/boost-multiprecision/include>"
```
jeaye, line 406: We need to not duplicate all of this between the two libs. Same goes for the linked libs and the common compile options.

compiler+runtime/include/cpp/jank/runtime/context.hpp
```diff
@@ -136,7 +136,9 @@ namespace jank::runtime
      * of previous code. This is essential for REPL use. */
     /* TODO: This needs to be synchronized. */
     analyze::processor an_prc{ *this };
+#ifndef JANK_STATIC_RUNTIME
```
jeaye, line 139: Let's def `JANK_DYNAMIC_RUNTIME` for dynamic builds and `JANK_STATIC_RUNTIME` for static builds. Checking for "not static" is both indirect and potentially too broad. For example, if we add a third runtime, this check would also match that.

compiler+runtime/include/cpp/jank/aot/processor.hpp
```diff
@@ -22,6 +22,7 @@ namespace jank::aot
     native_vector<jtl::immutable_string> define_macros;
     native_vector<jtl::immutable_string> libs;
 
+    jtl::immutable_string runtime;
```
jeaye, line 25: We want an enum for this, not a string. It only has two possible values, whereas the domain of a string is infinite.

compiler+runtime/src/cpp/jank/evaluate_static_rt.cpp
```diff
@@ -0,0 +1,135 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+
+  analyze::expr::function_ref wrap_expression(
+    analyze::expression_ref const expr,
+    jtl::immutable_string const &name,
+    native_vector<runtime::obj::symbol_ref>
+      params) // NOLINT(performance-unnecessary-value-param): This is a dummy implementation for static runtime
+  {
+    throw std::runtime_error{ "Eval disabled in static runtime." };
```
jeaye, line 14: Let's define a `constexpr` C string at the top of the file with this message, instead of duplicating it 24 times.

compiler+runtime/src/cpp/jank/evaluate_static_rt.cpp
```diff
@@ -0,0 +1,135 @@
...
+namespace jank::evaluate
+{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+
+  analyze::expr::function_ref wrap_expression(
+    analyze::expression_ref const expr,
+    jtl::immutable_string const &name,
+    native_vector<runtime::obj::symbol_ref>
+      params) // NOLINT(performance-unnecessary-value-param): This is a dummy implementation for static runtime
```
jeaye, line 12: Please put the comment before the arg, use `NOLINTNEXTLINE`, use `/* */` style comments, and end the comment in a `.` since comments are documentation.

compiler+runtime/src/cpp/jank/evaluate_static_rt.cpp
```diff
@@ -0,0 +1,135 @@
+#include <jank/evaluate.hpp>
+
+namespace jank::evaluate
+{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
```
jeaye, line 6: This is not needed if you just remove the param names.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -659,11 +663,18 @@ namespace jank::runtime::module
       locked_ordered_modules->push_back(module);
     }
     return ok();
+#else
+    return err("Load or eval not allowed in static runtime");
```
jeaye, line 667: Every error must be a complete sentence or paragraph, so please end this with a `.`.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -695,11 +706,18 @@ namespace jank::runtime::module
     load();
 
     return ok();
+#else
+    return err("Can't load file in static runtime");
```
jeaye, line 710: Please use correct grammar and punctuation for error messages.

> Dynamic file loading is disable in jank's static runtime.

Let's use this for `load_o`, `load_jank`, and `load_cpp`. Put it into a `constexpr` global so we're not repeating it. Making sure we write helpful error messages is important in every part of the system.

jeaye (changes requested):
Nice! This came together well. I have some cleanup requests, since there are a few different forms of duplication throughout, but I think the overall approach will work well.

Assuming you've tried this with your AOT compilation, will you please provide the `ldd` (linux) or `otool -L` (mac) output for the resulting static runtime AOT binary in the PR description?

jeaye:
Closing, since this is very old and no doubt needs some reworking before we can merge. If you want to reopen this when we have a working static runtime, we can do that. Or make a new PR. Up to you!

Issue #324, opened 2025-04-16
Support Direct Linking


stmonty:
### Direct Linking

We want to support direct linking Jank modules for more efficient function invocation. In Jank, function's are called through their vars, but with direct linking this indirection will be removed for the module that was compiled with the direct linking optimization.

The removal of this indirection will increase the efficiency of the compiled module. Compiling Jank's `clojure.core` will be a good example of a module that can gain speed improvements through direct linking.

#### Drawback

The main drawback of direct linking is that it limits the dynamic nature of Jank as redefining a var won't automatically change all the call-sites to use the new var. This can be mitigated by marking a var with `^:dynamic` or `^:deref` but this use case won't be supported for now.

### Implementation Plan
The plan is to add a new compilation flag `--direct-linking` that when set would change the IR generation.

The main IR functions that are currently used are `jank_deref` for derefing the var to get the function ptr and then performing `jank_call` with the function pointer and arguments.

Example:

```
  %17 = call ptr @jank_deref(ptr %16)
  %18 = load ptr, ptr @string_510969203, align 8
  %19 = load ptr, ptr @var_clojure.core_SLASH_foo, align 8
  %20 = call ptr @jank_deref(ptr %19)
  %21 = call ptr @jank_call0(ptr %20)
  %22 = call ptr @jank_call2(ptr %17, ptr %18, ptr %21)
```

The idea would be to simplify this by not requiring the loading of the `var` and the need to dereference it. 
At compile time, all the call-sites for a function will be replaced with the static call to the function. The `var` will still exist in the global namespace map.

### Unknowns

How much does this rely on AOT compilation working to fully test with `clojure.core`?
Are there new `c_api` functions that are needed to support this use case?

stmonty:
# Table of Contents

1.  [Use Cases](#orgf71ae58)
    1.  [Single Module (eval)](#org3537f57)
        1.  [Where to put `var_root` global?](#org8866268)
    2.  [Single Module (AoT)](#org48e498e)
        1.  [Where to put the `var_root` global?](#org2071b0a)
    3.  [Evaling a Module that has AoT dependency](#org96bff4a)
        1.  [Where to put the `var_root` global?](#org08ab306)
    4.  [AoT compiling a module that has an AoT dependency](#orga5436b2)
        1.  [Where to put the `var_root` global?](#org86e3830)
        2.  [This is the simple case](#orge960a6c)
[...]

PR #376, opened 2025-07-30
feature: Direct Linking Part 1

stmonty:
A long time coming. Part of the work on #324 
I still need to add some comments about the different cases and what they represent. I also don't really like the name for the enum members that I chose.

### DIFF
--- a/compiler+runtime/include/cpp/jank/codegen/llvm_processor.hpp
+++ b/compiler+runtime/include/cpp/jank/codegen/llvm_processor.hpp
@@ -122,12 +122,7 @@ namespace jank::codegen
      * 3. load_init: Initialized and derefed in the "jank_load" IR function.
      *    This like 2. applies to ahead-of-time compiled modules.
      */
-    enum class var_root_kind : u8
-    {
-      global_init,
-      binded_def,
-      load_init
-    };
+    enum class var_root_kind : u8;
 
     llvm_processor() = delete;
     llvm_processor(analyze::expr::function_ref const expr,
--- a/compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
+++ b/compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
@@ -68,6 +68,13 @@ namespace jank::codegen
     pb.crossRegisterProxies(*lam, *fam, *cgam, *mam);
   }
 
+  enum class llvm_processor::var_root_kind : u8
+  {
+    global_init,
+    binded_def,
+    load_init
+  };
+
   llvm_processor::llvm_processor(expr::function_ref const expr,
                                  jtl::immutable_string const &module_name,
                                  compilation_target const target)


compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -298,14 +304,26 @@ namespace jank::codegen
...
-    auto const call(ctx->builder->CreateCall(fn, args));
+    llvm::Value *call{};
+    auto const var_qualified_name(make_box<obj::symbol>(expr->var->n, expr->var->name));
+    if(__rt_ctx->opts.direct_linking && !(expr->var->dynamic))
+    {
+      if(root_fn->name.starts_with("jank_load"))
+      {
+        call = gen_var_root(var_qualified_name, var_root_kind::raw_global_init);
+      }
+      call = gen_var_root(var_qualified_name);
```
jeaye: I'm guessing this needs to be in an `else`. Otherwise we're losing the value in `call` from line 313.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -298,14 +304,26 @@ namespace jank::codegen
...
-      llvm::FunctionType::get(ctx->builder->getPtrTy(), { ctx->builder->getPtrTy() }, false));
-    auto const fn(ctx->module->getOrInsertFunction("jank_deref", fn_type));
-
-    llvm::SmallVector<llvm::Value *, 1> const args{ ref };
-    auto const call(ctx->builder->CreateCall(fn, args));
+    llvm::Value *call{};
+    auto const var_qualified_name(make_box<obj::symbol>(expr->var->n, expr->var->name));
+    if(__rt_ctx->opts.direct_linking && !(expr->var->dynamic))
+    {
+      if(root_fn->name.starts_with("jank_load"))
```
jeaye, line 329: This warrants a comment explaining why this is a special case.

compiler+runtime/include/cpp/jank/codegen/llvm_processor.hpp
```diff
@@ -67,6 +67,13 @@ namespace jank::codegen
     eval
   };
 
+  enum class var_root_kind : u8
```
jeaye: This warrants explanation. Also, if it's not used in the header, don't put it there!

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -18,6 +18,7 @@
 #include <jank/analyze/rtti.hpp>
 #include <jank/profile/time.hpp>
 #include <jank/util/fmt.hpp>
+#include <jank/util/fmt/print.hpp>
```
jeaye: This isn't needed. Was just used for debug code, presumably.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -251,9 +252,14 @@ namespace jank::codegen
                                 { ctx->builder->getPtrTy(), ctx->builder->getPtrTy() },
                                 false));
       auto const fn(ctx->module->getOrInsertFunction("jank_var_bind_root", fn_type));
-
-      llvm::SmallVector<llvm::Value *, 2> const args{ ref, gen(expr->value.unwrap(), arity) };
+      auto const var_root(gen(expr->value.unwrap(), arity));
+      llvm::SmallVector<llvm::Value *, 2> const args{ ref, var_root };
       ctx->builder->CreateCall(fn, args);
+      if(__rt_ctx->opts.direct_linking && target == compilation_target::module)
```
jeaye: This warrants explanation. Why modules only?

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -298,14 +304,26 @@ namespace jank::codegen
...
-    auto const ref(gen_var(make_box<obj::symbol>(expr->var->n, expr->var->name)));
-    auto const fn_type(
-      llvm::FunctionType::get(ctx->builder->getPtrTy(), { ctx->builder->getPtrTy() }, false));
-    auto const fn(ctx->module->getOrInsertFunction("jank_deref", fn_type));
-
-    llvm::SmallVector<llvm::Value *, 1> const args{ ref };
-    auto const call(ctx->builder->CreateCall(fn, args));
+    llvm::Value *call{};
+    auto const var_qualified_name(make_box<obj::symbol>(expr->var->n, expr->var->name));
+    if(__rt_ctx->opts.direct_linking && !(expr->var->dynamic))
```
jeaye: We need a TODO for `:deref` meta here.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1083,6 +1101,59 @@ namespace jank::codegen
     return ctx->builder->CreateLoad(ctx->builder->getPtrTy(), global);
   }
 
+  llvm::Value *
+  llvm_processor::gen_var_root(obj::symbol_ref qualified_name, var_root_kind kind) const
```
jeaye: These params can be `const`.

compiler+runtime/src/cpp/jank/jit/processor.cpp
```diff
@@ -179,8 +179,9 @@ namespace jank::jit
   {
     profile::timer const timer{ util::format("jit ir module {}",
                                              static_cast<std::string_view>(m->getName())) };
-    //m->print(llvm::outs(), nullptr);
-
+    // Print module
+    // m->print(llvm::outs(), nullptr);
+
```
jeaye: Please remove these changes.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -335,7 +336,8 @@ namespace jank::runtime
                               module_path.c_str(),
                               file_error.message()));
     }
-    //codegen_ctx->module->print(llvm::outs(), nullptr);
+    // Turn on printing
+    // module->print(llvm::outs(), nullptr);
```
jeaye: Please remove these changes.

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -25,6 +25,7 @@ namespace jank::util::cli
                    opts.profiler_file,
                    "The file to write profile entries (will be overwritten).");
     cli.add_flag("--gc-incremental", opts.gc_incremental, "Enable incremental GC collection.");
+    cli.add_flag("--fdirect-linking", opts.direct_linking, "Compile with direct linking.");
```
jeaye: I know I suggested this naming originally, but let's use `--direct-call` instead and update the option name, too. In C++, we actually have static linking, dynamic linking, etc. I'd rather use a different name than Clojure and avoid confusion with the rest of the native world.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1083,6 +1101,59 @@ namespace jank::codegen
...
+  {
+    auto it(ctx->var_root_globals.find(qualified_name));
+    if(it != ctx->var_root_globals.end())
+    {
+      if(kind == var_root_kind::runtime_value)
```
jeaye: This warrants a comment.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1083,6 +1101,59 @@ namespace jank::codegen
...
+      }
+      return it->second;
+    }
+    auto const name(util::format("var_root_{}", munge(qualified_name->to_string())));
+    auto const var_root(create_global_var(name));
+    ctx->module->insertGlobalVariable(var_root);
+    auto &global(ctx->var_root_globals[qualified_name]);
+    global = var_root;
+
+    if(kind == var_root_kind::raw_global)
```
jeaye: These two ifs warrant a comment.

compiler+runtime/include/cpp/jank/codegen/llvm_processor.hpp
```diff
@@ -144,6 +152,8 @@ namespace jank::codegen
     llvm::Value *gen(analyze::expr::case_ref, analyze::expr::function_arity const &);
 
     llvm::Value *gen_var(obj::symbol_ref qualified_name) const;
+    llvm::Value *gen_var_root(obj::symbol_ref qualified_name,
+                              var_root_kind kind = var_root_kind::runtime_value) const;
```
jeaye: Please don't use default values here. Just specify the value when you call the function. Default values should be used *very* sparingly snf er only call `gen_var_root` in three places.

jeaye (changes requested):
Nice work getting this going, Monty! I know you have documentation remaining, so I've left some notes as to where to add it. There are a couple of code changes requested, too.

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -25,6 +25,7 @@ namespace jank::util::cli
                    opts.profiler_file,
                    "The file to write profile entries (will be overwritten).");
     cli.add_flag("--gc-incremental", opts.gc_incremental, "Enable incremental GC collection.");
+    cli.add_flag("--fdirect-call", opts.direct_calls, "Compile with direct calls.");
```
jeaye: Please use `--direct-call`. Also, let's provide a more helpful message. The flag already says "direct call", but what does that mean? For example "Elide var derefs when compiling function calls."

compiler+runtime/src/cpp/jank/jit/processor.cpp
jeaye: This has not been addressed.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
jeaye: This has not been addressed.

compiler+runtime/include/cpp/jank/codegen/llvm_processor.hpp
```diff
@@ -66,6 +66,17 @@ namespace jank::codegen
     function,
     eval
   };
+  /* There are three places where a var-root could be generated depending on different circumstances.
```
jeaye: This comment is not connected to the `compilation_target` enum. Please use a new line for spacing.

compiler+runtime/include/cpp/jank/codegen/llvm_processor.hpp
jeaye: The second part of this hasn't been addressed. The values of this enum are internal to `llvm_processor`, so we should forward declare the enum in the header and then define it in the cpp file. We do this to keep our headers small, since that impacts both compile-time perf and jank's startup perf.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -251,9 +251,17 @@ namespace jank::codegen
                                 { ctx->builder->getPtrTy(), ctx->builder->getPtrTy() },
                                 false));
       auto const fn(ctx->module->getOrInsertFunction("jank_var_bind_root", fn_type));
-
-      llvm::SmallVector<llvm::Value *, 2> const args{ ref, gen(expr->value.unwrap(), arity) };
+      auto const var_root(gen(expr->value.unwrap(), arity));
+      llvm::SmallVector<llvm::Value *, 2> const args{ ref, var_root };
       ctx->builder->CreateCall(fn, args);
+      /* Here if compiling into a jank module, a var-root can be initialized directly with the parameter
```
jeaye, line 265: Please add a new line before this.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -298,14 +306,35 @@ namespace jank::codegen
...
-    auto const ref(gen_var(make_box<obj::symbol>(expr->var->n, expr->var->name)));
-    auto const fn_type(
-      llvm::FunctionType::get(ctx->builder->getPtrTy(), { ctx->builder->getPtrTy() }, false));
-    auto const fn(ctx->module->getOrInsertFunction("jank_deref", fn_type));
-
-    llvm::SmallVector<llvm::Value *, 1> const args{ ref };
-    auto const call(ctx->builder->CreateCall(fn, args));
+    llvm::Value *call{};
+    auto const var_qualified_name(make_box<obj::symbol>(expr->var->n, expr->var->name));
+    /*
```
jeaye, line 320: Please add a new line before this. Block comments like this should always have one. You'll find that to be true throughout the jank code base.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -298,14 +306,35 @@ namespace jank::codegen
...
+     * For direct-calls, when derefing a var we need to handle two different types of var-derefs.
+     * First, we only direct-call vars that are not dynamic.
+     * When generating the IR for a var-root, if the function is a jank_load function, the var_root is derefed directly in the jank_load functoin.
```
jeaye: 1. Spelling error: functoin
2. Please follow the style guidelines with regard to line length. We have a column limit of 100.
3. We say say there are two things and we identify"first", but not not "second"

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1083,6 +1112,64 @@ namespace jank::codegen
...
   }
 
+  llvm::Value *
+  llvm_processor::gen_var_root(obj::symbol_ref qualified_name, var_root_kind kind) const
+  {
+    auto it(ctx->var_root_globals.find(qualified_name));
+    if(it != ctx->var_root_globals.end())
+    {
+      /* If the var-root already exists, and is part of a normal var-deref, create a load in the IR
+       * Else if part of the special cases, just return the value of the var-root directly
```
jeaye: This comment has multiple issues.

1. It's describing *what* the code is doing, not *why*
2. The second line, in particular, is not a coherent sentence, since it's written as pseudo code
3. None of the sentences in the comments here and the rest of this function end in periods

Please treat comments as documentation, like you would read in a book or user guide. Sentences should be clear, use correct grammar and spelling, have correct punctuation, and have a clear meaning for being in the guide in the first place.

This is likely to be dissimilar to comments you have seen in your experience coding so far. Those comments were generally pretty awful. Please revisit ALL of your comments in this PR and write them with this in mind. **Explain clearly.**

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1083,6 +1112,64 @@ namespace jank::codegen
...
+    ctx->module->insertGlobalVariable(var_root);
+    auto &global(ctx->var_root_globals[qualified_name]);
+    global = var_root;
+
+    /* When generating a var-root in a context where we are already calling var_bind_root, we can just return the global var-root */
+    if(kind == var_root_kind::binded_def)
+    {
+      return global;
+    }
+    /* Directly derefing the var in the jank_load function */
```
jeaye: This is not a complete sentence. It lacks a subject.

But, more importantly, it's just not easy to read and understand.

Compare it to all of my comments, either in this file, or on your PR.

jeaye (changes requested):

PR #384, opened 2025-08-05
Remove bdwgc::gccpp and add bdwgc::gctba

jianlingzhong:
This is to prevent overriding the global new/deleted keywords

Overriding the global new/delete does not play well with third party libs like libtorch


jeaye:
Thanks!

PR #389, opened 2025-08-06
fix empty line exception

### LINKED ISSUE #387: Empty line leads to exceptions
```
user=>
Uncaught exception: unable to map file /tmp/jank-repl-dLCg4u due to error: Mapping failed for unknown reason

#4 in jank::runtime::context::eval_file at jank
#5 in main::$_0::operator at jank
#6 in jank_init at jank
#7 at libc.so.6
#8 in __libc_start_main at libc.so.6

user=>
```

JVM's repl:

```
user=>

user=>

user=>

```

Flinner:
Fix #387

### DIFF
--- a/compiler+runtime/src/cpp/main.cpp
+++ b/compiler+runtime/src/cpp/main.cpp
@@ -167,6 +167,12 @@ namespace jank
       auto &line(*buf);
       util::trim(line);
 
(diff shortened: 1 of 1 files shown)


jeaye:
Thanks!

PR #395, opened 2025-08-13
Add an issue template for bug reports

shantanu-sardesai:
Standardising bug reporting for jank.

### DIFF
--- a/.github/ISSUE_TEMPLATE/1.bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/1.bug_report.yml
@@ -6,11 +6,9 @@ body:
     attributes:
       label: jank health check
       description: |
-        Output from the `jank check-health` command.
-
-        While not required in all cases, including this in bug reports can help save everyone time 🙂.
+        Output from the `jank check-health` command. Also include details about your operating system, distribution, and architecture.
     validations:
-      required: false
+      required: true
 
   - type: textarea
     attributes:
@@ -41,38 +39,16 @@ body:
 
   - type: textarea
     attributes:
-      label: Backtrace
+      label: Stack trace
       description: |
-        If the crash involves a segmentation fault, please include a backtrace.
-
-        > For accurate debugger output, `jank` must be [built locally](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/doc/build.md#debug) with the `debug` flag enabled.
-        
-        > The current `jank` runtime depends on changes that have not yet been upstreamed to LLVM.  
-        > LLVM must also be built locally—follow [these instructions](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/doc/build.md#compiling-clangllvm) to build LLVM.
+        If the crash involves a segmentation fault, please include a stack trace.
 
         The steps differ slightly between `lldb` (commonly used on macOS) and `gdb` (commonly used on Linux).
-        - Run the `jank` binary inside the debugger:
-          - For the REPL:
-            ```sh
-            lldb -- jank repl
-            ```
-            or
-            ```sh
-            gdb --args jank repl
-            ```
-          - For running a `.jank` file:
-            ```sh
-            lldb -- jank run <file-name>.jank
-            ```
-            or
-            ```sh
-            gdb --args jank run <file-name>.jank
-            ```
-        - The debugger REPL will start.
+        - Run the `jank` binary inside the debugger.
         - Run the program by typing `r` in the debugger REPL.
         - Follow your reproduction steps until the program crashes.
         - When it crashes, you will return to the debugger REPL.
         - Type `bt` and paste the output here.
           - Ensure the output is properly formatted, with all stack frames (`frame #<n>`) listed in ascending order.
     validations:
-      required: true
+      required: false
--- a/.github/ISSUE_TEMPLATE/1.bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/1.bug_report.yml
@@ -12,18 +12,6 @@ body:
 
   - type: textarea
     attributes:
-      label: Environment information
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,78 @@
...
+    attributes:
+      label: Environment information
+      description: |
+        Details about your operating system, distribution, and architecture.
+      render: bash
+      placeholder: |
+        Operating System:
+        Arch:
+    validations:
+      required: true
```
jeaye: This is to be included in the health check. The health check should be required.

.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,78 @@
...
+  - type: textarea
+    attributes:
+      label: Backtrace
+      description: |
+        If the crash involves a segmentation fault, please include a backtrace.
+
+        > For accurate debugger output, `jank` must be [built locally](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/doc/build.md#debug) with the `debug` flag enabled.
+        
+        > The current `jank` runtime depends on changes that have not yet been upstreamed to LLVM.  
+        > LLVM must also be built locally—follow [these instructions](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/doc/build.md#compiling-clangllvm) to build LLVM.
```
jeaye: This is not correct. We can remove it entirely. It's impractical to expect someone with a production build of jank to compile jank themselves in order to get a debug stack trace.

A production stack trace is just fine and can be acquired via the same means.

.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,78 @@
...
+            gdb --args jank run <file-name>.jank
+            ```
+        - The debugger REPL will start.
+        - Run the program by typing `r` in the debugger REPL.
+        - Follow your reproduction steps until the program crashes.
+        - When it crashes, you will return to the debugger REPL.
+        - Type `bt` and paste the output here.
+          - Ensure the output is properly formatted, with all stack frames (`frame #<n>`) listed in ascending order.
+    validations:
+      required: true
```
jeaye: I don't think we can require this, since not every bug has a stack trace.

.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,78 @@
...
+        - The code being executed.
+        - Any additional setup details (e.g., third-party dependencies) not covered above.
+
+        Be as descriptive as needed—more detail helps us reproduce and fix the issue faster.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Backtrace
```
jeaye: Let's call this "Stack trace".

.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,78 @@
...
+description: A bug is when the program is not behaving as you'd expect while a crash is defined as the case where the program exits unexpectedly (Segmentation fault, etc.).
+labels: ["bug"]
+body:
+  - type: textarea
+    attributes:
+      label: jank health check
+      description: |
+        Output from the `jank check-health` command.
+
+        While not required in all cases, including this in bug reports can help save everyone time 🙂.
```
jeaye: Is there any way we can ensure this gets put into a code fence? Interpreting the output as markdown isn't going to work well, I think.

.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,78 @@
...
+        The steps differ slightly between `lldb` (commonly used on macOS) and `gdb` (commonly used on Linux).
+        - Run the `jank` binary inside the debugger:
+          - For the REPL:
+            ```sh
+            lldb -- jank repl
+            ```
+            or
+            ```sh
+            gdb --args jank repl
+            ```
```
jeaye: We can remove this. I don't think we need to spell out both of these scenarios. Just spin up lldb/gdb, run your program to trigger the issue, and then get the stack trace.

.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,78 @@
...
+            or
+            ```sh
+            gdb --args jank run <file-name>.jank
+            ```
+        - The debugger REPL will start.
+        - Run the program by typing `r` in the debugger REPL.
+        - Follow your reproduction steps until the program crashes.
+        - When it crashes, you will return to the debugger REPL.
+        - Type `bt` and paste the output here.
+          - Ensure the output is properly formatted, with all stack frames (`frame #<n>`) listed in ascending order.
```
jeaye, line 40: Ideally we can require that this is put in a code fence. Not sure if Github allows this.

jeaye (changes requested):
Thanks, Shantanu!

shantanu-sardesai:
Code fence would require more looking into, but based on their [docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema) and this [thread of discussion](https://github.com/orgs/community/discussions/10227) I'm not very hopeful. We might need to get creative with how we validate issues in that case.

shantanu-sardesai:
We could initially observe the way the community reports the bugs and add additional checks if required. For now re-requesting a review if in case this is not a blocking concern.

jeaye (approved):
Let's give it a go!

PR #397, opened 2025-08-15
Reparse source for interop errors

shantanu-sardesai:
- Reparse source for specific errors.
- Remove redundant notes from interop errors.


compiler+runtime/test/jank/cpp/member/fail-protected.jank
```diff
@@ -0,0 +1,11 @@
+(cpp/raw "namespace jank::cpp::member::fail_private
```
jeaye, line 1: This is a critical mistake, since this namespace was copied from another test. C++ doesn't allow defining types or functions or variables multiple times, which is why each of these tests has its own C++ namespace.

I will fix this when merging, though.

Issue #362, opened 2025-07-13
`\ ` should be treated as `\space`


jeaye:
This is a lexer change. Should be quite minimal.

PR #399, opened 2025-08-16
Fix `\ ` Lexing

E-A-Griffin:
Fixes https://github.com/jank-lang/jank/issues/362

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -574,6 +574,7 @@ namespace jank::read::lex
           {
             return ch.expect_err();
           }
+          // NOTE: Does not handle unicode space characters, though Clojure does
           else if(ch.expect_ok().character != ' ' && std::iswspace(static_cast<int>(ch.expect_ok().character)))
           {
             return error::lex_incomplete_character("A \\ must be followed by a character value.",
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -589,7 +589,7 @@ namespace jank::read::lex
 
           if(data == "\\ ")
           {
-            return ok(token{token_start, pos, token_kind::character, "\\space" });
+            return ok(token{ token_start, pos, token_kind::character, "\\space" });
           }
 
           return ok(token{ token_start, pos, token_kind::character, data });
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -575,7 +575,8 @@ namespace jank::read::lex
             return ch.expect_err();
           }
           // NOTE: Does not handle unicode space characters, though Clojure does
-          else if(ch.expect_ok().character != ' ' && std::iswspace(static_cast<int>(ch.expect_ok().character)))
+          else if(ch.expect_ok().character != ' '
+                  && std::iswspace(static_cast<int>(ch.expect_ok().character)))
           {
             return error::lex_incomplete_character("A \\ must be followed by a character value.",
                                                    { token_start, pos });
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -574,13 +574,6 @@ namespace jank::read::lex
           {
             return ch.expect_err();
           }
-          // NOTE: Does not handle unicode space characters, though Clojure does
-          else if(ch.expect_ok().character != ' '
-                  && std::iswspace(static_cast<int>(ch.expect_ok().character)))
-          {
-            return error::lex_incomplete_character("A \\ must be followed by a character value.",
-                                                   { token_start, pos });
-          }
 
           while(pos <= file.size())
           {
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -587,7 +587,8 @@ namespace jank::read::lex
 
           jtl::immutable_string_view const data{ file.data() + token_start, ++pos - token_start };
 
-          if(data == "\\ ") {
+          if(data == "\\ ")
+          {
             return ok(token{token_start, pos, token_kind::character, "\\space" });
           }
 
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -574,7 +574,7 @@ namespace jank::read::lex
           {
             return ch.expect_err();
           }
-          else if(std::iswspace(static_cast<int>(ch.expect_ok().character)))
+          else if(ch.expect_ok().character != ' ' && std::iswspace(static_cast<int>(ch.expect_ok().character)))
           {
             return error::lex_incomplete_character("A \\ must be followed by a character value.",
                                                    { token_start, pos });
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -586,6 +586,11 @@ namespace jank::read::lex
           }
 
           jtl::immutable_string_view const data{ file.data() + token_start, ++pos - token_start };
+
+          if(data == "\\ ") {
+            return ok(token{token_start, pos, token_kind::character, "\\space" });
+          }
+
           return ok(token{ token_start, pos, token_kind::character, data });
         }
       case ';':
--- a/compiler+runtime/test/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/test/cpp/jank/read/lex.cpp
@@ -1286,7 +1286,7 @@ namespace jank::read::lex
         CHECK(tokens
               == make_tokens({
                 { 0, 2, token_kind::character, "\\ "sv }
-        }))
+        }));
       }
 
       SUBCASE("Dangling \\")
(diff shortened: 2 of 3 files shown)


compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -574,7 +574,9 @@ namespace jank::read::lex
           {
             return ch.expect_err();
           }
-          else if(std::iswspace(static_cast<int>(ch.expect_ok().character)))
+          // NOTE: Does not handle unicode space characters, though Clojure does
```
jeaye: 1. Please only use `/* */` style comments. Also, comments are documentation, so please always use complete sentences, correct capitalization, and necessary punctuation. Imagine your comments are snippets from a book explaining why the code is as it is.

2. We have this `iswspace` function, which supports Unicode. We're using it here, but failing on anything which isn't an ASCII space. Why aren't we just accepting anything which is `iswspace`?

3. If we do, please add tests for those Unicode whitespace cases. If we don't, the comment this needs to be a TODO, since lack of parity with Clojure means we're not done yet.

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -1283,7 +1283,10 @@ namespace jank::read::lex
       {
         processor p{ R"(\ )" };
         native_vector<jtl::result<token, error_ref>> const tokens(p.begin(), p.end());
-        CHECK(tokens == make_results({ { make_error(kind::lex_incomplete_character, 0, 1) } }));
+        CHECK(tokens
+              == make_tokens({
+                { 0, 2, token_kind::character, "\\ "sv }
+        }));
```
jeaye: I think we need to test this on the parse side (in `test/cpp/jank/read/parse.cpp`), too. `\ ` and `\space` should provide the same character object. Right now, I'm not convinced they do.

jeaye (changes requested):
Thanks, Emma!

E-A-Griffin:
@jeaye running into something weird you may know the answer to
```
/Users/emma/programming/jank/compiler+runtime/test/cpp/jank/read/parse.cpp:215: ERROR: CHECK( r.expect_ok().unwrap().start == lex::token{ offset, len, lex::token_kind::character, ch } ) is NOT correct!
  values: CHECK( token(source_position(0, 1:1), source_position(2, 1:3), character, "\\space") == token(source_position(0, 1:1), source_position(6, 1:7), character, "\\space") )

/Users/emma/programming/jank/compiler+runtime/test/cpp/jank/read/parse.cpp:215: ERROR: CHECK( r.expect_ok().unwrap().start == lex::token{ offset, len, lex::token_kind::character, ch } ) is NOT correct!
  values: CHECK( token(source_position(2, 1:3), source_position(8, 1:9), character, "\\space") == token(source_position(6, 1:7), source_position(12, 1:13), character, "\\space") )
```

I copied some assertion templates for testing parsing `\ ` as `\space` but these seem to check that parsing certain regex strings matches against certain string literals, but when the expected string is longer than the lexed + parsed regex the source positions aren't going to line up. Should I skip checking this for this case or am I reading this wrong? (trying to handle this before I jump to the unicode cases)

jeaye:
> @jeaye running into something weird you may know the answer to
> 
> ```
> /Users/emma/programming/jank/compiler+runtime/test/cpp/jank/read/parse.cpp:215: ERROR: CHECK( r.expect_ok().unwrap().start == lex::token{ offset, len, lex::token_kind::character, ch } ) is NOT correct!
>   values: CHECK( token(source_position(0, 1:1), source_position(2, 1:3), character, "\\space") == token(source_position(0, 1:1), source_position(6, 1:7), character, "\\space") )
> 
> /Users/emma/programming/jank/compiler+runtime/test/cpp/jank/read/parse.cpp:215: ERROR: CHECK( r.expect_ok().unwrap().start == lex::token{ offset, len, lex::token_kind::character, ch } ) is NOT correct!
>   values: CHECK( token(source_position(2, 1:3), source_position(8, 1:9), character, "\\space") == token(source_position(6, 1:7), source_position(12, 1:13), character, "\\space") )
> ```
> 
> I copied some assertion templates for testing parsing `\ ` as `\space` but these seem to check that parsing certain regex strings matches against certain string literals, but when the expected string is longer than the lexed + parsed regex the source positions aren't going to line up. Should I skip checking this for this case or am I reading this wrong? (trying to handle this before I jump to the unicode cases)

Hm, there must be some misunderstanding here. A couple of things to note:

1. There is no regex going on here. The `R"()"` stuff is just C++ raw strings, which don't require escapes.
2. Source positions *must* be correct in tests. If source positions are not correct, errors will not be pointed at the correct place.

From your failing test code, I believe I see the issue. In the source, `\ ` takes up two characters, whereas `\space` takes up six. Both are correct, but your test code is always incrementing by six (`ch.size()`) instead of by the size of the token.

So, even though we're internally representing it as `\space`, it's written as `\ ` and the source info matches that (as it should).

E-A-Griffin:
@jeaye thanks so much! Question about current state of lexing unicode characters. In my clojure repl the reader can handle `\ ` or `\u00a0` for the same character, but `jank` seems to only accept the latter, is this intentional, or maybe indicative of something wrong with my setup on my local end? 

```c++
  static jtl::result<codepoint, error_ref>
  convert_to_codepoint(jtl::immutable_string_view const sv, movable_position const &pos)
  {
    std::mbstate_t state{};
    wchar_t wc{};
    auto const len{ std::mbrtowc(&wc, sv.data(), sv.size(), &state) };

    if(std::cmp_equal(len, static_cast<size_t>(-1)))
    {
      return error::lex_invalid_unicode("Unfinished character.", pos);
    }
    else if(std::cmp_equal(len, static_cast<size_t>(-2)))
    {
      return error::lex_invalid_unicode("Invalid Unicode character.", pos);
    }
    return ok(codepoint{ static_cast<char32_t>(wc), static_cast<u8>(len) });
  }
```

Seems to return an error in the latter case from this function

E-A-Griffin:
```sh
emma@emma ~/programming/jank/compiler+runtime (egriffin/fix-lex-space-character)$ ./build/jank repl
user=> \ 
─ lex/invalid-unicode ──────────────────────────────────────────────────────────────────────────────
error: Invalid Unicode character.                                                                   
                                                                                                    
     │ /var/folders/9v/3f6t_wfj2ql2tsv0g8zkk5fr0000gp/T/jank-repl-p4UlnY                            
  1  │ \                                                                                            
     │  ^ Found here.                                                                               
```

I think this is probably worth handling in a separate issue, but your call, it does seem like the jank reader lexes and parses the `\uWXYZ` format properly and after closer inspection, my original code comment was wrong and Clojure reads these as specific unicode characters distinct from space (i.e. `(= \  \space) ;; ASCII space` `(not= \  \space) ;; Unicode Non-breaking space`)

E-A-Griffin:
> ```shell
> emma@emma ~/programming/jank/compiler+runtime (egriffin/fix-lex-space-character)$ ./build/jank repl
> user=> \ 
> ─ lex/invalid-unicode ──────────────────────────────────────────────────────────────────────────────
> error: Invalid Unicode character.                                                                   
>                                                                                                     
> ─────┬──────────────────────────────────────────────────────────────────────────────────────────────
>      │ /var/folders/9v/3f6t_wfj2ql2tsv0g8zkk5fr0000gp/T/jank-repl-p4UlnY                            
> ─────┼──────────────────────────────────────────────────────────────────────────────────────────────
>   1  │ \                                                                                            
>      │  ^ Found here.                                                                               
> ─────┴──────────────────────────────────────────────────────────────────────────────────────────────
> ```
> 
> I think this is probably worth handling in a separate issue, but your call, it does seem like the jank reader lexes and parses the `\uWXYZ` format properly and after closer inspection, my original code comment was wrong and Clojure reads these as specific unicode characters distinct from space (i.e. `(= \ \space) ;; ASCII space` `(not= \  \space) ;; Unicode Non-breaking space`)

This doesn't seem unique to unicode whitespace characters, it seems like while lexing hex unicode characters is supported, lexing unicode text literals (if that's the correct term) isn't, so I think that's out of scope

E-A-Griffin:
Yeah I think this is probably ready for re-review, existing unicode lexing issues acknowledged

jeaye:
> > ```shell
> > emma@emma ~/programming/jank/compiler+runtime (egriffin/fix-lex-space-character)$ ./build/jank repl
> > user=> \ 
> > ─ lex/invalid-unicode ──────────────────────────────────────────────────────────────────────────────
> > error: Invalid Unicode character.                                                                   
> >                                                                                                     
> > ─────┬──────────────────────────────────────────────────────────────────────────────────────────────
> >      │ /var/folders/9v/3f6t_wfj2ql2tsv0g8zkk5fr0000gp/T/jank-repl-p4UlnY                            
> > ─────┼──────────────────────────────────────────────────────────────────────────────────────────────
> >   1  │ \                                                                                            
> >      │  ^ Found here.                                                                               
> > ─────┴──────────────────────────────────────────────────────────────────────────────────────────────
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this is probably worth handling in a separate issue, but your call, it does seem like the jank reader lexes and parses the `\uWXYZ` format properly and after closer inspection, my original code comment was wrong and Clojure reads these as specific unicode characters distinct from space (i.e. `(= \ \space) ;; ASCII space` `(not= \  \space) ;; Unicode Non-breaking space`)
> 
> This doesn't seem unique to unicode whitespace characters, it seems like while lexing hex unicode characters is supported, lexing unicode text literals (if that's the correct term) isn't, so I think that's out of scope

We do have support for lexing Unicode: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/test/cpp/jank/read/lex.cpp#L1735

However, it could be that it's not being used when lexing characters (starting with `\`). This looks like it warrants a ticket of its own, as you said. Up for creating it?

jeaye:
Thanks, Emma!

PR #402, opened 2025-08-16
fix: conflicts due to non-namespaced unique_string

Samy-33:


.gitignore
```diff
@@ -4,6 +4,7 @@ jank-generated.hpp
 a.out
 .jank-native-repl-history
 .jank-repl-history
+**/.envrc
```
jeaye: No need for `**/` here. By default, a name on its own will match in any directory.

compiler+runtime/bin/jank/compiler+runtime/core.clj
```diff
@@ -23,7 +23,7 @@
                                            :coverage (util/get-env "JANK_COVERAGE" "off")})
 
   ; Bash tests
-  #_(jank.compiler+runtime.bash-test/-main {:enabled? build?})
+  (jank.compiler+runtime.bash-test/-main {:enabled? build?})
```
jeaye: CI in this PR is failing, from what I can tell. Is that expected? If so, we need to leave these disabled.

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -180,6 +181,9 @@ namespace jank::util
     clang::driver::Driver driver{ clang_path.c_str(), target_triple, diags, "jank", vfs };
     driver.setCheckInputsExist(true);
 
+    /* The first argument should be the clang executable. */
+    args.insert(args.begin(), clang_path.c_str());
```
jeaye, line 185: I imagine this needs a `strdup` as well, since we're expecting to `free` later.

jeaye (changes requested):
Thanks, Saket!

compiler+runtime/src/cpp/jank/util/clang.cpp
Samy-33: We `free` in the caller method. Because the `args` vector is passed by value and the `clang_path` is a cpp string, shouldn't it be freed automatically at the end of this function's execution?

compiler+runtime/bin/jank/compiler+runtime/core.clj
Samy-33: Disabled these tests as the binary can't find the precompiled headers yet.

compiler+runtime/src/cpp/jank/util/clang.cpp
jeaye: Ah, right, `args` is copied. Ok then, no change needed.

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -90,7 +90,7 @@ namespace jank::util::cli
     cli_compile_module
       .add_option("module", opts.target_module, "Module to compile (must be on the module path).")
       ->required();
-    cli_compile_module.add_option("-o", opts.output_filename, "Output object file name.");
+    cli_compile_module.add_option("-o", opts.output_object_filename, "Output object file name.");
```
jeaye, line 93: Was this needed because the output during AOT is set to an executable name, but we tried to write our object files to that?

Samy-33: Yes

.gitignore
```diff
@@ -32,6 +33,7 @@ inst
 deb-cache
 build
 compiler+runtime/build-*
+compiler+runtime/third-party
```
jeaye: No, this is not right. We have submodules in this directory, which are part of git. By ignoring it, all of our submodule usage will be broken. We will also be unable to add any new submodules to that directory, since git will never see them.

jeaye:
Thanks, Saket!

PR #405, opened 2025-08-20
Remove mentions of `import`

E-A-Griffin:
As per https://clojurians.slack.com/archives/C03SRH97FDK/p1755641194640599, `import` and `:import` are not officially supported in jank, though the var for `import` should still be bound, though to a function that always throws.

I did my best to keep docstrings and error messages as grammatically correct, full sentences, but I'm open to suggestions since I know a lot of the existing examples are carry-overs from clj/cljs that don't match the intent of comments/docstrings/error messages in jank.

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -5102,7 +5102,7 @@
   (throw "TODO: port await-for"))
 
 (defn import
-  "import is not implemented for jank, but var is still bound for portability. import always throws an exception"
+  "import is not implemented for jank, but a var is still bound to its symbol for portability. import always throws an exception"
   [& _]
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -5101,22 +5101,10 @@
...
-                   import-symbols-or-lists)]
-    `(do ~@(map #(list 'clojure.core/import* %)
-                (reduce (fn [v spec]
-                          (if (symbol? spec)
-                            (conj v (name spec))
-                            (let [p (first spec) cs (rest spec)]
-                              (into v (map #(str p "." %) cs)))))
-                        [] specs)))))
+(defn import
+  "import is not implemented for jank, but a var is still bound to its symbol for portability. import always throws an exception"
```
E-A-Griffin, line 5105: I could also add the suggestion about the workaround in case folks are using `(doc import)` in the repl and looking to understand how to use interop in jank

jeaye (approved):
Thanks, Emma!

PR #407, opened 2025-08-20
fix: Named Variadic Recursion

### LINKED ISSUE #302: Named recursion for variadic function, across another function, leads to JIT linker error
```
clojure.core=> (defn foo [& args] (fn [] (foo 1 2)))
JIT session error: Symbols not found: [ clojure_core_foo_7608_2 ]
Failure value returned from cantFail wrapped call
Failed to materialize symbols: { (main, { clojure_core_fn_7609_0, __orc_init_func.clojure_core-clojure.core$clojure_core_foo_7608-7615, clojure_core_jank_global_init_7614, clojure_core_repl_fn_7612_0, clojure_core_foo_7608_1, $.clojure_core-clojure.core$clojure_core_foo_7608-7615.__inits.0 }) }
UNREACHABLE executed at /nix/store/6pxb0i74lizr4s7qkjfq1spqn3y93zsn-llvm-19.1.5-dev/include/llvm/Support/Error.h:766!
Aborted (core dumped)
```

This is a codegen problem with named recursion. We're trying to call `clojure_core_foo_7608_2`, where the `_2` is the arity, but the arity should be `_1` since it only takes one arg: `args`. We need to see that the fn is variadic and act accordingly.

stmonty:
Fixes #302

### DIFF
--- a/compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
+++ b/compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
@@ -1164,7 +1164,8 @@ namespace jank::codegen
 
     if(expr->recursion_ref.fn_ctx->is_variadic)
     {
-      arg_handles.emplace_back(gen_function_instance(arity.fn_ctx->fn.as_ref(), arity));
+      arg_handles.emplace_back(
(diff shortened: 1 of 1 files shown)


jeaye:
Thanks, Monty!

PR #419, opened 2025-08-24
Add nix flake support for custom Clang/LLVM build

kylc:
To test:

```
nix develop . --print-build-logs
# wait ...
cd compiler+runtime/
./bin/configure -GNinja -DCMAKE_BUILD_TYPE=Release && ./bin/compile
./build/jank check-health
```

### DIFF
--- a/llvm.nix
+++ b/llvm.nix
@@ -2,6 +2,7 @@
   stdenv,
   lib,
   pkgs,
+  wrapCC,
   ...
 }: let
   gccForLibs = stdenv.cc.cc;
@@ -27,7 +28,7 @@
     "-L${stdenv.cc.libc}/lib"
   ];
 in
-  stdenv.mkDerivation {
+  wrapCC (stdenv.mkDerivation {
     pname = "llvm-jank";
     version = "21.0.0-git";
 
@@ -79,10 +80,5 @@ in
       )
     '';
 
-    # Create wrapped clang/clang++ executables for compiling the jank compiler,
-    # and for in turn compiling jank code.
-    postInstall = ''
-      wrapProgram $out/bin/clang --add-flags "${runtimeCompileFlags + " " + runtimeLinkFlags}"
-      wrapProgram $out/bin/clang++ --add-flags "${runtimeCompileFlags + " " + runtimeLinkFlags}"
-    '';
-  }
+    passthru.isClang = true;
+  })


compiler+runtime/CMakeLists.txt
```diff
@@ -192,7 +192,7 @@ separate_arguments(clang_system_include_dirs)
 
 set(clang_system_include_flags "")
 foreach(dir ${clang_system_include_dirs})
-  list(APPEND clang_system_include_flags -I${dir})
+  list(APPEND clang_system_include_flags ${dir})
```
kylc, line 195: On my system, the output from the command above already includes `-I` prefixes. This results in strings like `-I-I/include/path` being passed to the compiler. Please double check.

jeaye: Hm, you're right. Looks like we're adding it in the `sed` above.

jeaye:
Thanks, Kyle! Very exciting to see this. We've had a few people try to get this working (myself included), with no success. I'm giving this a go right now. If all works well, I'll be very happy to merge.

jeaye:
Were you able to build and run jank with this LLVM? I built it, configured jank with:

```bash
./bin/configure -GNinja -DCMAKE_BUILD_TYPE=Debug -Djank_test=on -Djank_sanitize=none -Djank_local_clang=off -Djank_analyze=off
```

But I cannot run the produced jank:

```
❯ ./build/jank-phase-1
Could not start dynamically linked executable: ./build/jank-phase-1
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
```

Even though I have all of the dynamic libs:

```
❯ ldd build/jank-phase-1
	linux-vdso.so.1 (0x00007ffff7fc4000)
	libzip.so.5 => /nix/store/my45gaz55q1gxdwdkfb4ixyb2sks14gx-libzip-1.11.4/lib/libzip.so.5 (0x00007ffff7f9d000)
	libz.so.1 => /nix/store/97jdzvwjgwy2g4xcijimadl0vpj6laqh-zlib-1.3.1/lib/libz.so.1 (0x00007ffff7f7f000)
	libclang-cpp.so.21.0git => /nix/store/dksfg3b5x2zngv6wghyqcxihcsv8fnjp-llvm-jank-21.0.0-git/lib/libclang-cpp.so.21.0git (0x00007ffff2400000)
	libcrypto.so.3 => /nix/store/lyl9yxxz8a3mlaxvm0jln6mglpbf2fha-openssl-3.5.1/lib/libcrypto.so.3 (0x00007ffff1c00000)
	libLLVM.so.21.0git => /nix/store/dksfg3b5x2zngv6wghyqcxihcsv8fnjp-llvm-jank-21.0.0-git/lib/libLLVM.so.21.0git (0x00007fffed400000)
	libm.so.6 => /nix/store/lmn7lwydprqibdkghw7wgcn21yhllz13-glibc-2.40-66/lib/libm.so.6 (0x00007ffff7e95000)
	libgcc_s.so.1 => /nix/store/fkw48vh7ivlvlmhp4j30hy2gvg00jgin-gcc-14.3.0-lib/lib/libgcc_s.so.1 (0x00007ffff7e67000)
	libc.so.6 => /nix/store/lmn7lwydprqibdkghw7wgcn21yhllz13-glibc-2.40-66/lib/libc.so.6 (0x00007fffed000000)
	/lib64/ld-linux-x86-64.so.2 => /nix/store/lmn7lwydprqibdkghw7wgcn21yhllz13-glibc-2.40-66/lib64/ld-linux-x86-64.so.2 (0x00007ffff7fc6000)
	libstdc++.so.6 => /nix/store/fkw48vh7ivlvlmhp4j30hy2gvg00jgin-gcc-14.3.0-lib/lib/libstdc++.so.6 (0x00007fffecc00000)
	libdl.so.2 => /nix/store/lmn7lwydprqibdkghw7wgcn21yhllz13-glibc-2.40-66/lib/libdl.so.2 (0x00007ffff7e60000)
	libpthread.so.0 => /nix/store/lmn7lwydprqibdkghw7wgcn21yhllz13-glibc-2.40-66/lib/libpthread.so.0 (0x00007ffff7e5b000)
	librt.so.1 => /nix/store/lmn7lwydprqibdkghw7wgcn21yhllz13-glibc-2.40-66/lib/librt.so.1 (0x00007ffff7e56000)
```

I looked into overriding the ld interpreter, but it's already set to what it should be.

```bash
❯ patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2 build/jank-phase-1

❯ patchelf --print-interpreter build/jank-phase-1
/lib64/ld-linux-x86-64.so.2

❯ file build/jank-phase-1
build/jank-phase-1: ELF 64-bit LSB pie executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.10.0, with debug_info, not stripped
```

kylc:
It does work on my system, but I suspect it's because I have `programs.nix-ld.enable = true;` in my NixOS configuration. I will try to get it working without nix-ld.

In the meantime, you probably want something like this:
```
patchelf --set-interpreter $(cat $NIX_CC/nix-support/dynamic-linker) build/jank
```

jeaye:
> In the meantime, you probably want something like this:
> 
> ```
> patchelf --set-interpreter $(cat $NIX_CC/nix-support/dynamic-linker) build/jank
> ```

You're exactly right. That did the trick. I had a vague memory of needing to grab that linker from somewhere, but couldn't find any search results with that in it.

> I will try to get it working without nix-ld.

I'm surprised this isn't using Nix's ld by default. The flake we had before, using LLVM from nixpkgs, was. Perhaps this is due to the LLVM we're configuring/building not being set up to use Nix's ld for linking?

Let's see if we can make it happen, though. Ideally this flake can work without relying on other global Nix configs.

Thanks so much for tackling this, Kyle. It'll allow me to stop working in distroboxes, since NixOS is my desktop OS. :slightly_smiling_face:

kylc:
The latest commit works for me without `nix-ld` in my config. It uses the same builtin `wrapCC` that the standard nixpkgs gcc/clang use to wrap the compiler executables with all the necessary nix-specific flags.

jeaye:
> The latest commit works for me without `nix-ld` in my config. It uses the same builtin `wrapCC` that the standard nixpkgs gcc/clang use to wrap the compiler executables with all the necessary nix-specific flags.

Excellent work, Kyle! I've tried this locally and it works very well.

On the off chance that you're up for it, we've had a few different people try to package jank for Nix, but none have succeeded. Now that we have working packages for Ubuntu and AUR and Homebrew, it may be a much better time to try. If you're up for it, our AUR build would be the easiest to follow. It's here: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=jank-git

PR #420, opened 2025-08-24
Fix odr violation

shantanu-sardesai:
Oops.


jeaye:
I commented this in your original PR, but it seems to have been forgotten. I've since fixed it on my branch. Also, this PR is insufficient for actually fixing the test issue, since you still use `fail_private` later on in the test.

PR #424, opened 2025-08-28
Add uuid support

djblue:
I was able to make some progress for https://github.com/jank-lang/jank/pull/193, sorry it took so long. The major updates are:

- Switched to https://github.com/mariusbancila/stduuid instead of using boost
  - Not sure I have the cmake stuff done correctly
  - Also, I couldn't get `-DUUID_SYSTEM_GENERATOR=ON` to link so I went with `std::random_device rd` for generating uuids
- Updated the code to use `object_ref`, `jtl::immutable_string` and `uhash`
- Removed runtime tests and added a few more compiler tests
- Update the reader to work as it was also broken from the previous version

I can't seem to figure out how to get `gen_global` to work correctly as was described in the previous PR. I get the following error with the current implementation:
```clojure
user=> (fn [] #uuid "21f06332-ef0d-4156-85ef-11a1601921d2")
Uncaught exception: Unimplemented constant codegen: 21f06332-ef0d-4156-85ef-11a1601921d2
```
I wasn't able to carry over any of the commits from the previous PR so I though it would be best to open a new one.


compiler+runtime/src/cpp/jank/runtime/obj/uuid.cpp
```diff
@@ -0,0 +1,61 @@
...
+  {
+    static std::random_device rd;
+    std::mt19937 g(rd());
+    uuids::uuid_random_generator gen{ g };
+    value = gen();
+  }
+
+  uuid::uuid(jtl::immutable_string const &s)
+  {
+    value = uuids::uuid::from_string(static_cast<std::string>(s)).value();
```
jeaye: No need to create a `std::string`. Just use `s.c_str()` and we can avoid another allocation.

compiler+runtime/src/cpp/jank/runtime/obj/uuid.cpp
```diff
@@ -0,0 +1,61 @@
...
+    return buff.release();
+  }
+
+  uhash uuid::to_hash() const
+  {
+    if(hash)
+    {
+      return hash;
+    }
+    static auto to_hash = std::hash<uuids::uuid>{};
```
jeaye: Pleas always use direct initialization. Also, there is no need for `auto` here.

compiler+runtime/test/jank/reader-macro/uuid/fail-literal-invalid-string.jank
```diff
@@ -0,0 +1 @@
+#uuid "random-string"
```
jeaye, line 1: I think we'll want a test with an empty string, too.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -810,6 +812,46 @@ namespace jank::read::parse
...
+    {
+      auto str_result(next());
+
+      if(str_result.is_err())
+      {
+        return str_result;
+      }
+      else if(str_result.expect_ok().is_none())
+      {
+        return error::parse_invalid_reader_symbolic_value("value after #uuid must be present",
```
jeaye: Please use complete sentences and correct grammar for error messages. Also, try to be as helpful and actionable as you can. Use single quotes for references to types or values. For example:

> A string literal value after '#uuid' must be present.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -810,6 +812,46 @@ namespace jank::read::parse
     return object_source_info{ wrapped, start_token, sym_end };
   }
 
+  processor::object_result processor::parse_reader_macro_tagged()
+  {
+    auto const start_token(token_current.latest.unwrap().expect_ok());
+    auto sym_result(next());
```
jeaye: I think this can be `const`.

jeaye (changes requested):
Great job, Chris. I've identified some things to change, but overall this looks good.

To fix your codegen issue, look at then `gen` function for `primitive_literal`, which is the expression type you're building for uuids.

djblue:
@jeaye I think I made all the updates you requested correctly. https://github.com/jank-lang/jank/actions/runs/17308535851/job/49144500329 failed, but uncertain why. Is there an easy way to reproduce this locally?

jeaye:
> @jeaye I think I made all the updates you requested correctly. https://github.com/jank-lang/jank/actions/runs/17308535851/job/49144500329 failed, but uncertain why. Is there an easy way to reproduce this locally?

Take a look at the errors there. Each one should be pretty clear. You can reproduce locally by configuring with `-Djank_analyze=on`.

djblue:
@jeaye Ahh, I found the relevant errors when I looked closely, thanks 💯 

It looks like `uuids::uuid::from_string` returns a `std::optional` that I needed to check before I could use the value.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -810,6 +812,49 @@ namespace jank::read::parse
...
+          "A string literal value after '#uuid' must be present.",
+          { start_token.start, latest_token.end });
+      }
+
+      auto const str_end(str_result.expect_ok().unwrap().end);
+
+      if(str_end.kind != lex::token_kind::string)
+      {
+        return error::parse_invalid_reader_symbolic_value(
+          "A string literal value after '#uuid' must be present.",
```
jeaye, line 842: As a general rule for good error messaging, if you find yourself repeating the same message for different problems, the message is not as helpful as it can be. In this case, the message works well for the first error, but for the second one it's orthogonal to the problem, since the user DID provide a value, but the value is not a string literal. So, instead, a more helpful message would be something like:

> The value after '#uuid' must be a string literal.

compiler+runtime/src/cpp/jank/runtime/obj/uuid.cpp
```diff
@@ -0,0 +1,71 @@
+#include <uuid.h>
```
jeaye, line 1: This is included in `uuid.hpp`, so it doesn't buy us anything here.

PR #425, opened 2025-08-28
Fix: LLVM Absolute Path build issues on MacOS

stmonty:
All credit to Michael Bradley on the Clojurian Slack for this fix.
Just making this PR to resolve the issue that MacOS users are having.

I tested on my Linux build that the change to the executing bash command did not break it, but have not personally tested on MacOS. Tested by others on the Clojurian Slack.

### DIFF
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -180,7 +180,8 @@ endfunction()
 # can configure jank correctly during the build phase and then it'll just
 # continue to work with that setup once it's installed.
 execute_process(
-  COMMAND bash -c "${CMAKE_CXX_COMPILER} -xc++ -E -v /dev/null 2>&1 | sed -n -e '/^.include/,\${' -e '/^ \\/.*/p' -e '}' | sed 's/^[[:space:]]\\{1,\\}/-I/' | sed 's/ (framework directory)//'"  RESULT_VARIABLE clang_system_includes_result
+  COMMAND bash -c "${CMAKE_CXX_COMPILER} -xc++ -E -v /dev/null 2>&1 | sed -n -e '/^.include/,\${' -e '/^ \\/.*/p' -e '}' | sed 's/^[[:space:]]\\{1,\\}/-I/' | sed 's/ (framework directory)//'"
+  RESULT_VARIABLE clang_system_includes_result
   OUTPUT_VARIABLE clang_system_includes_unescaped
   OUTPUT_STRIP_TRAILING_WHITESPACE
 )


shantanu-sardesai:
Yup, works on my local.

michaelsbradleyjr:
Thanks, @stmonty!

An issue here, I think, is that a proper fix is also needed re: `CMAKE_INSTALL_RPATH` which is not yet introduced in `CMakeLists.txt`.

compiler+runtime/CMakeLists.txt
```diff
@@ -176,8 +180,7 @@ endfunction()
 # can configure jank correctly during the build phase and then it'll just
 # continue to work with that setup once it's installed.
 execute_process(
-  COMMAND bash -c "${CMAKE_CXX_COMPILER} -xc++ -E -v /dev/null 2>&1 | sed -n -e '/^.include/,\${' -e '/^ \\/.*/p' -e '}' | sed 's/^\\s\\+/-I/' | sed 's/ (framework directory)//'"
-  RESULT_VARIABLE clang_system_includes_result
+  COMMAND bash -c "${CMAKE_CXX_COMPILER} -xc++ -E -v /dev/null 2>&1 | sed -n -e '/^.include/,\${' -e '/^ \\/.*/p' -e '}' | sed 's/^[[:space:]]\\{1,\\}/-I/' | sed 's/ (framework directory)//'"  RESULT_VARIABLE clang_system_includes_result
```
michaelsbradleyjr: Looks like a new line before `RESULT_VARIABLE` was accidentally removed.

michaelsbradleyjr: Also, in case it's not apparent why these changes were made, it's because the built-in `sed` on macOS is the BSD variety not GNU and there are subtle differences in what expressions they support. `[[:space:]]` et al. are universal, so doing it this way supports BSD and GNU varieties of `sed`.

It's easy for this difference to be masked on macOS if you `brew install gnu-sed` and setup `sed` as an alias to `gsed`. I've bumped into that problem several times over the years.

jeaye:
I understand the `sed` issue. Is the rpath issue a blocker, or does it just lead to warnings? If it's a blocker, have we tried removing this instead? https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/cmake/install.cmake#L2

I know Monty is at work right now, so these questions are mainly for @michaelsbradleyjr.

michaelsbradleyjr:
With no additional changes relative to this PR, `./bin/install` failed on my system with:
```
CMake Error at build/cmake_install.cmake:8760 (file):
  file INSTALL cannot find
  "/Users/michael/repos/jank/compiler+runtime/build/llvm-install/usr/local/lib/libLLVM.so.21.0git":
  No such file or directory.
```

That being said, the changes in this PR do allow `./bin/compile` to succeed on macOS whereas before it was failing.

I will look into the suggested change  in `install.cmake` and other changes that may be necessary. Should an attempt be made to resolve those things in this PR or a later one?

compiler+runtime/CMakeLists.txt
stmonty: Fixed the new-line typo, thanks for catching that.

jeaye:
Thanks!

PR #426, opened 2025-08-28
Add missing actions/checkout in build-clang job

michaelsbradleyjr:
When no cache is present, e.g. in initial GHA runs of a fork, the build-clang job would fail because an expected path does not exist. Cloning the repo before the "Build Clang/LLVM" step fixes that problem; it is not necessary for this purpose to checkout submodules.


jeaye:
Thanks. We'll also need to install the necessary apt packages to build. I will figure out what those are and add that.

PR #427, opened 2025-08-28
Use commit hash in cache key for build-clang job

michaelsbradleyjr:
If the `jank` branch of `jank-lang/llvm-project` is updated then Clang/LLVM should be rebuilt.

You may not desire this change in caching behavior, but I thought it might be useful.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,24 +21,37 @@ jobs:
     name: Build Clang
     runs-on: ubuntu-24.04
     timeout-minutes: 150
+    outputs:
+      sha: ${{ steps.cache-clang-key.outputs.sha }}
     steps:
+      - name: Get latest commit on jank-lang/llvm-project@jank
+        id: cache-clang-key
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
+        with:
+          script: |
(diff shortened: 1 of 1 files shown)


jeaye:
I think this is a neat idea! However, it's not something I want to merge because it's a bit too magical (indirect). My CI battle scars force me to keep our CI as simple and dumb as it possibly can be. We don't have an action step unless we absolutely need it. This is also why the whole flow was written in babashka as well, so we could actually reproduce CI results in a docker container.

I apologize for closing this after you've done the work. If you're ever unsure if I'll want to merge, ask me in Slack!

michaelsbradleyjr:
It's no problem. In this case I thought it would be easier to show how it would work in a PR rather than chat about it, but I'll lean toward asking in chat first going forward.

I'm also keeping in mind your preference to do as much as possible in `babashka`, and I really like that idea.

In the case of caching specific to Github Actions, it's tricky because seems like it might not be a good idea to try to account for that in e.g. `check_everthing.clj` i.e. it's starting to mix concerns of "how to build" with concerns of reducing build time in CI.

Can chat about it in Slack, but if jank will be using a fork of Clang/LLVM for the long haul, should making the `jank-lang/llvm-project` fork a submodule of this repo be considered? It would then be more natural to check if cached clang should be rebuilt.

PR #435, opened 2025-08-30
Add `#cpp` tagged literal for c-style strings

djblue:


compiler+runtime/test/jank/reader-macro/cstring/fail-missing-string-cstring.jank
```diff
@@ -0,0 +1 @@
+#cpp
```
jeaye: Let's move these to `test/jank/reader-macro/cpp/string/`, since the reader macro is `cpp` and the value that follows is a string. That way we can group all `#cpp` types under that same directory.

compiler+runtime/test/jank/reader-macro/cstring/pass-small-cstring.jank
```diff
@@ -0,0 +1,3 @@
+#cpp "hello"
```
jeaye: We need a test to ensure the escaping is handled properly. i.e. `#cpp "\"foo\"\n\\bar"` needs to come out right on the other end.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -849,8 +849,42 @@ namespace jank::read::parse
...
+          "The form after '#cpp' must be a string literal.",
+          { start_token.start, latest_token.end });
+      }
+
+      auto const str(expect_object<obj::persistent_string>(str_result.expect_ok().unwrap().ptr));
+
+      auto const wrapped(
+        make_box<obj::persistent_list>(std::in_place,
+                                       make_box<obj::symbol>("cpp/value"),
+                                       make_box(util::format("\"{}\"", str->data)).erase()));
```
jeaye: I think we'll need `util::escape` for the string. Check out my test suggestions below.

jeaye (changes requested):
Looks good! Just a few things to note.

jeaye:
Nice! All set for this to merge?

djblue:
@jeaye, yup, I think this PR is good to merge 👍

jeaye:
Great job, Chris! This is a huge win.

PR #436, opened 2025-08-30
Fix: Extra test for Var within Vars

stmonty:


compiler+runtime/test/jank/call/pass-var-within-var.jank
```diff
@@ -6,4 +6,13 @@
...
 (assert (== 1 (c 1)))
 
+(defn d [x]
+  (identity x))
+
+(def e #'d)
+
+(def f #'e)
+
+(assert (== 2 (f 2)))
```
jeaye: This doesn't test for anything different. :upside_down_face: 

We wanted the call to have codegen, but it's still evaluated.

jeaye (changes requested):

compiler+runtime/test/jank/call/pass-var-within-var.jank
stmonty: I see, I misunderstood and thought that the original var needed to be codegen'd, but you're saying the call to the var needs to be wrapped in a function?

jeaye: Your code change for vars within vars was in `evaluate`, which is the code path taken by running top-level forms  in a file or repl. But calling a var within a var when you're inside of a function will not go through `evaluate`, it will go through codegen. We're talking about the call, not the var or the function within the var.

compiler+runtime/src/cpp/jank/runtime/behavior/callable.cpp
```diff
@@ -44,7 +44,7 @@ namespace jank::runtime
 
   object_ref dynamic_call(object_ref source, object_ref const a1)
   {
-    if(source->type == object_type::var)
+    while(source->type == object_type::var)
```
jeaye: There are at least ten different `dynamic_call` overloads. As I said in Slack, we should add a static helper to this file so we can do `source = pass_through_vars(source);`

Changing this one function is not enough.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/behavior/callable.cpp
stmonty: You're right, my mistake. I think I fixed it now.

compiler+runtime/src/cpp/jank/runtime/behavior/callable.cpp
```diff
@@ -10,13 +10,18 @@ namespace jank::runtime
...
     {
       source = runtime::deref(source);
     }
+    return source;
+  }
 
+  object_ref dynamic_call(object_ref source)
+  {
+    auto const processed_source(pass_through_vars(source));
```
jeaye, line 24: If you're not changing `source` for all of these anymore, then `source` can be `const`.

jeaye:
Thanks, Monty!

PR #437, opened 2025-08-30
Fix `:keys` destructuring for unqualified keywords

### LINKED ISSUE #293: `:keys` destructuring is broken for unqualified keywords
```clojure
; :keys is broken
clojure.core=> (let [{:keys [b] :as m} {:b :b}] [b m])
[nil {:b :b}]

; :strs and :syms are ok
clojure.core=> (let [{:strs [b] :as m} {"b" :b}] [b m])
[:b {"b" :b}]
clojure.core=> (let [{:syms [b] :as m} {'b :b}] [b m])
[:b {b :b}]
```

The problem seems to be the keyword `:b` is being converted to `(clojure.core/get clojure_core-map-4419 :nil/b)`:

```clojure
clojure.core=> (macroexpand-1 '(let [{:keys [b] :as m} {:b :b}] [b m]))
(let* [clojure_core-map-4419 {:b :b} clojure_core-map-4419 (if (clojure.core/seq? clojure_core-map-4419) (if (clojure.core/next clojure_core-map-4419) (clojure.core/into {} clojure_core-map-4419) (if (clojure.core/seq clojure_core-map-4419) (clojure.core/first clojure_core-map-4419) {})) clojure_core-map-4419) m clojure_core-map-4419 b (clojure.core/get clojure_core-map-4419 :nil/b)] [b m])
```

djblue:
Fixes #293

With the examples from the issue:

```clojure
% jank repl
user=> (let [{:keys [b] :as m} {:b :b}] [b m])
[:b {:b :b}]

user=> (let [{:strs [b] :as m} {"b" :b}] [b m])
[:b {"b" :b}]

user=> (let [{:syms [b] :as m} {'b :b}] [b m])
[:b {b :b}]

user=> (macroexpand-1 '(let [{:keys [b] :as m} {:b :b}] [b m]))
(let* [map-59 {:b :b} map-59 (if (clojure.core/seq? map-59) (if (clojure.core/next map-59) (clojure.core/into {} map-59) (if (clojure.core/seq map-59) (clojure.core/first map-59) {})) map-59) m map-59 b (clojure.core/get map-59 :b)] [b m])
```


djblue:
Although, maybe the fix should be to update `keyword` to ignore `nil` namespaces:

```clojure
% clj                                                 
Clojure 1.12.2
user=> (keyword nil "hi")
:hi
```

```clojure
% jank repl
user=> (keyword nil "hi")
:nil/hi
```

jeaye:
> Although, maybe the fix should be to update `keyword` to ignore `nil` namespaces:
> 
> ```clojure
> % clj                                                 
> Clojure 1.12.2
> user=> (keyword nil "hi")
> :hi
> ```
> 
> ```clojure
> % jank repl
> user=> (keyword nil "hi")
> :nil/hi
> ```

Yeah, I think that makes more sense. The code for destructuring is taken right from Clojure. We shouldn't need to change it. If `keyword` is the problem, let's fix that.

djblue:
@jeaye should this be fixed in jank's clojure.core or in the cpp runtime implementation of keyword?

jeaye:
> @jeaye should this be fixed in jank's clojure.core or in the cpp runtime implementation of keyword?

jank's `clojure.core/keyword` just calls into `clojure.core-native/keyword`. That'll ultimately call into `runtime::context::intern_keyword`.


```clojure
(defn keyword
  "Returns a Keyword with the given namespace and name.  Do not use :
   in the keyword strings, it will be added automatically."
  ([name]
   (if (keyword? name)
     name
     (clojure.core-native/keyword "" name)))
  ([ns name]
   (clojure.core-native/keyword ns name)))
```

We should fix this at is lowest point, I think.

djblue:
I think a problem might be that `clojure.core-native/keyword` turns the value `nil` into `"nil"` via `runtime::to_string(ns)`.

Which introduces another incompatibility:

```clojure
% clj     
Clojure 1.12.2
user=> (keyword 1 "hi")
Execution error (ClassCastException) at user/eval1 (REPL:1).
class java.lang.Long cannot be cast to class java.lang.String (java.lang.Long and java.lang.String are in module java.base of loader 'bootstrap')

% jank repl
user=> (keyword 1 "hi")
:1/hi
```

jeaye:
> I think a problem might be that `clojure.core-native/keyword` turns the value `nil` into `"nil"` via `runtime::to_string(ns)`.
> 
> Which introduces another incompatibility:
> 
> ```clojure
> % clj     
> Clojure 1.12.2
> user=> (keyword 1 "hi")
> Execution error (ClassCastException) at user/eval1 (REPL:1).
> class java.lang.Long cannot be cast to class java.lang.String (java.lang.Long and java.lang.String are in module java.base of loader 'bootstrap')
> 
> % jank repl
> user=> (keyword 1 "hi")
> :1/hi
> ```

Agreed. There's our problem! We'd hit this one we're able to run the clojure-test-suite as well, I'm sure. Let's update that function to check for `nil` and also non-strings.

djblue:
So it seems the implementation of `clojure.core/keyword` has a few more quirks than I expected.

```clojure
Clojure 1.12.2
user=> (keyword [])
nil
user=> (keyword #{})
nil
user=> (keyword {})
nil
user=> (keyword nil [])
Execution error (ClassCastException) at user/eval7 (REPL:1).
class clojure.lang.PersistentVector cannot be cast to class java.lang.String (clojure.lang.PersistentVector is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
user=> (keyword nil #{})
Execution error (ClassCastException) at user/eval9 (REPL:1).
class clojure.lang.PersistentHashSet cannot be cast to class java.lang.String (clojure.lang.PersistentHashSet is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
user=> (keyword nil {})
Execution error (ClassCastException) at user/eval11 (REPL:1).
class clojure.lang.PersistentArrayMap cannot be cast to class java.lang.String (clojure.lang.PersistentArrayMap is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
user=> 
```

The single arity version of the fn completely ignores non `keyword`, `symbol` and `string` arguments, returning nil. The two arity version expects only `nil` or `string` for the namespace and only `string` for the name, throwing exceptions otherwise.

I think I've captured all the subtle behaviors in my latest commit.

djblue:
There is one more case that I just noticed:

```clojure
% clj 
Clojure 1.12.2
user=> (keyword "" "hi")
:/hi

% jank repl
user=> (keyword "" "hi")
:hi
```

And for that, I think I need to dig into `intern_keyword`.

djblue:
After looking into jank's internal implementation of keywords (which are actually just symbols), there doesn't appear to be a way to represent a namespace as `""` as the zero value for `jtl::immutable_string` is `""`. This edge case might not be worth addressing as I've never seen it used in practice, but might be worth noting somewhere.

jeaye:
> There is one more case that I just noticed:
> 
> ```clojure
> % clj 
> Clojure 1.12.2
> user=> (keyword "" "hi")
> :/hi
> 
> % jank repl
> user=> (keyword "" "hi")
> :hi
> ```
> 
> And for that, I think I need to dig into `intern_keyword`.

Very interesting. I think we can expand our coverage of `keyword` in the clojure-test-suite, based on your findings here. Is that something you're open to? I'd want jank to follow Clojure's `keyword` exactly, but we don't need to worry about this particular issue yet until it's a failing test case for us in the clojure-test-suite.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -279,6 +279,24 @@ namespace jank::runtime
 
   object_ref keyword(object_ref const ns, object_ref const name)
   {
+    if(ns != jank_nil && ns->type != object_type::persistent_string)
+    {
+      throw std::runtime_error{ util::format(
+        "The 'keyword' function expects a namespace to be 'nil' or a 'string', got {} instead.",
+        runtime::to_string(ns)) };
```
jeaye: Let's always use `to_code_string` for printing out values. That way strings get their quotes, characters get their backslash, etc. We don't want it for the actual call to `intern_keyword`, though. Just for the errors.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -279,6 +279,24 @@ namespace jank::runtime
 
   object_ref keyword(object_ref const ns, object_ref const name)
   {
+    if(ns != jank_nil && ns->type != object_type::persistent_string)
```
jeaye: `object_ref` has an `is_nil` and `is_some`, which is preferable to the `!=` and `==` used in this function.

djblue:
> > There is one more case that I just noticed:
> > ```clojure
> > % clj 
> > Clojure 1.12.2
> > user=> (keyword "" "hi")
> > :/hi
> > 
> > % jank repl
> > user=> (keyword "" "hi")
> > :hi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And for that, I think I need to dig into `intern_keyword`.
> 
> Very interesting. I think we can expand our coverage of `keyword` in the clojure-test-suite, based on your findings here. Is that something you're open to? I'd want jank to follow Clojure's `keyword` exactly, but we don't need to worry about this particular issue yet until it's a failing test case for us in the clojure-test-suite.

I'm down to add some more examples to flesh out keywords in the clojure-test-suite 💯

djblue:
Is https://github.com/jank-lang/clojure-test-suite/pull/95 what you were thinking?

jeaye:
> Is [jank-lang/clojure-test-suite#95](https://github.com/jank-lang/clojure-test-suite/pull/95) what you were thinking?

Yeah, that should do it. jank will no doubt fail those, as is. :joy:

jeaye:
Nice work, Chris! :muscle:

PR #442, opened 2025-09-01
Add `#inst` support

djblue:
Still working through what string formats need to be support but wanted to put this up to make sure I was heading in the right direction.

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -883,7 +883,7 @@ namespace jank::read::parse
       }
       catch(jank::runtime::object * const e)
       {
-        return error::parse_invalid_inst(expect_object<obj::persistent_string>(e)->data,
+        return error::parse_invalid_inst(try_object<obj::persistent_string>(e)->data,
                                          { start_token.start, latest_token.end });
       }
     }


compiler+runtime/include/cpp/jank/runtime/obj/inst.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <chrono>
```
djblue, line 3: Not sure how to forward declare the types I use from this header, similar to how was done for [`uuid`](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/include/cpp/jank/runtime/obj/uuid.hpp#L5-L8).

compiler+runtime/include/cpp/jank/runtime/obj/inst.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <jank/runtime/object.hpp>
+
+namespace jank::runtime::obj
+{
+  using inst_ref = oref<struct inst>;
+  using inst_time_point = std::chrono::time_point<std::chrono::system_clock>;
```
djblue, line 10: This is my best guess as to what underlying cpp type should be used to represent an instant.

compiler+runtime/src/cpp/jank/runtime/obj/inst.cpp
```diff
@@ -0,0 +1,84 @@
...
+    : value{ now() }
+  {
+  }
+
+  static jtl::ref<inst_time_point> from_string(jtl::immutable_string const &s)
+  {
+    inst_time_point o;
+    std::istringstream is{ static_cast<std::string>(s) };
+    is.imbue(std::locale("en_US.utf-8"));
+    is >> std::chrono::parse("%FT%T", o);
```
djblue: I'm going to update this bit to loop through various supported formats in hopes of finding one that succeeds.

compiler+runtime/include/cpp/jank/runtime/obj/inst.hpp
jeaye: I think we're already using `chrono` in our PCH. Don't worry about it for now.

compiler+runtime/include/cpp/jank/runtime/obj/inst.hpp
jeaye: Yeah, looks fine.

compiler+runtime/src/cpp/jank/runtime/obj/inst.cpp
jeaye: Does Clojure use different `#inst` formats?

djblue: I think for printing, it's one format but for parsing, it supports many. The following are all valid:
```clojure
#inst "2025-09-01T19:10:28.691-00:00"
#inst "2025-09-01T19:10:28.691Z"
#inst "2025-09-01T19:10:28.691"
#inst "2025-09-01T19:10:28"
#inst "2025-09-01"
```

djblue:
@jeaye Ok, I think the PR is ready for review.

compiler+runtime/include/cpp/jank/runtime/obj/inst.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    /* behavior::object_like */
+    bool equal(object const &) const;
+    jtl::immutable_string to_string() const;
+    void to_string(jtl::string_builder &buff) const;
+    jtl::immutable_string to_code_string() const;
+    uhash to_hash() const;
+
+    object base{ obj_type };
+    jtl::ref<inst_time_point> value;
```
jeaye: No need for the `ref` here. This was used with the uuid in order for the forward declaration to happen (if you hold a pointer to some `T`, you don't need to know the size of the `T`, since all pointers are the same size, so `T` can be forward declared). Since we're not forward declaring, just hold the actual `inst_time_point` as a value.

compiler+runtime/include/cpp/jank/runtime/obj/inst.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+namespace jank::runtime::obj
+{
+  using inst_ref = oref<struct inst>;
+  using inst_time_point = std::chrono::time_point<std::chrono::system_clock>;
+
+  struct inst : gc
+  {
+    static constexpr object_type obj_type{ object_type::inst };
+    static constexpr bool pointer_free{ false };
```
jeaye: This can be true, since this type doesn't hold any pointers to GC memory.

compiler+runtime/src/cpp/jank/runtime/obj/inst.cpp
```diff
@@ -0,0 +1,102 @@
+#include <chrono>
```
jeaye: No need to include things which are already in the `inst.hpp` header which we're including.

compiler+runtime/src/cpp/jank/runtime/obj/inst.cpp
```diff
@@ -0,0 +1,102 @@
...
+  inst::inst()
+    : value{ now() }
+  {
+  }
+
+  static jtl::ref<inst_time_point> inst_from_string(jtl::immutable_string const &s)
+  {
+    inst_time_point o;
+    auto const formats = { "%FT%T%Oz", "%FT%T%z", "%FT%T", "%F" };
+    auto parsed = false;
```
jeaye: Please always use direct initialization. i.e. no ` = `.

Also, `formats` can be `static`, since we don't need to allocate space for it every call.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -849,6 +849,37 @@ namespace jank::read::parse
...
+
+      if(str_end.kind != lex::token_kind::string && str_end.kind != lex::token_kind::escaped_string)
+      {
+        return error::parse_invalid_reader_symbolic_value(
+          "The form after '#inst' must be a string literal.",
+          { start_token.start, latest_token.end });
+      }
+
+      auto const str(expect_object<obj::persistent_string>(str_result.expect_ok().unwrap().ptr));
+      auto const wrapped(make_box<obj::inst>(str->data));
```
jeaye: So, this will throw if the format is invalid, but that would just throw a normal exception with no source information. Thus, we won't get a nice compiler error output pointing at the string and saying the error. To do that, we'll need to catch that exception and then turn it into a compiler error.

We could get away with using `parse_invalid_reader_symbolic_value` again, but preferably we add a new `parse_invalid_inst` error kind.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/obj/inst.cpp
djblue: So for bools, direct initialization is `bool parsed{ false };`, but how do I direct initialize formats?

jeaye: Same thing for both of them. You're already using type inference, but you're putting a ` = ` which does nothing. Just remove the ` = ` from your original code and use `{}`.

```cpp
auto const formats{ "%FT%T%Oz", "%FT%T%z", "%FT%T", "%F" };
auto parsed{ false };

/* Or with an explicit type. I generally do this one. */
bool parsed{};
```

jeaye: When `auto` gets multiple types in the initializer, the type becomes `std::initializer_list<T>`, where `T` is calculated based on the types of the values provided.

djblue: I think the different length strings is messing with auto:

```
inst.cpp:19:44: error: initializer for variable 'formats' with type 'const auto' contains multiple expressions
   19 |     static auto const formats{ "%FT%T%Oz", "%FT%T%z", "%FT%T", "%F" };
```
however, the following works:

```cpp
    static std::initializer_list<char const *> const formats{ "%FT%T%Oz",
                                                              "%FT%T%z",
                                                              "%FT%T",
                                                              "%F" };
```

is this suitable?

compiler+runtime/src/cpp/jank/read/parse.cpp
djblue: How would you feel about adding `parse_invalid_tagged_reader_value`? Then we can update `#cpp` and `#uuid` to this error instead?

compiler+runtime/src/cpp/jank/runtime/obj/inst.cpp
jeaye: My fault. I forgot this rule [changed](https://stackoverflow.com/questions/51166807/initializer-list-with-auto-contains-multiple-expressions) in the standard. I'd probably just do this. The compiler can figure out the template arg itself.

```cpp
static std::vector const formats{ "%FT%T%Oz", "%FT%T%z", "%FT%T", "%F" };
```

compiler+runtime/src/cpp/jank/read/parse.cpp
jeaye: I would prefer that each of these reader tags has its own error. The reason is that these errors will relate to documentation, for search engine optimization. Each error id gets its own standalone web page, which should be the first search result for that page. In that page is a description of the error and common workarounds and fixes.

By using a super general error, we can't provide much useful info, or we'll have to provide way too much info, since the error could be related to `#cpp`, `#inst`, `#uuid`, etc. I think it's better to have a dedicated error, and thus a dedicated web page, for each one.

jeaye: Adding a new error involves:

1. Adding it to the enum in `error.hpp`
2. Adding it to the `kind_str` fn in `error.hpp`
3. Adding it to the `kind_to_message` in `error.cpp`
4. Adding a function with the same name to `error/analyze.hpp`
5. Implementing that function in `error/analyze.cpp` (they all just call `make_error`)

djblue: Since this error is produced by the parser, should it be a parser error or should I include `#include <jank/error/analyze.hpp>` in the parser?

jeaye: `s/analyze/parse/g` in my last message.

djblue: Ok, cool. That's what I stumbled upon initially. Also, once this PR gets merged, I'll update `#uuid` and `#cpp` to have their own custom error types 👍 I really like the motivation of optimizing error messages around search 😂

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -849,6 +849,45 @@ namespace jank::read::parse
...
+      auto const str(expect_object<obj::persistent_string>(str_result.expect_ok().unwrap().ptr));
+
+      try
+      {
+        auto const wrapped(make_box<obj::inst>(str->data));
+        return object_source_info{ wrapped, start_token, str_end };
+      }
+      catch(jank::runtime::object * const e)
+      {
+        return error::parse_invalid_inst(expect_object<obj::persistent_string>(e)->data,
```
jeaye: Let's use `try_object` here instead of `expect_object`. We should consider `expect_object` as *dangerous*, because it is. If anything else is thrown and we `expect_object` the wrong type, we're into undefined behavior land.

In this case, what if some other code involved in creating an `inst` throws, such as something within `jtl::immutable_string`, or something in the future which we don't have now? Better to play it safe.

jeaye (approved):
Thanks, Chris! I had one final comment which I've fixed locally and pushed to your branch.

PR #444, opened 2025-09-01
Basic `slurp` / `spit` implementation

djblue:
Not sure if this is how we want to implement `slurp` / `spit` but the idea of using cpp interop in `clojure.core` feels right 😂

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -6737,24 +6737,28 @@ fails, attempts to require sym's namespace and retries."
       [:encoding (first opts)])
     opts))
 
+(cpp/raw "#include <fstream>")
+(cpp/raw "#include <iostream>")
+
 (defn slurp
   "Opens a reader on f and reads all its contents, returning a string.
   See clojure.java.io/reader for a complete list of supported arguments."
   ([f & opts]
-    ;;  (let [opts (normalize-slurp-opts opts)
-    ;;        sw (java.io.StringWriter.)]
-    ;;    (with-open [^java.io.Reader r (apply jio/reader f opts)]
-    ;;      (jio/copy r sw)
-    ;;      (.toString sw)))
-   (throw "TODO: port slurp")))
+   (let [opts (normalize-slurp-opts opts)
+         file (cpp/std.ifstream. (cpp/cast cpp/std.string f))
+         buffer (cpp/std.stringstream.)]
+     (cpp/<< buffer (cpp/.rdbuf file))
+     (cpp/.str buffer))))
 
 (defn spit
   "Opposite of slurp.  Opens f with writer, writes content, then
   closes f. Options passed to clojure.java.io/writer."
   [f content & options]
-  ;; (with-open [^java.io.Writer w (apply jio/writer f options)]
-  ;;   (.write w (str content)))
-  (throw "TODO: port spit"))
+  (let [file (cpp/std.ofstream.)]
+    (cpp/.open file (cpp/cast cpp/std.string f))
+    (cpp/<< file (cpp/cast cpp/std.string content))
+    (cpp/.close file)
+    nil))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; futures (needs proxy);;;;;;;;;;;;;;;;;;
 (defn future-call


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -6737,24 +6737,28 @@ fails, attempts to require sym's namespace and retries."
...
 
 (defn spit
   "Opposite of slurp.  Opens f with writer, writes content, then
   closes f. Options passed to clojure.java.io/writer."
   [f content & options]
-  ;; (with-open [^java.io.Writer w (apply jio/writer f options)]
-  ;;   (.write w (str content)))
-  (throw "TODO: port spit"))
+  (let [file (cpp/std.ofstream.)]
+    (cpp/.open file (cpp/cast cpp/std.string f))
```
djblue: I tried using try/catch here but got an exception when compiling.

jeaye: Yep, there's an issue I've found around accessing a native local through a `try`. I have a failing test for it locally already. Although, we should count on the destructor of the stream to close things. We don't need to worry about it ourselves. :smile:

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -6737,24 +6737,28 @@ fails, attempts to require sym's namespace and retries."
...
   See clojure.java.io/reader for a complete list of supported arguments."
   ([f & opts]
-    ;;  (let [opts (normalize-slurp-opts opts)
-    ;;        sw (java.io.StringWriter.)]
-    ;;    (with-open [^java.io.Reader r (apply jio/reader f opts)]
-    ;;      (jio/copy r sw)
-    ;;      (.toString sw)))
-   (throw "TODO: port slurp")))
+   (let [opts (normalize-slurp-opts opts)
+         file (cpp/std.ifstream. (cpp/cast cpp/std.string f))
```
jeaye: Let's cast to a `char const*` so we can use the second ctor here: https://en.cppreference.com/w/cpp/io/basic_ifstream/basic_ifstream.html

That will save us from allocating a `std::string` (and doing a deep string copy), since casting to a C string actually just pulls out the underlying string data from the jank string.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -6737,24 +6737,28 @@ fails, attempts to require sym's namespace and retries."
       [:encoding (first opts)])
     opts))
 
+(cpp/raw "#include <fstream>")
+(cpp/raw "#include <iostream>")
+
```
jeaye: Let's move these to the top of the file, just like in C++. That way we don't end up with includes spread all throughout the file in various places.

Also `<fstream>` and `<sstream>` should be all we need for these, I think.

djblue:
I can't seem to replicate this failure locally 🤔

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -6741,20 +6744,23 @@ fails, attempts to require sym's namespace and retries."
...
-    ;;    (with-open [^java.io.Reader r (apply jio/reader f opts)]
-    ;;      (jio/copy r sw)
-    ;;      (.toString sw)))
-   (throw "TODO: port slurp")))
+   (let [opts (normalize-slurp-opts opts)
+         file (cpp/std.ifstream. (cpp/cast (cpp/type "char const*") f))
+         buffer (cpp/std.stringstream.)]
+     (when-not (cpp/.is_open file)
+       (throw (str f " (No such file or directory)")))
+     (cpp/<< buffer (cpp/.rdbuf file))
```
kylc: BTW you can avoid reallocation in the stringstream by measuring the file size ahead of time, allocating a fixed size string to serve as a buffer, and then reading into it. See `load_string_file` in Boost: https://www.boost.org/doc/libs/1_74_0/boost/filesystem/string_file.hpp

jeaye:
> I can't seem to replicate this failure locally 🤔

Hm, ok. This is an interop failure. I'm still smoothing things out in that area, so don't worry about it for now. By the time we get there, `clojure.core` has been compiled, so this change should be considered fine.

compiler+runtime/src/jank/clojure/core.jank
jeaye: Thanks, Kyle!

@djblue I'm cool if you want to tackle this now. If not, let's put a TODO in there so we can remember it for when we're focusing on perf.

djblue: I tried porting the code from boost via interop and struggled to figure out how to express `&str[0]`. Anything I tried wouldn't match the signature for `ifstream.read`. I could use a _pointer_ if it's easy enough and I'm just missing something. I also pushed up the todo if we want to call this done for now.

kylc: `&s.front()`/`(cpp/& (cpp/.front s))` is the same thing but doesn't require any array interop.

jeaye: `&str[0]` should also be achievable via `(cpp/& (cpp/* str))`.

djblue: `(cpp/& (cpp/* str))` didn't seem to work, but `(cpp/& (cpp/.front s))` did the trick, thanks for the help!

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -6741,20 +6745,25 @@ fails, attempts to require sym's namespace and retries."
...
-    ;;      (jio/copy r sw)
-    ;;      (.toString sw)))
-   (throw "TODO: port slurp")))
+   (let [opts (normalize-slurp-opts opts)
+         fname (cpp/cast (cpp/type "char const*") f)
+         file (cpp/std.ifstream. fname)]
+     (if-not (cpp/.is_open file)
+       (throw (str f " (No such file or directory)"))
+       (let [size (cpp/std.filesystem.file_size fname)
+             out (cpp/std.string. size (cpp/cast (cpp/type "char") 0))]
```
jeaye: Should just be able to do `(cpp/char)` to get a zeroed char.

jeaye: `cpp/type` and `cpp/value` are expensive operations, since they involve the whole C++ parser, analyzer, and codegen. We can get through `(cpp/char)` much more quickly.

jeaye:
Ignoring the unrelated CI failures, is this good to merge?

djblue:
@jeaye I think so, but should we wait as this would start causing other PRs to start failing?

jeaye:
> @jeaye I think so, but should we wait as this would start causing other PRs to start failing?

The failures in CI are unrelated to your changes, for all that I can tell. We can wait to see if we observe them on another PR to know for sure.

jeaye:
Good move on waiting to merge. I found an issue which was related to C++ interop being used in AOT compiled modules which are then loaded at runtime. I've fixed that up and also merged my initial porting of around 15% of `clojure.core` to use seamless interop instead of `clojure.core-native`.

If this sounds like something you might want to carry forward, to port more bits to use interop, let me know!

PR #445, opened 2025-09-02
Add nix package declaration

kylc:
What works:
```
# from within local jank checkout
nix build .?submodules=1
nix run .?submodules=1 check-health
nix run .?submodules=1 repl
# etc.
```

What doesn't work yet:
- Building without the submodules flag: nix will not copy any of the third-party sources into the build environment, so the build will fail.
- Building via a remote source like `nix build github:jank-lang/jank`: when pulling from GitHub, nix will download a .tar.gz which will not contain submodule sources. It seems like there is still active discussions around how to handle this within the nix community.
- Stack traces are probably somewhat limited in this build, see TODO comment in the code.

For future investigation:
- Deploy a binary cache via cachix so users on standard platforms don't need to build manually
- Run nix build step in CI (and push to cache?)
- Submit to nixpkgs, perhaps on the first official release

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,6 +15,41 @@
       }: {
         legacyPackages = pkgs;
         formatter = pkgs.alejandra;
+
+        packages.default = pkgs.stdenv.mkDerivation {
+          pname = "jank";
+          version = "git";
+          src = ./.;
+
+          nativeBuildInputs = with pkgs; [
+            git
+            cmake
+            ninja
+            (pkgs.callPackage ./llvm.nix {})
+          ];
+          buildInputs = with pkgs; [libzip openssl];
+
+          postPatch = ''
+            patchShebangs ./compiler+runtime/bin/ar-merge
+          '';
+
+          cmakeBuildDir = "./compiler+runtime/build";
+          cmakeDir = "..";
+          cmakeFlags = [
+            "-DCMAKE_C_COMPILER=clang"
+            "-DCMAKE_CXX_COMPILER=clang++"
+            # TODO: Updating RPATHs during install causes the step to fail as it
+            # tries to rewrite non-existent RPATHs like /lib. Needs more
+            # investigation.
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -15,6 +15,41 @@
...
+          ];
+          buildInputs = with pkgs; [libzip openssl];
+
+          postPatch = ''
+            patchShebangs ./compiler+runtime/bin/ar-merge
+          '';
+
+          cmakeBuildDir = "./compiler+runtime/build";
+          cmakeDir = "..";
+          cmakeFlags = [
```
jeaye, line 38: I think we'll want these enabled:

* `-DCMAKE_BUILD_TYPE=Release` (enables production optimizations)
* `-Djank_unity_build=on` (speeds up compilation with unity/jumbo builds)

jeaye:
This looks great, Kyle! It's a very minimal change, too. Are you looking to merge this as is and leave the other items for later, or keep this open while you iterate? Either's cool with me.

flake.nix
kylc: The build type is [handled by nix](https://github.com/NixOS/nixpkgs/blob/7d45281e45fe6c2263c120494548a9781f4245e8/pkgs/by-name/cm/cmake/setup-hook.sh#L109) (Release is default). Enabled the unity build in the latest commit.

kylc:
I think it can go in now, although at least the libdwarf/stack trace TODO should be resolved before this becomes the recommended way to use jank on Nix.

jeaye:
> I think it can go in now, although at least the libdwarf/stack trace TODO should be resolved before this becomes the recommended way to use jank on Nix.

Ok, perfect. I'll merge this now!

For libdwarf, we could add it as a submodule in our fork of cpptrace, rather than use `FetchContent`. That might be the easiest way to get it building.

PR #452, opened 2025-09-03
Add regex support

djblue:
I mainly tested the code via examples found at:

- https://clojuredocs.org/clojure.core/re-find
- https://clojuredocs.org/clojure.core/re-groups
- https://clojuredocs.org/clojure.core/re-matcher
- https://clojuredocs.org/clojure.core/re-matches

Also, since the underlying `std::regex` doesn't support (?...) option parsing, regexes like `#"(?i)hello"` will fail.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
@@ -104,8 +104,8 @@ namespace jank::runtime::obj
     }
   }
 
-  /* NOLINTNEXTLINE(cppcoreguidelines-noexcept-swap,bugprone-exception-escape) */
   object_ref
+  /* NOLINTNEXTLINE(cppcoreguidelines-noexcept-swap,bugprone-exception-escape) */
   atom::swap(object_ref const fn, object_ref const a1, object_ref const a2, object_ref const rest)
   {
     while(true)


compiler+runtime/src/cpp/jank/runtime/obj/regex.cpp
```diff
@@ -0,0 +1,97 @@
...
+    buff("#<re_matcher>");
+    return buff.release();
+  }
+
+  jtl::immutable_string re_matcher::to_code_string() const
+  {
+    return to_string();
+  }
+
+  uhash re_matcher::to_hash() const
```
djblue: This object doesn't really have value semantics, so I'm not sure what to put here.

compiler+runtime/src/cpp/jank/runtime/obj/regex.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  void re_matcher::to_string(jtl::string_builder &buff) const
+  {
+    buff("#<re_matcher>");
+  }
+
+  jtl::immutable_string re_matcher::to_string() const
+  {
+    jtl::string_builder buff;
+    buff("#<re_matcher>");
```
djblue: No sure what a good string representation for this object would be.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -5739,36 +5739,22 @@
...
-    ;;   (if (zero? gc)
-    ;;     (. m (group))
-    ;;     (loop [ret [] c 0]
-    ;;       (if (<= c gc)
-    ;;         (recur (conj ret (. m (group c))) (inc c))
-    ;;         ret))))
-  (throw "TODO: port re-groups"))
+  clojure.core-native/re-groups)
 
 (defn re-seq
```
djblue, line 5759: Skipping re-seq until https://github.com/jank-lang/jank/issues/457 is resolved.

compiler+runtime/src/cpp/jank/runtime/obj/regex.cpp
jeaye: Let's follow what we do for fns. https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/runtime/obj/jit_function.cpp#L52

compiler+runtime/src/cpp/jank/runtime/obj/regex.cpp
jeaye: This is what we do when we don't have a good representation: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/runtime/obj/transient_vector.cpp#L46

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -496,6 +496,144 @@ namespace jank::runtime
     return o->type == object_type::tagged_literal;
   }
 
+  object_ref re_pattern(object_ref o)
```
jeaye: `o` can be `const`.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -496,6 +496,144 @@ namespace jank::runtime
...
+  {
+    if(o->type == object_type::persistent_string)
+    {
+      return make_box<obj::re_pattern>(expect_object<obj::persistent_string>(o)->data);
+    }
+    else
+    {
+      throw std::runtime_error{ util::format("Expected string, got {}", object_type_str(o->type)) };
+    }
```
jeaye: Using `try_object` would allow you to remove this whole `if/else/throw` pattern. The same applies to the other functions.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -496,6 +496,144 @@ namespace jank::runtime
...
+
+  object_ref re_find(object_ref const m)
+  {
+    if(m->type != object_type::re_matcher)
+    {
+      throw std::runtime_error{ util::format("Expected matcher, got {}",
+                                             object_type_str(m->type)) };
+    }
+
+    auto const matcher = expect_object<obj::re_matcher>(m);
```
jeaye: Please always use direct initialization.

compiler+runtime/include/cpp/jank/runtime/obj/regex.hpp
```diff
@@ -0,0 +1,59 @@
...
+    bool equal(object const &) const;
+    jtl::immutable_string to_string() const;
+    void to_string(jtl::string_builder &buff) const;
+    jtl::immutable_string to_code_string() const;
+    uhash to_hash() const;
+
+    object base{ obj_type };
+
+    std::string s{};
+    std::smatch m{};
```
jeaye: Please give these meaningful names. We may use `s` or `m` for local variables, but that shouldn't be the case for members.

compiler+runtime/include/cpp/jank/runtime/obj/regex.hpp
```diff
@@ -0,0 +1,59 @@
...
+    /* behavior::object_like */
+    bool equal(object const &) const;
+    jtl::immutable_string to_string() const;
+    void to_string(jtl::string_builder &buff) const;
+    jtl::immutable_string to_code_string() const;
+    uhash to_hash() const;
+
+    object base{ obj_type };
+
+    std::string s{};
```
jeaye: Let's use a `jtl::immutable_string` here instead (and remove the `<string>` include). The `regex_search` function has overloads for C strings, so we can just call `c_str()` on our jtl string instead of keeping a `std::string`.

compiler+runtime/src/cpp/jank/runtime/obj/regex.cpp
```diff
@@ -0,0 +1,97 @@
...
+  {
+    if(hash)
+    {
+      return hash;
+    }
+
+    return hash = pattern.to_hash();
+  }
+
+  re_matcher::re_matcher(re_pattern_ref re, jtl::immutable_string const &s)
```
jeaye: `re` can be `const`.

compiler+runtime/include/cpp/jank/runtime/obj/regex.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <regex>
+#include <string>
+
+#include <jank/runtime/object.hpp>
+
+namespace jank::runtime::obj
+{
+  using re_pattern_ref = oref<struct re_pattern>;
+
+  struct re_pattern : gc
```
jeaye: I know it's more of a chore, but will you please split these out into their own `obj` headers? Right now, we only have one header which has multiple (`number.hpp`), which was one of the first one I ever did. At some point, that will be split up, too.

We should be able to file search `obj` and then the object type to find the exact file. This is just one of those sanity things.

jeaye: We need `gc_cleanup` instead of `gc` here, to run the dtors for our regex values. Same with `re_matcher`.

compiler+runtime/include/cpp/jank/runtime/obj/regex.hpp
```diff
@@ -0,0 +1,59 @@
...
+    void to_string(jtl::string_builder &buff) const;
+    jtl::immutable_string to_code_string() const;
+    uhash to_hash() const;
+
+    object base{ obj_type };
+
+    jtl::immutable_string pattern{};
+    std::regex regex{};
+
+    mutable uhash hash{};
```
jeaye: No need for the `hash` member on either of these objects, since they'll just return their address as a hash.

compiler+runtime/src/cpp/jank/runtime/obj/regex.cpp
djblue: How would you feel about going with `#object [type 0xaddress]`? This is similar to what Clojure/Script does.

I think the main advantage is that this is read-able edn if you use `tagged-literal` as the default handler for unknown tagged values.

jeaye: I'm not opposed. Perhaps you'll be up for updating the other instances of `{}@{}` we have for objects?

compiler+runtime/include/cpp/jank/runtime/obj/regex.hpp
djblue: When I tried to use a cstring, none of the overloads would match:

```
/home/chris/repos/jank/compiler+runtime/src/cpp/jank/runtime/core.cpp:564:5: error: no matching function for call to 'regex_search'
  564 |     std::regex_search(str,
      |     ^~~~~~~~~~~~~~~~~
/usr/include/c++/15.2.1/bits/regex.h:2440:5: note: candidate template ignored: deduced conflicting types for parameter '_Bi_iter' ('const_pointer_type' (aka 'const char *') vs. 'std::smatch' (aka 'match_results<__normal_iterator<const char *, std::basic_string<char, std::char_traits<char>, std::allocator<char>>>>'))
 2440 |     regex_search(_Bi_iter __s, _Bi_iter __e,
      |     ^
/usr/include/c++/15.2.1/bits/regex.h:2463:5: note: candidate template ignored: deduced conflicting types for parameter '_Bi_iter' ('const_pointer_type' (aka 'const char *') vs. 'std::smatch' (aka 'match_results<__normal_iterator<const char *, std::basic_string<char, std::char_traits<char>, std::allocator<char>>>>'))
 2463 |     regex_search(_Bi_iter __first, _Bi_iter __last,
      |     ^
/usr/include/c++/15.2.1/bits/regex.h:2486:5: note: candidate template ignored: could not match 'const type-parameter-0-0 *' against '__gnu_cxx::__normal_iterator<const char *, std::basic_string<char>>'
 2486 |     regex_search(const _Ch_type* __s,
      |     ^
/usr/include/c++/15.2.1/bits/regex.h:2547:5: note: candidate template ignored: could not match 'basic_string<_Ch_type, _Ch_traits, _Ch_alloc>' against 'const_pointer_type' (aka 'const char *')
 2547 |     regex_search(const basic_string<_Ch_type, _Ch_traits, _Ch_alloc>& __s,
      |     ^
/usr/include/c++/15.2.1/bits/regex.h:2562:5: note: candidate template ignored: could not match 'basic_string<_Ch_type, _Ch_traits, _Ch_alloc>' against 'const_pointer_type' (aka 'const char *')
 2562 |     regex_search(const basic_string<_Ch_type, _Ch_traits, _Ch_alloc>&&,
      |     ^
/usr/include/c++/15.2.1/bits/regex.h:2505:5: note: candidate function template not viable: requires at most 3 arguments, but 4 were provided
 2505 |     regex_search(const _Ch_type* __s,
      |     ^            ~~~~~~~~~~~~~~~~~~~~
 2506 |                  const basic_regex<_Ch_type, _Rx_traits>& __e,
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 2507 |                  regex_constants::match_flag_type __f
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 2508 |                  = regex_constants::match_default)
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/15.2.1/bits/regex.h:2524:5: note: candidate function template not viable: requires at most 3 arguments, but 4 were provided
 2524 |     regex_search(const basic_string<_Ch_type, _Ch_traits,
      |     ^            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 2525 |                  _String_allocator>& __s,
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~
 2526 |                  const basic_regex<_Ch_type, _Rx_traits>& __e,
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 2527 |                  regex_constants::match_flag_type __flags
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 2528 |                  = regex_constants::match_default)
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

compiler+runtime/include/cpp/jank/runtime/obj/regex.hpp
djblue: I think I did this correctly in https://github.com/jank-lang/jank/pull/452/commits/ff41a8f09fad760f3cc21e9137463dd1508dda37 but not sure.

compiler+runtime/include/cpp/jank/runtime/obj/regex.hpp
djblue: I realized we don't actually need to store `std::smatch` so I removed it and renames `s` to `match_input`.

djblue:
The CI failures appear to be for functions I didn't touch:

```
/home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:66:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
   66 |   object_ref atom::swap(object_ref const fn)
      |                    ^
/home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:80:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
   80 |   object_ref atom::swap(object_ref const fn, object_ref const a1)
      |                    ^
/home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:94:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
   94 |   object_ref atom::swap(object_ref const fn, object_ref const a1, object_ref const a2)
      |                    ^
/home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:109:9: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
  109 |   atom::swap(object_ref const fn, object_ref const a1, object_ref const a2, object_ref const rest)
      |         ^
```

Did I do something that would cause this transitively?

compiler+runtime/include/cpp/jank/runtime/obj/re_matcher.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <string>
+
+#include <jank/runtime/obj/re_pattern.hpp>
+#include <jank/runtime/object.hpp>
+
+namespace jank::runtime::obj
+{
+  using re_matcher_ref = oref<struct re_matcher>;
+
+  struct re_matcher : gc
```
jeaye: We need `gc_cleanup` here, due to the `std::string` member.

jeaye:
> The CI failures appear to be for functions I didn't touch:
> 
> ```
> /home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:66:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
>    66 |   object_ref atom::swap(object_ref const fn)
>       |                    ^
> /home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:80:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
>    80 |   object_ref atom::swap(object_ref const fn, object_ref const a1)
>       |                    ^
> /home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:94:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
>    94 |   object_ref atom::swap(object_ref const fn, object_ref const a1, object_ref const a2)
>       |                    ^
> /home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:109:9: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
>   109 |   atom::swap(object_ref const fn, object_ref const a1, object_ref const a2, object_ref const rest)
>       |         ^
> ```
> 
> Did I do something that would cause this transitively?

Hm, I'm not sure about this. Could be related to the regex object. Looks like main is green, and this is from static analysis, so I would expect this to be related to your changes.

We have those lines ignored for `cppcoreguidelines-noexcept-swap` but not for `bugprone-exception-escape`. Want to add the latter there? Should be able to just use `(cppcoreguidelines-noexcept-swap,bugprone-exception-escape)` I think.

djblue:
> > The CI failures appear to be for functions I didn't touch:
> > ```
> > /home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:66:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
> >    66 |   object_ref atom::swap(object_ref const fn)
> >       |                    ^
> > /home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:80:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
> >    80 |   object_ref atom::swap(object_ref const fn, object_ref const a1)
> >       |                    ^
> > /home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:94:20: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
> >    94 |   object_ref atom::swap(object_ref const fn, object_ref const a1, object_ref const a2)
> >       |                    ^
> > /home/runner/work/jank/jank/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp:109:9: error: an exception may be thrown in function 'swap' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
> >   109 |   atom::swap(object_ref const fn, object_ref const a1, object_ref const a2, object_ref const rest)
> >       |         ^
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Did I do something that would cause this transitively?
> 
> Hm, I'm not sure about this. Could be related to the regex object. Looks like main is green, and this is from static analysis, so I would expect this to be related to your changes.
> 
> We have those lines ignored for `cppcoreguidelines-noexcept-swap` but not for `bugprone-exception-escape`. Want to add the latter there? Should be able to just use `(cppcoreguidelines-noexcept-swap,bugprone-exception-escape)` I think.

That seems to have fixed 3/4 of the errors, which is even more odd 🤔

compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
```diff
@@ -104,7 +104,7 @@ namespace jank::runtime::obj
     }
   }
 
-  /* NOLINTNEXTLINE(cppcoreguidelines-noexcept-swap) */
+  /* NOLINTNEXTLINE(cppcoreguidelines-noexcept-swap,bugprone-exception-escape) */
```
jeaye: The next line is literally just `object_type`, which is why this doesn't work. If we swap this line and the next line, all will be well.

jeaye (approved):
Thanks, Chris! We'll see if CI is green. If so, I'll merge this in!

PR #453, opened 2025-09-03
Build custom libdwarf-lite in nix package

kylc:
Builds cpptrace with the same libdwarf/zstd versions which are pulled down in a non-sandboxed build via `FetchContent`. Fixes poor stracktrace support in nix package builds from https://github.com/jank-lang/jank/pull/445.

Large diff due to indentation.


jeaye:
Oh, nice. Even better than moving them to a submodule! Excellent work, Kyle.

PR #456, opened 2025-09-05
Add nix build job to CI

kylc:
WIP. This will require some Github/Cachix setup before merging. Example run here: https://github.com/kylc/jank/actions/runs/17479404939/job/49646646076

Some cool things:
- After the CI job finishes, users who have the exact same inputs (same checkout, not a dirty tree, etc.) can just run `nix run .?submodules=1 ...` and nix will automatically pull down the cached version, no building required (if they are using the cache, instructions via `cachix use kylc`)
- If you have write access to the cache, you can also build a clean local copy and push it into the cache, which will then be picked up by CI so the build job will be skipped.
- All of the LLVM caching is handled transparently. You can even build it locally and push it to the cache before pushing your jank changes to speed up the CI process.
- We can now trivially build docker images or even DEB/RPM packages with `nix bundle`, although the latter are less useful because they bring all of their own dependencies and install to `/nix/store/`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,3 +172,27 @@ jobs:
               git push origin main
             popd
           popd
+
+  build-nix:
+    name: "nix - release"
+    runs-on: ubuntu-latest
+    steps:
+    - name: "Set up nix"
+      uses: cachix/install-nix-action@v31
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - name: "Set up cachix"
+      uses: cachix/cachix-action@v16
+      with:
+        name: jank-lang
+        # If you chose API tokens for write access OR if you have a private cache
+        authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+        # Only push to cache on main branch, not PRs
+        skipPush: "${{ github.ref != 'refs/heads/main' }}"
+    - uses: actions/checkout@v4
+      with:
+        submodules: 'recursive'
+    - name: "Build jank"
+      run: nix build .?submodules=1 -L
+    - name: "Check health"
+      run: nix run .?submodules=1 check-health
--- a/flake.nix
+++ b/flake.nix
@@ -36,14 +36,20 @@
           pkgs.stdenv.mkDerivation (finalAttrs: {
             pname = "jank";
             version = "git";
-            src = ./.;
 
-            nativeBuildInputs = with pkgs; [
-              git
-              cmake
-              ninja
-              llvm-jank
-            ];
+            # Add only essential files so that the source hash is consistent.
+            src = lib.fileset.toSource {
+              root = ./.;
+              fileset =
+                lib.fileset.difference
+                (lib.fileset.unions [
+                  ./.clang-format
+                  ./compiler+runtime
+                ])
+                (lib.fileset.fileFilter (file: lib.strings.hasInfix ".git" file.name) ./.);
+            };
+
+            nativeBuildInputs = with pkgs; [git cmake ninja llvm-jank];
             buildInputs = with pkgs; [libzip openssl];
             checkInputs = with pkgs; [glibcLocales doctest];
 


.github/workflows/build-nix.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: cachix/install-nix-action@v31
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - uses: cachix/cachix-action@v16
+      with:
+        name: kylc
```
kylc: TODO: set up jank cachix

.github/workflows/build-nix.yml
```diff
@@ -0,0 +1,25 @@
...
+        # If you chose API tokens for write access OR if you have a private cache
+        authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
```
kylc: TODO: add Github repo secret

jeaye:
Incredible. I'm interested, just let me know what you need from me. :slightly_smiling_face:

kylc:
You will need to:
- Sign up for https://cachix.org and create a new cache (maybe `jank-lang`). It's free and the limits for open source projects are generous enough to store many past clang/jank builds.
- In the cache settings, generate a write+read auth token. Add this in the Github repo settings as a repo secret called `CACHIX_AUTH_TOKEN`.

Then I will update the commits to point to the new jank cache instead of my personal one. Later on we can update the jank install instructions to point users to the binary cache.

---

Do you think this action should run on all pull requests? That would ensure any changes don't break the nix build. Alternatively we can run it only on pushes to `main`. That will at least ensure the binary cache is up-to-date for end users.

jeaye:
Ok, I've created the `jank-lang` cache: `jank-lang.cachix.org-1:iLjYBD9b/v1D6FxUByF976x3BQ/AJGsQL1Rm49Sw7Fg=`

I've also created the `CACHIX_AUTH_TOKEN` secret in this repo.

> Do you think this action should run on all pull requests? That would ensure any changes don't break the nix buil

I think the ideal is to build for Nix on PRs but only store the cache for main. It'd be quite wasteful to store the PR builds in the cache, since we have no common workflow for people to build someone else's PR.

If that isn't feasible, I'll be ok caching the PR builds. I'd rather have Nix builds for PR CI than not have them at all. What do you think?

kylc:
- Updated to use jank-lang cachix
- Cache artifacts are not pushed from PR branches

I *think* this is ready to go now, although I couldn't get it to populate the cache even if I disabled the non-main branch check. I'm pretty sure this is just because forks don't have secret access to the main repo (could be dangerous), so it will pick up write access to the cache once it is merged :crossed_fingers:. PRs can still pull from the cache since no key is required for read access.

I pushed up a dummy PR to my own fork using my own cache to make sure, and it all looks good: https://github.com/kylc/jank/pull/1

There will be one followup to tie all the nix work together and document how to use the cache locally to install jank.

jeaye:
> * Updated to use jank-lang cachix
> 
>     * Cache artifacts are not pushed from PR branches
> 
> 
> I _think_ this is ready to go now, although I couldn't get it to populate the cache even if I disabled the non-main branch check. I'm pretty sure this is just because forks don't have secret access to the main repo (could be dangerous), so it will pick up write access to the cache once it is merged 🤞. PRs can still pull from the cache since no key is required for read access.
> 
> I pushed up a dummy PR to my own fork using my own cache to make sure, and it all looks good: [kylc#1](https://github.com/kylc/jank/pull/1)
> 
> There will be one followup to tie all the nix work together and document how to use the cache locally to install jank.

Nice! So, one question. For each new PR, is this going to rebuild Clang/LLVM in Nix? Or will it be able to use a cached Clang/LLVM and just build jank?

kylc:
It will pull Clang/LLVM from the cache. The only case that won't happen is if you make changes to LLVM and bump the rev in `llvm.nix`. Then it will build it once and upload the new Clang/LLVM artifact to the cache.

The reason it's rebuilding Clang/LLVM every time I push to this PR is because the cache has not been populated, due to what I mentioned above. Further, in normal operation, if you update only files which are not part of the jank file set, it can even skip building jank itself!

jeaye:
> It will pull Clang/LLVM from the cache. The only case that won't happen is if you make changes to LLVM and bump the rev in `llvm.nix`. Then it will build it once and upload the new Clang/LLVM artifact to the cache.
> 
> The reason it's rebuilding Clang/LLVM every time I push to this PR is because the cache has not been populated, due to what I mentioned above. Further, in normal operation, if you update only files which are not part of the jank file set, it can even skip building jank itself!


perfect

PR #464, opened 2025-09-06
Bump llvm-project rev in nix build

kylc:
Unfortunately we need to update this whenever we make LLVM changes :melting_face:. In the future it can be caught by the Nix CI build action.


jeaye:
Ah, good catch. With your latest fix, hopefully we're very close to just being able to use upstream LLVM 22. :crossed_fingers:

PR #465, opened 2025-09-07
fix: Variadic function missing parameter vector

### LINKED ISSUE #207: Crash parsing malformed `fn*`
```clojure
clojure.core=> (fn* a ())
Assertion failed: (set), function unwrap, file option.hpp, line 163.
zsh: abort      jank repl
```

stmonty:
Fixes #207 
I am not sure about the phrasing of the error message. Let me know what you think.
image


jeaye:
Thanks, Monty! The logic for spotting the error looks good, but the message itself is incorrect. This function is not variadic. A variadic function in Clojure `& args` to capture any number of arguments. In fact, since we don't have the parameter vector, we have no idea what arity this function is.

Maybe something simple like:

> This function is missing its parameter vector.

The text for the note looks great.

stmonty:
Oh you're right 🤦‍♂️ guess I am too tired. I meant multi-arity function instead of variadic. I can replace the error message to be more clear that it is meant to be a multi-arity function. But like you said, I think it is probably less confusing if we just make the message more generic.

PR #467, opened 2025-09-09
Update to_string to return edn

djblue:


jeaye:
I was just checking the EDN spec: https://github.com/edn-format/edn?tab=readme-ov-file#tagged-elements

Looks like it never says that `#object[...]` specifically is valid. I wonder because the `[...]` after `#object` would normally be parsed differently, right? Or is it actually a tag followed by a vector and we just render it without the space?

djblue:
It's a tag followed by a vector. A reader can interpret it in any way it wants, but I tend to do [this](https://clojuredocs.org/clojure.core/tagged-literal#example-63092f78e4b0b1e3652d7655) so I can parse any valid edn that also contains platform specific values. And yes, you are correct the space after the tag isn't required. `(= #inst"2025-09-09T01:10:55.796-00:00"  #inst "2025-09-09T01:10:55.796-00:00")`. Would you prefer a space?

I think Clojure doesn't include one which is probably why I did it that way:
```clojure
user=> (Object.)
#object[java.lang.Object 0x51c6e209 "java.lang.Object@51c6e209"]
user=> 
```

jeaye:
> It's a tag followed by a vector. A reader can interpret it in any way it wants, but I tend to do [this](https://clojuredocs.org/clojure.core/tagged-literal#example-63092f78e4b0b1e3652d7655) so I can parse any valid edn that also contains platform specific values. And yes, you are correct the space after the tag isn't required. `(= #inst"2025-09-09T01:10:55.796-00:00" #inst "2025-09-09T01:10:55.796-00:00")`. Would you prefer a space?

Ok, gotcha. Yeah, I think the space makes more sense and doesn't make it seem like something special. :sweat_smile:

djblue:
Updated to include space after tag 👍

PR #468, opened 2025-09-09
Update `re-*` fns to use cpp interop

djblue:


jeaye:
Nice! More and more interop. :grin:

PR #470, opened 2025-09-10
Only create `persistent_string` if `char const *` is not null

djblue:
Before:
```clojure
% jank repl
user=> (cpp/std.getenv (cpp/cast (cpp/type "char const*") "MISSING"))
Assertion failed! s != nullptr
```
After:
```clojure
% jank repl                                           
user=> (cpp/std.getenv (cpp/cast (cpp/type "char const*") "MISSING"))
nil

user=> (cpp/std.getenv (cpp/cast (cpp/type "char const*") "CXX"))
"/home/chris/repos/jank/compiler+runtime//build/llvm-install/usr/local/bin/clang++"
```


jeaye:
Makes sense to me!

PR #472, opened 2025-09-10
More cpp interop in `clojure.core`

djblue:


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -6403,13 +6403,15 @@ fails, attempts to require sym's namespace and retries."
...
+  [coll]
+  (cpp/jank.runtime.is_sorted coll))
 
-(def counted?
- "Returns true if coll implements count in constant time"
-  clojure.core-native/counted?)
+(defn counted?
+  "Returns true if coll implements count in constant time"
+  [coll]
+  (cpp/jank.runtime.is_counter coll))
```
jeaye: I'd expect this one to fail: `is_counter`

djblue: Why should this one fail? It worked at the REPL for me 🤔

```clojure
% compile && jank repl        
ninja: no work to do.
user=> (counted? (seq (list)))
false

user=> (counted? [])
true

user=> (counted? {})
true
```

jeaye: There's a typo in the name. `is_counter` should be `is_counted` (`Counted` is Clojure JVM nomenclature for `O(1)` count ability). But the typo is in `main`.

So, when I saw it I was like "No way this works."

djblue: Ohh, gotcha. Should I update the underlying cpp fn to have the correct name?

jeaye: Sure, if you're up for it. If not, I'll do it in my next PR.

jeaye (approved):
:100:

PR #473, opened 2025-09-10
Convert `clojure.string` to use cpp interop

djblue:


compiler+runtime/include/cpp/clojure/string_native.hpp
```diff
@@ -1,5 +1,20 @@
...
-#include <jank/c_api.h>
+#include <jank/runtime/object.hpp>
 
-extern "C" jank_object_ref jank_load_clojure_string_native();
+namespace clojure::string_native
+{
+  using namespace jank;
+  using namespace jank::runtime;
+
+  object_ref blank(object_ref s);
```
djblue, line 10: When I initially added these, I did so with `const`, which actually caused issues with the cpp interop. Pretty sure we don't need `const` here, so I removed them, but wanted to mention it as it might be the reason we are having issues with cpp interop in other places.

jeaye: Great catch. I made a local test for this and I'll fix it after the IR issue.

jeaye:
You're the man, Chris! :muscle:

PR #474, opened 2025-09-10
Add `clojure.string/replace-first`

djblue:
There is some duplication in this PR, but I wanted to get some feedback and focus on getting it correct before tidying up the code.

### DIFF
--- a/compiler+runtime/src/cpp/clojure/string_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/string_native.cpp
@@ -75,15 +75,18 @@ namespace clojure::string_native
       return make_box(s_str);
     }
 
-    jtl::string_builder buff;
+    auto const replacement_char(try_object<obj::character>(replacement)->data);
+
+    auto const s_size(s_str.size());
+
+    jtl::string_builder buff{ s_size };
     buff(s_str.substr(0, i));
 
-    auto const replacement_char(try_object<obj::character>(replacement)->data);
     buff(replacement_char);
 
     auto const rest_i(i + 1);
 
-    if(rest_i < s_str.size())
+    if(rest_i < s_size)
     {
       buff(s_str.substr(rest_i));
     }
@@ -104,15 +107,18 @@ namespace clojure::string_native
       return make_box(s_str);
     }
 
-    jtl::string_builder buff;
-    buff(s_str.substr(0, i));
-
     auto const replacement_str(try_object<obj::persistent_string>(replacement)->data);
+
+    auto const replacement_size(replacement_str.size());
+    auto const s_size(s_str.size());
+
+    jtl::string_builder buff{ s_size - match_str.size() + replacement_size };
+    buff(s_str.substr(0, i));
     buff(replacement_str);
 
-    auto const rest_i(i + replacement_str.size());
+    auto const rest_i(i + replacement_size);
 
-    if(rest_i < s_str.size())
+    if(rest_i < s_size)
     {
       buff(s_str.substr(rest_i));
     }
--- a/compiler+runtime/src/cpp/clojure/string_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/string_native.cpp
@@ -145,7 +145,7 @@ namespace clojure::string_native
     }
   }
 
-  object_ref
+  static object_ref
   replace_first_re_pattern(object_ref const s, object_ref const match, object_ref const replacement)
   {
     auto const s_str(runtime::to_string(s));
--- a/compiler+runtime/src/cpp/clojure/string_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/string_native.cpp
@@ -98,7 +98,8 @@ namespace clojure::string_native
 
   static object_ref match_results_collect(std::smatch const &match_results)
   {
-    switch(match_results.size())
+    auto const size(match_results.size());
+    switch(size)
     {
       case 0:
         return jank_nil;
@@ -109,10 +110,11 @@ namespace clojure::string_native
       default:
         {
           native_vector<object_ref> vec;
+          vec.reserve(size);
 
           for(auto const s : match_results)
           {
-            vec.push_back(make_box<obj::persistent_string>(s.str()));
+            vec.emplace_back(make_box<obj::persistent_string>(s.str()));
           }
 
           return make_box<obj::persistent_vector>(
--- a/compiler+runtime/src/cpp/clojure/string_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/string_native.cpp
@@ -96,11 +96,26 @@ namespace clojure::string_native
     return make_box(buff.release());
   }
 
+  static object_ref replace_first_re_pattern_string(object_ref const s,
+                                                    object_ref const match,
+                                                    object_ref const replacement)
+  {
+    auto const is_string(s->type == object_type::persistent_string);
+    auto const s_str(is_string ? try_object<obj::persistent_string>(s)->data.c_str()
+                               : runtime::to_string(s).c_str());
+    auto const match_regex(try_object<obj::re_pattern>(match)->regex);
+    auto const replacement_str(try_object<obj::persistent_string>(replacement)->data.c_str());
+    auto const out_str(std::regex_replace(s_str,
+                                          match_regex,
+                                          replacement_str,
+                                          std::regex_constants::format_first_only));
+    return make_box<obj::persistent_string>(out_str.c_str()).erase();
+  }
+
   static object_ref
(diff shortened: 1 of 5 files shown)


compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,162 @@ namespace clojure::string_native
...
+
+    if(rest_i < s_str.size())
+    {
+      buff(s_str.substr(rest_i));
+    }
+
+    return make_box(buff.release());
+  }
+
+  static object_ref match_results_collect(std::smatch const &match_results)
```
djblue: What is a good way to use this code which is copied from https://github.com/jank-lang/jank/blob/1c8cb9e6cd72e0c0a80d03c1ce3946e102ab058a/compiler%2Bruntime/src/cpp/jank/runtime/core.cpp#L610-L630?

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,162 @@ namespace clojure::string_native
...
+    }
+
+    return make_box(buff.release());
+  }
+
+  object_ref replace_first(object_ref const s, object_ref const match, object_ref const replacement)
+  {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wswitch-enum"
+    switch(match->type)
```
djblue, line 144: Not sure what should be used here: `if`, `switch` or `visit`.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,162 @@ namespace clojure::string_native
...
+    auto const match_char(expect_object<obj::character>(match)->data);
+    auto const i(s_str.find(match_char));
+
+    if(i == jtl::immutable_string::npos)
+    {
+      return make_box(s_str);
+    }
+
+    jtl::string_builder buff;
+    buff(s_str.substr(0, i));
```
djblue: I'm assuming this is a good way to build the new string.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,162 @@ namespace clojure::string_native
...
+  object_ref
+  replace_first_re_pattern(object_ref const s, object_ref const match, object_ref const replacement)
+  {
+    auto const s_str(runtime::to_string(s));
+
+    auto const match_regex(try_object<obj::re_pattern>(match)->regex);
+
+    std::smatch match_results{};
+    std::string const search_str{ s_str.c_str() };
+    std::regex_search(search_str, match_results, match_regex);
```
djblue: Not a fan of having this here.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,162 @@ namespace clojure::string_native
...
+
+    auto const i(match_results.position(0));
+
+    jtl::string_builder buff;
+    buff(s_str.substr(0, i));
+
+    if(replacement->type == object_type::persistent_string)
+    {
+      auto const replacement_str(try_object<obj::persistent_string>(replacement)->data);
+      buff(replacement_str);
```
djblue: Still need to work on supporting this variant:
```clojure
(clojure.string/replace-first "swap first two words" #"(\w+)(\s+)(\w+)" "$3$2$1")
;=> "first swap two words"
```

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,162 @@ namespace clojure::string_native
...
+
+    if(rest_i < s_str.size())
+    {
+      buff(s_str.substr(rest_i));
+    }
+
+    return make_box(buff.release());
+  }
+
+  object_ref replace_first(object_ref const s, object_ref const match, object_ref const replacement)
```
djblue: It might be useful to move some of the implementation for char/char and string/string replacement out of here and into `<jtl/immutable_string.hpp>`, but wasn't sure.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,168 @@ namespace clojure::string_native
     return make_box(util::to_uppercase(s_str));
   }
 
+  static object_ref
+  replace_first_character(object_ref const s, object_ref const match, object_ref const replacement)
+  {
+    auto const s_str(runtime::to_string(s));
+
+    auto const match_char(expect_object<obj::character>(match)->data);
```
jeaye: You can never use `expect_object` when you're receiving unknown input. Instead, you want `try_object`, which will throw if the object type is incorrect. With `expect_object`, you'll just get undefined behavior instead.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,168 @@ namespace clojure::string_native
...
+    auto const i(s_str.find(match_char));
```
jeaye: `match_char` is a `character_ref`, which is part of jank's runtime. `s_str` is a `immutable_string`, which is part of `jtl`. The `jtl` doesn't know about jank's runtime, but jank's runtime uses the `jtl`. So we'll need to search for the string within the character object.

Keep in mind that "characters" in Clojure are Unicode-aware, unlike `char` in C++ (which is always one byte). To support this, the `character` object in jank actually just stores an `immutable_string` for its data. So, what we're doing when we "replace the first character" in a string is actually replacing the first sub string.

This means that the only difference between `replace_first_character` and `replace_first_string` is the object they use. So, let's just have `replace_first_string` and have it take in an `immutable_string`. Then just call that with either the string data from a string object or the string data from a character object.

compiler+runtime/src/cpp/clojure/string_native.cpp
jeaye: Yeah, `string_builder` is a good way to do this efficiently, since it can release its internal string buffer to give to a new `immutable_string` without the need for a copy.

compiler+runtime/src/cpp/clojure/string_native.cpp
jeaye: You can add a helper to `core.hpp` which takes the match results by `const &` and returns an `object_ref`. Then just call that in both places. Also, consider calling `reserve` on the vector after creating it and using `emplace_back` for more efficient appending.

compiler+runtime/src/cpp/clojure/string_native.cpp
jeaye: I think it's fine.

compiler+runtime/src/cpp/clojure/string_native.cpp
jeaye: I addressed this in a comment above. character objects just use strings anyway, so we only have substring replacement.

compiler+runtime/src/cpp/clojure/string_native.cpp
jeaye: `switch` is fine here.

jeaye (changes requested):
Good questions, Chris. Hopefully I cleared things up.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,122 @@ namespace clojure::string_native
...
+      return s;
+    }
+
+    auto const i(match_results.position(0));
+
+    jtl::string_builder buff;
+    buff(s.substr(0, i));
+
+    auto const group(smatch_to_vector(match_results));
+    auto const replacement_value(dynamic_call(replacement, group));
```
djblue, line 125: I tried checking that replacement was callable, which wasn't working for collections as documented by https://github.com/jank-lang/jank/issues/487, but removed it as even Clojure doesn't check callable before invoking replacement with group.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -57,6 +62,122 @@ namespace clojure::string_native
...
+                                               runtime::to_code_string(match)) };
+    }
+#pragma clang diagnostic pop
+  }
+
+  object_ref replace_first(object_ref const s, object_ref const match, object_ref const replacement)
+  {
+    auto const is_string(s->type == object_type::persistent_string);
+    auto const s_str(is_string ? try_object<obj::persistent_string>(s)->data
+                               : runtime::to_string(s));
```
jeaye: You'll want `auto const &` to not copy the string data.

PR #484, opened 2025-09-12
Flatten declarive flags instead of string joining

christos-chatzifountas-biotz-io:
This simplifies the code a little bit and fixies the problem with of linked library arguments not passed correctly


jeaye:
Thank you!

PR #486, opened 2025-09-13
Add `re-seq`

djblue:
I tried adding direct interop calls to re-seq to avoid the extra indirection, but that generated runtime errors:

```clojure
(defn re-seq
  "Returns a lazy sequence of successive matches of pattern in string,
  using java.util.regex.Matcher.find(), each such match processed with
  re-groups."
  [re s]
  (let [m (cpp/jank.runtime.re_matcher re s)]
    ((fn step []
       (when (cpp/jank.runtime.re_find m)
         (cons (cpp/jank.runtime.re_groups m) (lazy-seq (step))))))))
```

But going through `clojure.core` fns worked just fine.

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -5791,17 +5791,6 @@
   [m]
   (cpp/jank.runtime.re_groups m))
 
-(defn re-seq
-  "Returns a lazy sequence of successive matches of pattern in string,
-  using java.util.regex.Matcher.find(), each such match processed with
-  re-groups."
-  [#_java.util.regex.Pattern re s]
-  ;; (let [m (re-matcher re s)]
-  ;;   ((fn step []
-  ;;      (when (. m (find))
-  ;;        (cons (re-groups m) (lazy-seq (step)))))))
-  (throw "TODO: port re-seq"))
-
 (defn re-matches
   "Returns the match, if any, of string to pattern, using
   java.util.regex.Matcher.matches().  Uses re-groups to return the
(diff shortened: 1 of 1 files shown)


jeaye:
The interop issue for this has been fixed here: https://github.com/jank-lang/jank/pull/490

So we should be able to use the C++ interop version now!

djblue:
The following version does work now:

```clojure
(defn re-seq [re s]
  (let [m (cpp/jank.runtime.re_matcher re s)]
    ((fn step []
       (when (re-find m)
         (cons (re-groups m) (lazy-seq (step))))))))
```

However, the following versions do compile but do not work at runtime:
```clojure
(defn re-seq [re s]
  (let [m (cpp/jank.runtime.re_matcher re s)]
    ((fn step []
       (when (re-find m)
         (cons (cpp/jank.runtime.re_groups m) (lazy-seq (step))))))))
```
```clojure
(defn re-seq [re s]
  (let [m (cpp/jank.runtime.re_matcher re s)]
    ((fn step []
       (when (cpp/jank.runtime.re_find m)
         (cons (re-groups m) (lazy-seq (step))))))))
```

In both version, the cpp interop call in `step`, results in `m` being the same following pointer value:
```cpp
(const jank::runtime::object_ref) {
  data = 0x0000000000000038
}
```

jeaye:
> The following version does work now:
> 
> ```clojure
> (defn re-seq [re s]
>   (let [m (cpp/jank.runtime.re_matcher re s)]
>     ((fn step []
>        (when (re-find m)
>          (cons (re-groups m) (lazy-seq (step))))))))
> ```
> 
> However, the following versions do compile but do not work at runtime:
> 
> ```clojure
> (defn re-seq [re s]
>   (let [m (cpp/jank.runtime.re_matcher re s)]
>     ((fn step []
>        (when (re-find m)
>          (cons (cpp/jank.runtime.re_groups m) (lazy-seq (step))))))))
> ```
> 
> ```clojure
> (defn re-seq [re s]
>   (let [m (cpp/jank.runtime.re_matcher re s)]
>     ((fn step []
>        (when (cpp/jank.runtime.re_find m)
>          (cons (re-groups m) (lazy-seq (step))))))))
> ```
> 
> In both version, the cpp interop call in `step`, results in `m` being the same following pointer value:
> 
> ```c++
> (const jank::runtime::object_ref) {
>   data = 0x0000000000000038
> }
> ```

Ok, great to know. I will tackle those issues next.

jeaye:
For now, we can merge this using whichever format we can currently support.

PR #493, opened 2025-09-14
Wrap up and document nix flake workflow

kylc:
- Add instructions for "installing"/running from remote nix flake
- Final changes to support `nix run git+https://...`
- Move some source definitions to the top-level flake.nix to make updating them easier to remember
- Run `nix fmt`, misc. cleanup

To test you can run directly from my PR branch. The cache will only be populated once this is merged, so until then it will build jank from source.
```
nix run git+https://github.com/kylc/jank.git?ref=nix-cleanup -- check-health
```

### DIFF
--- a/compiler+runtime/doc/install.md
+++ b/compiler+runtime/doc/install.md
@@ -66,6 +66,33 @@ yay -S jank-git   # if you installed jank-git
 yay -S jank-bin   # if you installed jank-bin
 ```
 
+## Nix
+### Install
+We have binary packages in our cachix cache, so installation is quick and easy.
+You may also skip the cache setup and artifacts will be built from source.
+
+```bash
+# Make sure cachix is installed, then follow the prompts.
+cachix use jank-lang
+```
+
+Next, you can directly run the jank binary from the nix flake derivation. For
+example, to run the `jank check-health` subcommand:
+
+```bash
+nix run git+https://github.com/jank-lang/jank.git -- check-health
+```
+
+If you pull down the jank repository locally for development, you can enter the
+dev shell and then follow the build instructions linked below. Skip the LLVM
+build step as it's done for you by nix.
+
+```bash
+git clone https://github.com/jank-lang/jank.git
+cd jank
+nix develop
+```
+
 ## Anything else
 If nothing above matches what you have, you can still build jank by following
 the docs [here](./build.md).
--- a/compiler+runtime/doc/install.md
+++ b/compiler+runtime/doc/install.md
@@ -67,30 +67,52 @@ yay -S jank-bin   # if you installed jank-bin
 ```
 
 ## Nix
-### Install
-We have binary packages in our cachix cache, so installation is quick and easy.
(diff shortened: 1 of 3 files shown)


compiler+runtime/doc/install.md
```diff
@@ -66,6 +66,33 @@ yay -S jank-git   # if you installed jank-git
...
+
+If you pull down the jank repository locally for development, you can enter the
+dev shell and then follow the build instructions linked below. Skip the LLVM
+build step as it's done for you by nix.
+
+```bash
+git clone https://github.com/jank-lang/jank.git
+cd jank
+nix develop
+```
```
jeaye: We have docs for this in `build.md`, which is focused on building jank from source. So I *think* we're good to remove this bit from the install docs.

compiler+runtime/doc/install.md
```diff
@@ -66,6 +66,33 @@ yay -S jank-git   # if you installed jank-git
 yay -S jank-bin   # if you installed jank-bin
 ```
 
+## Nix
+### Install
```
jeaye: Each of these (Ubuntu, Homebrew, etc) has both an "Install" and an "Update" section. For Nix, we're missing that. So, how do folks who have used jank via nix before actually get the latest version?

compiler+runtime/doc/install.md
```diff
@@ -66,6 +66,33 @@ yay -S jank-git   # if you installed jank-git
...
+```bash
+# Make sure cachix is installed, then follow the prompts.
+cachix use jank-lang
+```
+
+Next, you can directly run the jank binary from the nix flake derivation. For
+example, to run the `jank check-health` subcommand:
+
+```bash
+nix run git+https://github.com/jank-lang/jank.git -- check-health
```
jeaye, line 85: Does this provide us with any means to install jank properly, such that there's a `jank` binary available within the user's shell? Or is that a later step, when we get into nixpkgs?

jeaye (changes requested):
Thank you, Kyle! I'm so pleased to see the Nix setup coming along.

compiler+runtime/doc/install.md
kylc: Removed in https://github.com/jank-lang/jank/pull/493/commits/2b138b4c08708535bd8a31390f7fba2a360d794f

compiler+runtime/doc/install.md
kylc: I've added some comments about how nix decides which version to use in different scenarios. Beyond that, I think users must be familiar with nix to know whether they must `nix flake update`, `nixos-rebuild`, or whatever else depending on their own setup.

Thoughts on renaming Install -> Usage since, in the nix case, there aren't really distinct install and update concepts?

compiler+runtime/doc/install.md
kylc: I've added a section which describes how to include the jank flake into any existing flake-based workflow (NixOS configuration or project-local).

jeaye:
Thanks, Kyle!

PR #494, opened 2025-09-14
Add `clojure.string/trim*` functions

djblue:


compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -193,4 +193,106 @@ namespace clojure::string_native
...
+
+    return make_box(s_str.substr(l));
+  }
+
+  static jtl::immutable_string::size_type trimr_index(jtl::immutable_string const &s)
+  {
+    auto const s_size(s.size());
+    jtl::immutable_string::size_type i{ s_size };
+
+    for(; i > 0; i--)
```
jeaye: It's always good practice, in C++, to use pre-increment/decrement, rather than post. This is because those operators are overloadable and values like iterators may provide them. Using the post version will require creating a copy of that value, compared to the pre version, which updates in place. So, this can become a performance concern, since iteration is basically always done in loops (where perf matters most).

All this to say, it's a red flag to see post usage, since it either means complex logic is happening, there's a perf concern, or somebody's not following the rules. :joy: 

So, while it won't affect the perf here, please change it to `--i` (same in the other fns) for both consistency and sanity.

jeaye (changes requested):
Looks great! I just have one bit of feedback.

compiler+runtime/src/cpp/clojure/string_native.cpp
djblue: Good to know, thanks for the breakdown!

jeaye (approved):

PR #496, opened 2025-09-15
Add aset macro in place of aset function in clojure.core

cjbarre:


compiler+runtime/test/jank/aset/fail-two-index-and-value.jank
```diff
@@ -0,0 +1,7 @@
+(cpp/raw "namespace jank::aset::pass_single_index_and_value {
```
jeaye: Please remove these tests. The clojure-test-suite repo is for testing `clojure.core` and other core libs: https://github.com/jank-lang/clojure-test-suite

This test suite is just for testing compiler features.

jeaye (changes requested):
Thanks! Just one request. :)

compiler+runtime/test/jank/aset/fail-two-index-and-value.jank
cjbarre: Done!

jeaye (approved):

jeaye:
Thanks, Cameron!

PR #497, opened 2025-09-15
Add `clojure.string/split*`

djblue:


compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -295,4 +295,69 @@ namespace clojure::string_native
...
+
+    std::string const search_str{ s_str.c_str() };
+
+    native_vector<object_ref> vec;
+    std::sregex_token_iterator iter(search_str.begin(), search_str.end(), regex, -1);
+    std::sregex_token_iterator end;
+
+    if(iter != end && iter->str().empty())
+    {
+      iter++;
```
jeaye: :wink:

jeaye (changes requested):
You're on a roll! One habit to break, though. :upside_down_face:

compiler+runtime/src/cpp/clojure/string_native.cpp
djblue: Ohh, I think I misunderstood your previous comment. I don't know why but I thought it wouldn't apply to iterators. Good to know that I should always prefer pre over post 👍

jeaye: It's specifically *because* of iterators, and other types which overload `++` and `--` that we need to be careful with them. They may not be cheap, since who knows how much state an iterator is holding? Sometimes it's a lot.

jeaye (approved):
:muscle:

PR #502, opened 2025-09-17
Get AOT builds working in debug + release

jeaye:
1. Improve AOT error reporting
2. Rework libjank-standalone to include `clojure.core` for easy linking
3. Setup a resource tracker and embed PCHs into AOT binaries
4. Forward VFS map to CppInterOp for the embedded PCH
5. Patch CppInterOp to create overlay FS prior to creating `clang::Interpreter`
6. Add additional AOT flags to improve perf and usability
7. Add AOT build test to health check
8. Add ability to append new module paths at runtime
9. Re-enable bash tests in CI

@Samy-33 for vis. (no action needed)


jeaye:
@kylc We have a failure in the Nix build now because `jank check-health` is doing an AOT build check. Doing AOT building requires us to have jank's dependency libs available for us to link. In the Nix environment we're using to run `check-health`, that doesn't seem to be the case: https://github.com/jank-lang/jank/actions/runs/17843678533/job/50738742452?pr=502

Basically, AOT works by just compiling every module down to a `.o` file and then invoking Clang from the CLI to link everything together. These are the libs we try to link: https://github.com/jank-lang/jank/pull/502/files#diff-575707f248b2596e1f380efbb419d13fe8b8d408a5d29d25b34b174011a05ca8R196-R207

Is this something we can easily fix for the Nix CI build? If not, I can disable the `check-health` step for now.

kylc:
@jeaye I have a prototype fix for the issues you're running into, but it's not suitable for merging. I won't have more time to look into this for a while, but hopefully this sets someone on the right path. https://github.com/kylc/jank/commit/2a046b80a85ae48106b08304b5f72bd67fa59968

I found two things which are causing the AOT build failures on Nix:
1. When the AOT subsystem invokes clang via the `clang::Driver` API, there is some translation done on the CLI args to put them into `-cc1` style. This is done before exec-ing the clang binary. So once the clang binary is executed, which on Nix points to the wrapper responsible for injecting all of the Nix-specific CFLAGS/LDFLAGS, the arguments are all in this clang-specific style. Nix specifically disables its wrapper functionality in this case since most of the common flags it injects won't be supported in this compiler mode, see [here](https://github.com/NixOS/nixpkgs/blob/ea26e8f23ffe674172ab68c46337b5543c7ccbbb/pkgs/build-support/cc-wrapper/cc-wrapper.sh#L221-L227). A hacky workaround for this is to manually exec the pre-translated clang compile command, rather than letting the driver do the translation before exec-ing clang.
    - You can track this down by running `strace -f -e execve jank check-health` to see the exact clang command being invoked.
3. In the jank cmake, `JANK_DEPS_LIBRARY_DIRS` does not contain what I would expect. On my Nix system it only contains `/lib` (which is empty) rather than a bunch of `/nix/store/...` directories where all of the target dependency shared libs are stored. This means that when the AOT compiler is invoked, it effectively has no library search path and fails to find the libs. As another hacky workaround, I've just injected `NIX_LDFLAGS` into this var, which does contain the proper linker search paths.

With these changes the AOT stage of `jank check-health` now succeeds.

jeaye:
> @jeaye I have a prototype fix for the issues you're running into, but it's not suitable for merging. I won't have more time to look into this for a while, but hopefully this sets someone on the right path. [kylc@2a046b8](https://github.com/kylc/jank/commit/2a046b80a85ae48106b08304b5f72bd67fa59968)
> 
> I found two things which are causing the AOT build failures on Nix:
> 
>     1. When the AOT subsystem invokes clang via the `clang::Driver` API, there is some translation done on the CLI args to put them into `-cc1` style. This is done before exec-ing the clang binary. So once the clang binary is executed, which on Nix points to the wrapper responsible for injecting all of the Nix-specific CFLAGS/LDFLAGS, the arguments are all in this clang-specific style. Nix specifically disables its wrapper functionality in this case since most of the common flags it injects won't be supported in this compiler mode, see [here](https://github.com/NixOS/nixpkgs/blob/ea26e8f23ffe674172ab68c46337b5543c7ccbbb/pkgs/build-support/cc-wrapper/cc-wrapper.sh#L221-L227). A hacky workaround for this is to manually exec the pre-translated clang compile command, rather than letting the driver do the translation before exec-ing clang.
>        
>        * You can track this down by running `strace -f -e execve jank check-health` to see the exact clang command being invoked.
> 
>     2. In the jank cmake, `JANK_DEPS_LIBRARY_DIRS` does not contain what I would expect. On my Nix system it only contains `/lib` (which is empty) rather than a bunch of `/nix/store/...` directories where all of the target dependency shared libs are stored. This means that when the AOT compiler is invoked, it effectively has no library search path and fails to find the libs. As another hacky workaround, I've just injected `NIX_LDFLAGS` into this var, which does contain the proper linker search paths.
> 
> 
> With these changes the AOT stage of `jank check-health` now succeeds.

This is excellent information, Kyle! Thanks for putting in the work to figure this out. I've lifted this up into a ticket: https://github.com/jank-lang/jank/issues/515

PR #503, opened 2025-09-17
Refactor doseq to look like clojure's

gavlooth:
This implements doseq
Had to brake it done to smaller functions. But is pretty close to the original.
Semantics are followed

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -803,11 +803,11 @@
   ([ref]
    (clojure.core-native/deref ref))
   ; TODO: Blocking.
-  ([ref timeout-ms timeout-val]
+  ([ref timeout-ms timeout-val]))
    ;(if (instance? clojure.lang.IBlockingDeref ref)
    ;  (.deref ^clojure.lang.IBlockingDeref ref timeout-ms timeout-val)
    ;  (deref-future ref timeout-ms timeout-val))
-   ))
+   
 
 (def reduced
   "Wraps x in a way such that a reduce will terminate with the value x"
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -803,11 +803,11 @@
   ([ref]
    (clojure.core-native/deref ref))
   ; TODO: Blocking.
-  ([ref timeout-ms timeout-val]
+  ([ref timeout-ms timeout-val]))
    ;(if (instance? clojure.lang.IBlockingDeref ref)
    ;  (.deref ^clojure.lang.IBlockingDeref ref timeout-ms timeout-val)
    ;  (deref-future ref timeout-ms timeout-val))
-   ))
+
```
jeaye, line 864: Looks like the entire file has been formatted. Please only change the lines related to `doseq`.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -1400,20 +1400,72 @@
...
+(defn ^:private assert-arguments
+  [& pairs]
+  (let [pairs (partition 2 pairs)]
+    (loop [ps pairs]
+      (when (seq ps)
+        (let [[pred msg] (first ps)]
+          (when (not pred)
+            (throw (str msg))))
+        (recur (next ps))))))
```
jeaye: I believe this is a duplicate of `assert-macro-args`, already in this file.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -1400,20 +1400,72 @@
...
-          nil
-          s)))
+  bindings and filtering as provided by \"for\".  Does not retain
+  the head of the sequence. Returns nil."
+  [seq-exprs & body]
+  (assert-arguments
+    (vector? seq-exprs) "doseq requires a vector for its binding"
+    (even? (count seq-exprs)) "doseq requires an even number of forms in binding vector")
+  (let [[_ form] (emit-doseq-step nil (seq seq-exprs) body)]
+    form))
```
jeaye: You mentioned this had to be broken up into separate functions. Why is that? If there's an issue preventing us from using the code that Clojure has, we should identify it so it can be fixed.

jeaye (changes requested):
Thanks for the PR! Apologies for the delay. I have some questions and requests, but I appreciate the help. :slightly_smiling_face:

compiler+runtime/src/jank/clojure/core.jank
gavlooth: @jeaye It was broken for clarity, the huge macro was difficult to follow. I don't remember the issue thought

gavlooth: I can rework it, no problem

jeaye: Ok, cool. Let's keep it as close to Clojure as we possibly can. This will make porting over new changes in Clojure much easier, since they'll be 1:1.

gavlooth: @jeaye doseq updated from clojure core, only tiny adjasments made

jeaye: Thanks! The PR is showing a huge diff right now, as well as conflicts. It might we easier to spin up a new one or rebase from main, so we can actually just see the doseq changes.

jeaye:
I have ported Clojure's `doseq` as part of https://github.com/jank-lang/jank/pull/661, so I'll close this one.

Issue #10, opened 2022-08-18
Excited about this project!


eccentric-j:
Haven't dived in yet but really love the concept and look forward to seeing it develop. Not crazy about the name but I appreciate the ambition either way! Hope you don't mind but wanted to show some support.

jeaye:
Thanks for reaching out, Jay! Also, thank you for that one time sponsorship. :)

I'm also really excited about what's upcoming for jank; hopefully the name grows on you.

rosshadden:
I'll make a few suggestions since an issue like this was opened:
 - if you enable discussions, people will be able to say stuff like this there (if that's what you want obviously)
- I suggest moving the language to its own GitHub org if you're open to it: `jank-lang/jank` or something. That will help not only the visibility and discovery of the project, but might also make it feel more like a community project and might solicit more outside contributors. I'm not a psychologist, but that's my theory 👨‍⚕️
- like the user above, I'm quite excited to see where this project goes, and if it goes where I think a lot of us think it can, it might end up being a new favorite language for quite a lot of people!

jeaye:
Great suggestions. Two weeks ago, it was just me. Now there's a whole lot more community management to do. I'm still catching up. :)

1. Done!
2. Done! Turns out I already had the jank-lang org created.

PR #509, opened 2025-09-18
More cpp interop in `clojure.core`

djblue:


compiler+runtime/include/cpp/clojure/core_native.hpp
```diff
@@ -1,5 +1,21 @@
 #pragma once
 
 #include <jank/c_api.h>
+#include <jank/runtime/object.hpp>
 
 extern "C" jank_object_ref jank_load_clojure_core_native();
+
+namespace clojure::core_native
```
jeaye, line 8: (I know this is a draft, so I'm not doing a full review. Just a note)

We ultimately want to get rid of `core_native` entirely. This stuff should go under `runtime/core.hpp` until it fits under another grouping like `runtime/core/seq.hpp`.

djblue: Should I make the move as part of this PR?

jeaye: Just for the new things. We don't want to be adding anything new to `core_native`.

Each of the `foo_native` files was a temporary solution, before jank had good interop. They should all be going away, with their functions just being a part of jank's runtime.

djblue: My current plan is to work towards removing all of the `clojure.core-native` vars in clojure.core, then refactor the cpp code to be better organized 👍

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -1317,9 +1317,10 @@
   ([x]
    x)
   ([l r]
-   (clojure.core-native/min l r))
+   (cpp/jank.runtime.min l r))
   ([l r & args]
-   (let [res (clojure.core-native/min l r)]
+   ;; TODO: swap to interop, currently broken
```
djblue: I've noticed that using interop here causes issues. This is true for other fns with the same style of:

 ```clojure
(let [res ..] (recur ...))
```

Comment (jeaye) [reply compiler+runtime/src/jank/clojure/core.jank]: Ok, taking a look.

Comment (jeaye) [reply compiler+runtime/src/jank/clojure/core.jank]: Hm, what's broken? That works for me locally.

I did:

```clojure
(defn add
  ([l r]
   (cpp/jank.runtime.add l r))
  ([l r & args]
   (let [res (cpp/jank.runtime.add l r)]
     (if (empty? args)
       res
       (recur res (first args) (next args))))))

(add 2 3)
```

djblue: Ohh, sorry for not clarifying. You need to call the varargs arity to trigger the exception:

```clojure
user=> (add 1 2 3)
Uncaught exception: invalid object type: unknown raw value 192

#4  in jank::runtime::visit_object<jank::runtime::to_code_string at visit.hpp:219
#5  in jank::runtime::to_code_string at to_string.cpp:23
#6  in jank::runtime::visit_number_like<jank::runtime::add at visit.hpp:437
#7  in jank::runtime::visit_number_like<jank::runtime::add at visit.hpp:424
#8  in jank::runtime::visit_number_like<jank::runtime::add at visit.hpp:434
#9  in jank::runtime::add at math.cpp:41
```

jeaye: :facepalm: No, not your fault. The change was obviously in the variadic arity. I just brain segfaulted.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -1288,48 +1390,58 @@
...
+(defn subs
   "Returns the substring of s beginning at start inclusive, and ending
    at end (defaults to length of string), exclusive."
-  clojure.core-native/subs)
+  ([s start]
+   (cpp/jank.runtime.subs s start))
+  ([s start end]
+   (cpp/jank.runtime.subs s start end)))
 
 (def ^:private first-index-of clojure.core-native/first-index-of)
```
djblue: I think these are unused and private, should I remove them?

jeaye: Go for it!

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -5421,20 +5421,17 @@
   ;; (clojure.lang.RT/uncheckedDoubleCast x)
   (throw "TODO: port unchecked-double"))
 
-(defn ratio?
+(def ratio?
   "Returns true if n is a Ratio"
-  [r]
-  (cpp/jank.runtime.is_ratio r))
```
djblue: It looks like these changes were the ones causing the CI errors:

```
jank-git> ==== SOURCE BEGIN ====
jank-git> #pragma clang diagnostic push
jank-git> #pragma clang diagnostic ignored "-Wformat-security"
jank-git> __attribute__((used)) __attribute__((annotate("__cling__ptrcheck(off)")))
jank-git> extern "C" [[gnu::always_inline]]
jank-git> inline void clojure_core_G__10734(void* obj, int nargs, void** args, [[gnu::nonnull]] void* ret)
jank-git> {
jank-git>    new (ret) (jank::runtime::obj::big_integer_ref) (jank::runtime::convert<native_big_integer>::into_object((const boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<>>&)*(const boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<>>*)args[0]));
jank-git> }
jank-git> #pragma clang diagnostic pop
jank-git> ==== SOURCE END ====
jank-git> Segmentation fault (core dumped)
jank-git> ninja: build stopped: subcommand failed.
error: Cannot build '/nix/store/05xwpgl5ghk24gcscgzr3jly9nx4736k-jank-git.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/5khm6j3qmcrmwv0vqvg06hjd3nf3x84p-jank-git
       Last 25 log lines:
       > [258/282] Generating classes/core-libraries
       > FAILED: classes/core-libraries core-libs/clojure/core.o /build/source/compiler+runtime/build/classes/core-libraries /build/source/compiler+runtime/build/core-libs/clojure/core.o
       > cd /build/source/compiler+runtime/build && /build/source/compiler+runtime/build/jank-phase-1 compile-module -o /build/source/compiler+runtime/build/core-libs/clojure/core.o clojure.core && touch /build/source/compiler+runtime/build/classes/core-libraries
       > In file included from <<< inputs >>>:1:
       > input_line_228:7:76: error: unknown type name 'native_big_integer'; did you mean 'jank::native_big_integer'?
       >     7 |    new (ret) (jank::runtime::obj::big_integer_ref) (jank::runtime::convert<native_big_integer>::into_object((const boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<>>&)*(const boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<>>*)args[0]));
       >       |                                                                            ^~~~~~~~~~~~~~~~~~
       >       |                                                                            jank::native_big_integer
       > /build/source/compiler+runtime/include/cpp/jank/type.hpp:32:9: note: 'jank::native_big_integer' declared here
       >    32 |   using native_big_integer = boost::multiprecision::cpp_int;
       >       |         ^
       > TClingCallFunc::make_wrapper:Failed to compile
       > ==== SOURCE BEGIN ====
       > #pragma clang diagnostic push
       > #pragma clang diagnostic ignored "-Wformat-security"
       > __attribute__((used)) __attribute__((annotate("__cling__ptrcheck(off)")))
       > extern "C" [[gnu::always_inline]]
       > inline void clojure_core_G__10734(void* obj, int nargs, void** args, [[gnu::nonnull]] void* ret)
       > {
       >    new (ret) (jank::runtime::obj::big_integer_ref) (jank::runtime::convert<native_big_integer>::into_object((const boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<>>&)*(const boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<>>*)args[0]));
       > }
       > #pragma clang diagnostic pop
       > ==== SOURCE END ====
```

compiler+runtime/include/cpp/clojure/core_native.hpp
djblue: Are you okay with these functions living in this namespace until `clojure.core-native` is removed? I haven't added any new functions, simply made current ones non-static to ease interop migration and make it easier to follow / verify.

djblue:
I think this PR is ready for a closer look. The remaining `clojure.core-native` usage needs further investigation to resolve issues I ran into when trying to use interop.

compiler+runtime/include/cpp/clojure/core_native.hpp
jeaye: Yeah, that's ok.

compiler+runtime/src/jank/clojure/core.jank
jeaye: I'm hoping this issue is fixed with https://github.com/jank-lang/jank/pull/513 now merged. Can you let me know?

djblue: Merged main locally but still seeing the problem.

jeaye: Thanks for checking. This is another CppInterOp library bug. I will look into this and the recursion issue you found in the coming few days.

jeaye:
@djblue I have fixed the two issues you hit on this PR. The changes have been pushed to your branch.

1. `recur` with a native object
2. Bad CppInterOp codegen

Given that CI is green now, I think you should be good to add any remaining bits to this PR which were previously blocked.

Note, I already reverted your revert commit, since that was used for testing.

djblue:
The rest of the usage of core-native in clojure.core has the following issues:

- Interop can't resolve static methods
  - `clojure.core-native/range`
  - `clojure.core-native/integer-range`

- Requires try/catch because it's in a binding
  - `clojure.core-native/load-module`

- Breaks `ns` and `require`, not sure why
  - `clojure.core-native/ns?`
  - `clojure.core-native/var-ns`
  - test/jank/syntax-quote/pass-unaffected.jank (fails)
  - test/jank/syntax-quote/pass-resolve-alias.jank (fails)

jeaye:
> The rest of the usage of core-native in clojure.core has the following issues:
> 
>     * Interop can't resolve static methods
>       
>       * `clojure.core-native/range`
>       * `clojure.core-native/integer-range`
> 

Ok, I'll look into these tomorrow.

>     * Requires try/catch because it's in a binding
>       
>       * `clojure.core-native/load-module`
> 

This will be fixed by merging the IR PR. We're currently blocked on Jianling for that. You could possibly take that PR over, but it's a much steeper learning curve, since it's all IR gen and it's quite complicated. We're likely better off waiting for him to get to it.

>     * Breaks `ns` and `require`, not sure why
>       
>       * `clojure.core-native/ns?`
>       * `clojure.core-native/var-ns`
>       * test/jank/syntax-quote/pass-unaffected.jank (fails)
>       * test/jank/syntax-quote/pass-resolve-alias.jank (fails)

Ok, I can look into these if you're stuck.

jeaye:
> The rest of the usage of core-native in clojure.core has the following issues:
> 
>     * Interop can't resolve static methods
>       
>       * `clojure.core-native/range`
>       * `clojure.core-native/integer-range`

This has been addressed. The issue wasn't that the fns couldn't be *resolved*. The issue was that the resolved fns couldn't be called, since we had an `object_ref`, but those static fns actually take an `integer_ref`. I've hooked up implicit trait conversions between typed and untyped objects so that we'll automatically try to do that now.

I'll tackle the `ns` and `require` issues next.

jeaye:
>     * Breaks `ns` and `require`, not sure why
>       
>       * `clojure.core-native/ns?`
>       * `clojure.core-native/var-ns`
>       * test/jank/syntax-quote/pass-unaffected.jank (fails)
>       * test/jank/syntax-quote/pass-resolve-alias.jank (fails)

These work for me now, so I suspect I fixed the issues with my other changes. Please give it a try on your end and let me know if any problems still exist.

So, aside from the exception IR, I think there are no other known issues here. Let me know if I missed anything.

djblue:
So it compiles, but with the following diff I get a runtime error:

```
diff --git a/compiler+runtime/src/jank/clojure/core.jank b/compiler+runtime/src/jank/clojure/core.jank
index da74ee53..1c9f5e8f 100644
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -4022,7 +4022,7 @@
    returns the namespace named by it, throwing an exception if not
    found."
   [ns-or-sym]
-  (if (clojure.core-native/ns? ns-or-sym)
+  (if (cpp/clojure.core_native.is_ns ns-or-sym)
     ns-or-sym
     (let [found (find-ns ns-or-sym)]
       (if (nil? found)
```

```clojure
% jank repl
Uncaught exception: invalid object type (expected ns found symbol)

#4  (inlined)          in jank::runtime::try_object<jank::runtime::ns> at rtti.hpp:41
#5  in clojure::core_native::alias at core_native.cpp:242
#6  in jank::runtime::obj::jit_function::call at jit_function.cpp:88
#7  in jank::runtime::dynamic_call at callable.cpp:117
#8  in jank::runtime::visit_object<jank::runtime::dynamic_call at visit.hpp:179
#9  in jank::runtime::dynamic_call at callable.cpp:96
#10 in jank_call2 at c_api.cpp:133
```
but the other var seems good to go 👍

jeaye:
> So it compiles, but with the following diff I get a runtime error:
> 
> ```
> diff --git a/compiler+runtime/src/jank/clojure/core.jank b/compiler+runtime/src/jank/clojure/core.jank
> index da74ee53..1c9f5e8f 100644
> --- a/compiler+runtime/src/jank/clojure/core.jank
> +++ b/compiler+runtime/src/jank/clojure/core.jank
> @@ -4022,7 +4022,7 @@
>     returns the namespace named by it, throwing an exception if not
>     found."
>    [ns-or-sym]
> -  (if (clojure.core-native/ns? ns-or-sym)
> +  (if (cpp/clojure.core_native.is_ns ns-or-sym)
>      ns-or-sym
>      (let [found (find-ns ns-or-sym)]
>        (if (nil? found)
> ```
> 
> ```clojure
> % jank repl
> user=> (require '[clojure.string :as str])
> Uncaught exception: invalid object type (expected ns found symbol)
> 
> Stack trace (most recent call first):
> #4  (inlined)          in jank::runtime::try_object<jank::runtime::ns> at rtti.hpp:41
> #5  in clojure::core_native::alias at core_native.cpp:242
> #6  in jank::runtime::obj::jit_function::call at jit_function.cpp:88
> #7  in jank::runtime::dynamic_call at callable.cpp:117
> #8  in jank::runtime::visit_object<jank::runtime::dynamic_call at visit.hpp:179
> #9  in jank::runtime::dynamic_call at callable.cpp:96
> #10 in jank_call2 at c_api.cpp:133
> ```
> 
> but the other var seems good to go 👍

A fix for this has been pushed, along with a test. Aside from the `try` issue, I think we should be all set.

jeaye:
Great work, Chris!

PR #510, opened 2025-09-19
Fix `parse_boolean`

djblue:
Before:

```clojure
% jank repl
user=> (identical? true true)
false
user=> (identical? false false)
false
user=> (identical? true false)
false
user=> (identical? false true)
false
```

After:

```clojure
% jank repl                                  
user=> (identical? true true)
true
user=> (identical? false false)
true
user=> (identical? true false)
false
user=> (identical? false true)
false
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -1513,7 +1513,7 @@ namespace jank::read::parse
     auto const token((*token_current).expect_ok());
     ++token_current;
     auto const b(std::get<bool>(token.data));
-    return object_source_info{ make_box<obj::boolean>(b), token, token };
+    return object_source_info{ b ? jank_true : jank_false, token, token };
   }
 
   processor::object_result processor::parse_symbol()


compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1513,7 +1513,7 @@ namespace jank::read::parse
     auto const token((*token_current).expect_ok());
     ++token_current;
     auto const b(std::get<bool>(token.data));
-    return object_source_info{ make_box<obj::boolean>(b), token, token };
+    return object_source_info{ b ? jank_true : jank_false, token, token };
```
djblue: An alternative would be to fix this in make_box, but I wasn't sure.

djblue: Although, it looks like it should be working [jank/runtime/core/make_box.hpp](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/include/cpp/jank/runtime/core/make_box.hpp#L21-L25) that way already 🤔

jeaye: This should have no effect. :thinking:

jeaye: Oh, it's because of the `<obj::boolean>`. It doesn't get into the overload you linked, since that overload is not a template. We should fix this in the `make_box` template instead.

It's getting here: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/include/cpp/jank/runtime/oref.hpp#L427

Ideally, we can specialize this template where `T::obj_type == object_type::boolean` to just return the constants. Otherwise, we're actually creating new bools which are not going to be identical with the constants. That shouldn't be possible.

djblue: I took my best shot at specializing the template. It seems to work, but not sure I did it correctly.

compiler+runtime/include/cpp/jank/runtime/oref.hpp
```diff
@@ -392,6 +398,12 @@ namespace jank::runtime
     return o;
   }
 
+  template <typename T>
+  constexpr oref<obj::boolean> make_box(bool const b)
```
jeaye: This isn't a specialization! It's just a new template. You'd need `template <>` and `make_box<obj::boolean>` to make this a specialization.

djblue: Ohh, gotcha. I tried a couple of others things, but since `jank_true`/`jank_false` are both `oref<struct boolean>`, I tried specializing:

```cpp
  template <typename T, typename... Args>
  requires behavior::object_like<T>
  oref<T> make_box(Args &&...args)
```
but since I forward declared the `struct boolean`, I couldn't get cpp to match this template because of `behavior::object_like`.

Not sure if I'm heading in the right direction or completely lost. I am no good at template meta programming lol

jeaye: > Ohh, gotcha. I tried a couple of others things, but since `jank_true`/`jank_false` are both `oref<struct boolean>`, I tried specializing:
> 
> ```c++
>   template <typename T, typename... Args>
>   requires behavior::object_like<T>
>   oref<T> make_box(Args &&...args)
> ```
> 
> but since I forward declared the `struct boolean`, I couldn't get cpp to match this template because of `behavior::object_like`.
> 
> Not sure if I'm heading in the right direction or completely lost. I am no good at template meta programming lol

Yeah, you don't have `boolean` yet. What you do have is `T::obj_type` and `object_type::boolean`, which you could compare.

```cpp
  template <typename T>
  requires (T::obj_type == object_type::boolean)
  oref<T> make_box(bool const b)
```

jeaye: If that doesn't work for you, let me know and I'll jump in and fix it.

djblue: Ohh, I didn't know you could use `T::obj_type == object_type::boolean` as part of requires, good to know. I added this specialization and it seems to work 👍

jeaye (approved):
You're the man, Chris.

PR #511, opened 2025-09-19
Fix `swap!` varargs arity

djblue:
Before
```clojure
% jank repl
user=> (swap! (atom {}) update :x assoc :y :z)
Uncaught exception: not conjable: #object [clojure.core/assoc jit_function 0x70ef1044ccf8]

#4  in jank::runtime::conj at seq.cpp:507
```
After:
```clojure
% jank repl
user=> (swap! (atom {}) update :x assoc :y :z)
{:x {:y :z}}
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
@@ -6,6 +6,8 @@
 
 namespace jank::runtime::obj
 {
+  using jank::runtime::cons;
+
   atom::atom(object_ref const o)
     : val{ o.data }
   {
@@ -111,7 +113,8 @@ namespace jank::runtime::obj
     while(true)
     {
       auto v(val.load());
-      auto const next(apply_to(fn, conj(a1, conj(a2, rest))));
+      auto args(cons(v, cons(a1, cons(a2, rest))));
+      auto const next(apply_to(fn, args));
       if(val.compare_exchange_weak(v, next.data))
       {
         return next;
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
```diff
@@ -111,7 +111,9 @@ namespace jank::runtime::obj
     while(true)
     {
       auto v(val.load());
-      auto const next(apply_to(fn, conj(a1, conj(a2, rest))));
+      auto const args(
+        jank::runtime::cons(v, jank::runtime::cons(a1, jank::runtime::cons(a2, rest))));
```
jeaye: No need for `jank::runtime::` in any of these lines. We're already in the `jank::runtime::obj` namespace.

jeaye (changes requested):
Good find!

compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
djblue: I think doing so causes a name collision with `struct cons`.

jeaye: Ah, in that case, we use `runtime::cons`. Basically, use as little qualification as possible. You should basically never need `jank::` within the jank code.

djblue: I was wondering how to do that 😂

PR #516, opened 2025-09-20
Fix `contains?` for `nil` in `array_map`

djblue:
I noticed that `hash-map` was working just fine for this use case and updated the `array-map` implementation to mirror it.

Before
```clojure
(contains? {:a nil} :a) ; false
(get {:a nil} :a :foo) ; :foo
(find {:a nil} :a) ; nil
```

After:
```clojure
(contains? {:a nil} :a) ; true
(get {:a nil} :a :foo) ; nil
(find {:a nil} :a) ; [:a nil]
```

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/detail/native_array_map.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/detail/native_array_map.hpp
@@ -2,6 +2,7 @@
 
 #include <type_traits>
 
+#include <jtl/option.hpp>
 #include <jtl/ref.hpp>
 
 #include <jank/runtime/object.hpp>
@@ -41,7 +42,7 @@ namespace jank::runtime::detail
 
     void erase(object_ref const key);
 
-    object_ref *find(object_ref const key) const;
+    jtl::option<object_ref> find(object_ref const key) const;
 
     uhash to_hash() const;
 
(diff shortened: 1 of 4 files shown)


compiler+runtime/include/cpp/jank/runtime/detail/native_array_map.hpp
```diff
@@ -41,7 +41,7 @@ namespace jank::runtime::detail
 
     void erase(object_ref const key);
 
-    object_ref find(object_ref const key) const;
+    object_ref *find(object_ref const key) const;
```
jeaye: We don't want to introduce raw pointers into any new APIs. They're considered very poor practice in C++, since they have no notion of ownership. If the goal is to distinguish "not found" from "nil", we should either use a `jtl::option` or at least a `jtl::ptr`.

jeaye (changes requested):
Thanks, Chris! One request about how we can do this in a more "modern C++" style.

compiler+runtime/include/cpp/jank/runtime/detail/native_array_map.hpp
djblue: I switched the raw pointer to `jtl::option` 👍

PR #517, opened 2025-09-21
Implement `add-watch` / `remove-watch` for atoms

djblue:
This should be implemented for other ref types, but I started with one I'm familiar with. For the most part, this code follows [`Aref.java`](https://github.com/clojure/clojure/blob/b775c8a289339b4a8fb3f1800a017147c53b0b42/src/jvm/clojure/lang/ARef.java#L72-L84) and [`Atom.java`](https://github.com/clojure/clojure/blob/b775c8a289339b4a8fb3f1800a017147c53b0b42/src/jvm/clojure/lang/Atom.java#L41).


compiler+runtime/include/cpp/jank/runtime/obj/atom.hpp
```diff
@@ -46,5 +47,6 @@ namespace jank::runtime::obj
 
     object base{ obj_type };
     std::atomic<object *> val{};
+    mutable persistent_hash_map_ref watches{};
```
jeaye: 1. I don't think this needs to be `mutable`. In C++, this keyword means that the member can be changed, even in a `const` member function. It's not needed just to say the member can be changed in general.
2. This will need synchronization, since it could be changed from multiple threads. Java does this with `synchronized` on `addWatch`. We will need to wrap this in a `folly::Synchronized` and update the usage.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -664,4 +664,20 @@ namespace jank::runtime
              expect_object<obj::inst>(o)->value.time_since_epoch())
       .count();
   }
+
+  object_ref add_watch(object_ref const reference, object_ref const key, object_ref const fn)
+  {
+    auto const a(try_object<obj::atom>(reference));
+    a->watches = a->watches->assoc(key, fn);
```
jeaye: When we support the others, I think we should add a `ref_like` behavior concept which requires an `add_watch` and `remove_watch` member. Then, in this core fn, we can just visit and check if the type is `ref_like`.

These concepts have no runtime cost, so we might as well use a consistent interface.

djblue: I went ahead and did this now since it didn't sound too difficult.

compiler+runtime/include/cpp/jank/runtime/behavior/ref_like.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+namespace jank::runtime::behavior
+{
+  template <typename T>
+  concept ref_like = requires(T * const t) {
```
jeaye, line 6: Please name the file `ref_like`, to match the concept.

compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
```diff
@@ -181,6 +207,23 @@ namespace jank::runtime::obj
...
-    return make_box(val.compare_exchange_weak(old, new_val.data));
+    auto const ret(val.compare_exchange_weak(old, new_val.data));
+    if(ret)
+    {
+      notify_watches(this, old_val, new_val);
+    }
+    return make_box(ret);
+  }
+
+  void atom::add_watch(object_ref key, object_ref fn)
```
jeaye: These params can be `const`.

compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
```diff
@@ -181,6 +207,23 @@ namespace jank::runtime::obj
...
+  {
+    auto locked_watches(this->watches.wlock());
+    *locked_watches = (*locked_watches)->assoc(key, fn);
+  }
+
+  void atom::remove_watch(object_ref key)
```
jeaye: `key` can be `const`.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -664,4 +665,48 @@ namespace jank::runtime
...
+
+  object_ref remove_watch(object_ref const reference, object_ref const key)
+  {
+    visit_object(
+      [=](auto const typed_o) -> void {
+        using T = typename decltype(typed_o)::value_type;
+
+        if constexpr(behavior::ref_like<T>)
+        {
+          return typed_o->remove_watch(key);
```
jeaye: It's generally considered confusing practice to use `return` on a call which returns `void`. (Same thing in `add_watch`)

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -664,4 +665,48 @@ namespace jank::runtime
...
+        }
+        else
+        {
+          throw std::runtime_error{ util::format(
+            "Value does not support 'remove-watch' because it is not ref_like: {}",
+            typed_o->to_string()) };
```
jeaye: We should always print values with `to_code_string` instead, to make sure that strings get quoted and such. (Same with `add_watch`)

compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
```diff
@@ -43,10 +45,25 @@ namespace jank::runtime::obj
     return val.load();
   }
 
+  static void notify_watches(atom_ref const a, object_ref const old_val, object_ref const new_val)
+  {
+    auto const locked_watches(a->watches.rlock());
+    for(auto const entry : (*locked_watches)->data)
+    {
+      auto const fn(entry.second);
+      if(!fn.is_nil())
```
jeaye: Nit: We have `is_some` for this.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -664,4 +665,48 @@ namespace jank::runtime
              expect_object<obj::inst>(o)->value.time_since_epoch())
       .count();
   }
+
+  object_ref add_watch(object_ref const reference, object_ref const key, object_ref const fn)
+  {
+    visit_object(
+      [=](auto const typed_o) -> void {
```
jeaye: We always name the visited param after the value we're visiting. So, this should be called `typed_reference`. (Same with the fn below)

compiler+runtime/src/cpp/jank/runtime/obj/atom.cpp
```diff
@@ -43,10 +45,25 @@ namespace jank::runtime::obj
     return val.load();
   }
 
+  static void notify_watches(atom_ref const a, object_ref const old_val, object_ref const new_val)
+  {
+    auto const locked_watches(a->watches.rlock());
+    for(auto const entry : (*locked_watches)->data)
```
jeaye, line 51: Just a question: does the order matter here? i.e. does Clojure promise any sort of ordering for notifications? If so, looping through the map will not provide stable ordering (as the map changes) and it will not persist the ordering in which the watches were added.

djblue: It looks like Clojure uses a [`PersistentHashMap`](https://github.com/clojure/clojure/blob/b775c8a289339b4a8fb3f1800a017147c53b0b42/src/jvm/clojure/lang/ARef.java#L19C43-L19C60) to store watches. So I assume they can't guarantee any order.

jeaye: Ok, sounds good. Thanks for looking it up! This is a good thing to leave in a comment, above the definition of `watches`, I think. Just as a friendly reminder to anyone in the code.

jeaye (approved):
Nice work, Chris!

PR #518, opened 2025-09-21
Make `persistent_vector` callable

djblue:


compiler+runtime/src/cpp/jank/evaluate.cpp
```diff
@@ -415,6 +415,7 @@ namespace jank::evaluate
             }
           }
           else if constexpr(std::same_as<T, obj::persistent_hash_set>
+                            || std::same_as<T, obj::persistent_vector>
```
jeaye, line 418: So, this will cover the non-codegen case. To cover the codegen case, we need to also update all of the `dynamic_call` functions in `callable.cpp`.

jeaye (changes requested):
Thank, Chris! One piece missing.

PR #519, opened 2025-09-21
Implement `assoc` for `persistent_vector`

djblue:
```clojure
user=> (associative? [])
true
user=> (assoc [] 0 1)
[1]
user=> (assoc [0 1 2] 0 :foo)
[:foo 1 2]
user=> (assoc [0 1 2] 3 :foo)
[0 1 2 :foo]
user=> (assoc [0 1 2] 4 :foo)
Uncaught exception: Index out of bounds.
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/obj/persistent_vector.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/persistent_vector.cpp
@@ -242,6 +242,36 @@ namespace jank::runtime::obj
     }
   }
 
+  persistent_vector_ref persistent_vector::assoc(object_ref const key, object_ref const val) const
+  {
+    if(key->type != object_type::integer)
+    {
+      throw std::runtime_error{ "Key must be integer." };
+    }
+
+    auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
+    auto const size(data.size());
+
+    if(i > size)
+    {
+      throw std::runtime_error{ "Index out of bounds." };
+    }
+
+    if(i == size)
+    {
+      auto vec(data.push_back(val));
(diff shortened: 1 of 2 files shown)


compiler+runtime/src/cpp/jank/runtime/obj/persistent_vector.cpp
```diff
@@ -242,6 +242,36 @@ namespace jank::runtime::obj
...
+    {
+      auto vec(data.push_back(val));
+      return make_box<persistent_vector>(meta, std::move(vec));
+    }
+
+    auto vec(data.set(i, val));
+    return make_box<persistent_vector>(meta, std::move(vec));
+  }
+
+  persistent_vector_ref persistent_vector::dissoc(object_ref const /*key*/) const
```
djblue, line 270: I implemented this to support `associative?`, but vectors don't support `dissoc`, only `assoc`.

compiler+runtime/src/cpp/jank/runtime/obj/persistent_vector.cpp
```diff
@@ -242,6 +242,36 @@ namespace jank::runtime::obj
...
+
+    auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
+    auto const size(data.size());
+
+    if(i > size)
+    {
+      throw std::runtime_error{ "Index out of bounds." };
+    }
+
+    if(i == size)
```
djblue, line 260: Since [`.set`](https://sinusoid.es/immer/containers.html#_CPPv2NK5immer6vector3setE9size_type10value_type) is `Undefined for index >= size()`, we need to check for this case explicitly and `push_back` instead.

PR #461, opened 2025-09-05
Big decimal Support

jianlingzhong:

PR #522, opened 2025-09-22
Address Big Decimal Requested Changes

djblue:
Since #461 is so close and I felt like I could address the requested changes, I thought I would help get this across the line.

### DIFF
--- a/compiler+runtime/src/cpp/jank/analyze/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/analyze/processor.cpp
@@ -4133,6 +4133,18 @@ namespace jank::analyze
                  " The type specified here should be the exact type of the value originally "
                  "passed to 'cpp/box'.",
                  Cpp::GetTypeAsString(type_expr->type)),
+               object_source(type_obj),
+               latest_expansion(macro_expansions))
+        ->add_usage(read::parse::reparse_nth(l, 1));
+    }
+
+    auto const value_type{ cpp_util::expression_type(value_expr) };
+    if(!cpp_util::is_any_object(value_type))
+    {
+      return error::analyze_invalid_cpp_unbox(
+               util::format("Unable to unbox value of type '{}', since it's not a jank object type."
+                            " You can only unbox the same object you get back from 'cpp/box'.",
+                            Cpp::GetTypeAsString(value_type)),
                object_source(value_obj),
                latest_expansion(macro_expansions))
         ->add_usage(read::parse::reparse_nth(l, 2));


compiler+runtime/include/cpp/jank/type.hpp
```diff
@@ -33,7 +33,8 @@ namespace jank
   using native_big_integer
     = boost::multiprecision::number<boost::multiprecision::cpp_int_backend<>>;
   using native_big_decimal
-    = boost::multiprecision::number<boost::multiprecision::cpp_dec_float<100>>;
+    = boost::multiprecision::number<boost::multiprecision::cpp_dec_float<100>,
```
djblue, line 36: I wasn't sure exactly how to address this analyzer warning, and this was the only thing that made sense that I could find online.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -224,6 +243,16 @@ namespace jank::read::lex
...
+  bool big_decimal::operator==(big_decimal const &rhs) const
+  {
+    return number_literal == rhs.number_literal;
+  }
+
+  bool big_decimal::operator!=(big_decimal const &rhs) const
+  {
+    return !(*this == rhs);
+  }
+
```
jeaye: These fns don't need to exist. The C++ compiler will give us the operators for free, since all members already implement them.

compiler+runtime/include/cpp/jank/runtime/core/math.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <jank/runtime/object.hpp>
+#include <jank/runtime/obj/big_decimal.hpp>
```
jeaye: We forward declare (on line 13) so that we don't need to include the header here. So let's remove this include.

compiler+runtime/include/cpp/jank/type.hpp
jeaye: Hm, what was the warning? I'm not sure what the problem was and what you changed.

compiler+runtime/src/cpp/jank/runtime/obj/big_decimal.cpp
```diff
@@ -0,0 +1,184 @@
+#include "jank/runtime/visit.hpp"
+#include <jank/util/fmt.hpp>
+
+
+#include <jank/runtime/obj/big_decimal.hpp>
```
jeaye: These should be combined into one group. Also, generally the corresponding `.hpp` for a `.cpp` goes first. Lastly, please use `<>` instead of `""` for all includes.

compiler+runtime/src/cpp/jank/runtime/obj/big_integer.cpp
```diff
@@ -1,3 +1,4 @@
+#include <jank/runtime/obj/big_decimal.hpp>
```
jeaye: This include is added, but never used in the file. Is it really needed?

compiler+runtime/test/jank/form/case/pass-big-decimal.jank
```diff
@@ -0,0 +1,9 @@
+(assert
+ (=
+  (case 99999.9M
+    :no-match                             :no-match
+    99999.9M                              :big-nines
+    :default)
+  :big-nines))
+
+:success
```
jeaye: This whole test can just be a single `case`, without the `assert` or `=` or trailing `:success`.

```clojure
(case 99999.9M
  :no-match :no-match
  99999.9M :success
  :default)
```

jeaye (changes requested):
Thanks for the changes, Chris! I did one more round of review and after that we should be good.

compiler+runtime/include/cpp/jank/type.hpp
djblue: The analyzer warning I got was the following:
```cpp
/home/chris/repos/jank/compiler+runtime/third-party/boost-multiprecision/include/boost/multiprecision/detail/default_ops.hpp:3796:4: error: Address of stack memory associated with temporary object of type 'detail::abs_funct<typename detail::backend_type<detail::expression<boost::multiprecision::detail::subtract_immediates, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100, int, void>, boost::multiprecision::et_on>, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100, int, void>, boost::multiprecision::et_on>, void, void>>::type>' is still referred to by a temporary object on the stack upon returning to the caller.  This will be a dangling reference [clang-analyzer-core.StackAddressEscape,-warnings-as-errors]
 3796 |    return detail::expression<
      |    ^
/home/chris/repos/jank/compiler+runtime/src/cpp/jank/runtime/obj/big_decimal.cpp:66:14: note: Calling 'operator=='
   66 |     return !(l == r);
      |              ^~~~~~
/home/chris/repos/jank/compiler+runtime/src/cpp/jank/runtime/obj/big_decimal.cpp:61:12: note: Calling 'operator=='
   61 |     return r == l;
      |            ^~~~~~
/home/chris/repos/jank/compiler+runtime/src/cpp/jank/runtime/obj/big_decimal.cpp:51:12: note: Calling 'abs<boost::multiprecision::detail::subtract_immediates, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100>>, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100>>, void, void>'
   51 |     return abs(l - r_as_decimal) < std::numeric_limits<native_big_decimal>::epsilon();
      |            ^~~~~~~~~~~~~~~~~~~~~
/home/chris/repos/jank/compiler+runtime/third-party/boost-multiprecision/include/boost/multiprecision/detail/default_ops.hpp:3796:4: note: Address of stack memory associated with temporary object of type 'detail::abs_funct<typename detail::backend_type<detail::expression<boost::multiprecision::detail::subtract_immediates, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100, int, void>, boost::multiprecision::et_on>, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100, int, void>, boost::multiprecision::et_on>, void, void>>::type>' is still referred to by a temporary object on the stack upon returning to the caller.  This will be a dangling reference
 3796 |    return detail::expression<
      |    ^
 3797 |        detail::function, detail::abs_funct<typename detail::backend_type<detail::expression<tag, A1, A2, A3, A4> >::type>, detail::expression<tag, A1, A2, A3, A4> >(
 3798 |        detail::abs_funct<typename detail::backend_type<detail::expression<tag, A1, A2, A3, A4> >::type>(), arg);
```

jeaye: Hm, which `abs` fn is being called there? I'm not finding the `abs` which takes a `native_big_decimal`.

djblue: I think it's calling this [abs template function from boost](https://github.com/boostorg/multiprecision/blob/boost-1.86.0/include/boost/multiprecision/detail/default_ops.hpp#L3790-L3799).

compiler+runtime/src/cpp/jank/read/lex.cpp
djblue: Removing this causes `undefined reference to jank::read::lex::big_decimal::operator==(jank::read::lex::big_decimal const&) const`.

compiler+runtime/src/cpp/jank/runtime/obj/big_integer.cpp
djblue: Removed.

compiler+runtime/include/cpp/jank/type.hpp
jeaye: Ok, it's possible that the `abs` fn is returning a `const&`, which is the problem. We could test this by moving the `l - r_as_decimal` into a local. Like so:

```cpp
auto const diff{ l - r_as_decimal };
return abs(diff) < std::numeric_limits<native_big_decimal>::epsilon();
```

If that fixes the issue, then it's because `abs` is returning an lvalue reference, but our input is a prvalue.

Aside from that, what was the change you made which fixed this on your end?

compiler+runtime/src/cpp/jank/read/lex.cpp
jeaye: You need to remove its declaration in the header too, brother!

compiler+runtime/include/cpp/jank/type.hpp
djblue: I tried the above, and still get the same analyzer warning. The only thing I needed to fix the issues is adding `boost::multiprecision::et_off`.

jeaye: Ok, gotcha. I read up on the docs for that. I think we can keep your change as is.

compiler+runtime/src/cpp/jank/read/lex.cpp
djblue: I get the following error when I remove it from the header:
```cpp
/home/chris/repos/jank/compiler+runtime/src/cpp/jank/read/lex.cpp:260:39: error: invalid operands to binary expression ('const std::variant<no_data, i64, f64, jtl::immutable_string_view, bool, ratio, big_integer, big_decimal>' (aka 'const variant<jank::read::lex::token::no_data, long long, double, jtl::immutable_string_view, bool, jank::read::lex::ratio, jank::read::lex::big_integer, jank::read::lex::big_decimal>') and 'const std::variant<no_data, i64, f64, jtl::immutable_string_view, bool, ratio, big_integer, big_decimal>')
  260 |       return kind == rhs.kind && data == rhs.data;
      |                                  ~~~~ ^  ~~~~~~~~
```

jeaye: Hm. Very odd, but at least we can define it with an `= default` instead of a manual implementation. The thing with those manual implementation is that it's _very_ easy to forget to update them when another member is added to the type.

djblue:
I think I addressed all the requested changes, and the CI failure feels unrelated, please let me know otherwise.

jeaye:
I found the issue and I've pushed a fix. We had a `fail-` test which was (mostly) reliably failing, but due to UB and not due to proper analysis. That's why, sometimes, it would not fail. That's all fixed now. If CI is green, I'll merge.

PR #524, opened 2025-09-25
Fix edge case for `clojure.string/trim`

djblue:
Fixes the following:
```clojure
% jank repl
nil
user=> (str/trim " ")
""
user=> (str/trim "  ")
GC Warning: Failed to expand heap by 18014398509481980 KiB
GC Warning: Out of Memory! Heap size: 21 MiB. Returning NULL!
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
zsh: IOT instruction (core dumped)  jank repl
```


jeaye (approved):
Thanks!

PR #538, opened 2025-10-09
Add clojure set namespace

cjbarre:


E-A-Griffin:
My guess is this won't work out of the box quite yet because certain `clojure.core` functions aren't yet implemented in `jank` (off the top of my head I'm guessing `persistent!` which is used here isn't implemented yet). @jeaye how do you normally handle functions/macros that depend on other functions/macros that aren't yet implemented? Should there be `;; TODO`s?

jeaye:
> My guess is this won't work out of the box quite yet because certain `clojure.core` functions aren't yet implemented in `jank` (off the top of my head I'm guessing `persistent!` which is used here isn't implemented yet). 

jank has support for `persistent!`. We have transients, in general.

> @jeaye how do you normally handle functions/macros that depend on other functions/macros that aren't yet implemented? Should there be `;; TODO`s?

If something isn't implemented yet, it should throw an exception saying it's not yet implemented and it should have a TODO comment there.

@cjbarre Have you loaded this file with jank to verify everything works well?

compiler+runtime/src/jank/clojure/set.jank
```diff
@@ -0,0 +1,179 @@
...
+(defn- bubble-max-key
+  "Move a maximal element of coll according to fn k (which returns a
+  number) to the front of coll."
+  [k coll]
+  (let [max (apply max-key k coll)]
+    (cons max (remove #(identical? max %) coll))))
+
+(defn union
+  "Return a set that is the union of the input sets"
+  {:added "1.0"}
```
jeaye: Please remove all of these meta maps. jank doesn't bother with `:added` or `:tag`.

compiler+runtime/src/jank/clojure/set.jank
```diff
@@ -0,0 +1,179 @@
...
+            {:a 3 :b 3 :c 3 :e 7}})
+
+  (join xs ys)
+  (join xs (rename ys {:b :yb :c :yc}) {:a :a})
+
+  (union #{:a :b :c} #{:c :d :e})
+  (difference #{:a :b :c} #{:c :d :e})
+  (intersection #{:a :b :c} #{:c :d :e})
+
+  (index ys [:b]))
```
jeaye: Please remove this, even if it came from upstream.

cjbarre:
@jeaye Yep, jank compiles, also loaded up in my project. All the functions appear to work.

PR #545, opened 2025-10-15
update code to address some clangd issues

jianlingzhong:
issues: Declaration shadows a field of 'jank::codegen::llvm_processor::impl' or a local variable

Also some cosmetic updates to comments.


jeaye:
Thanks for the PR, Jianling.

I'm cool with whatever the shadowing fix is, and grammar/spelling fixes in comments, but there are so many other changes here that I didn't even spot the shadowing fix. Renaming `fn` to `func` is not an improvement, especially since Clojure actually uses the term `fn` to mean function.

jianlingzhong:
@jeaye i reverted the change from `fn` to `func` (no shadowing found now). also, turns out that `typename` is indeed not needed. can you take another look?

jeaye (approved):
Thanks! I'll delete the `<ranges>` include when merging.

PR #565, opened 2025-10-24
Use LLVM/Clang from nixpkgs

kylc:
Upstream nixpkgs-unstable now has LLVM git from Oct 19, which includes the jank fixes.

Once LLVM 22 is released, we should just be able to replace the `_git` suffix with `_22`.

PR #567, opened 2025-10-25
Fix AOT on Nix

### LINKED ISSUE #515: Fix AOT builds on Nix
From here: https://github.com/jank-lang/jank/pull/502#issuecomment-3315159644

> @jeaye I have a prototype fix for the issues you're running into, but it's not suitable for merging. I won't have more time to look into this for a while, but hopefully this sets someone on the right path. https://github.com/kylc/jank/commit/2a046b80a85ae48106b08304b5f72bd67fa59968
> 
> I found two things which are causing the AOT build failures on Nix:
> 1. When the AOT subsystem invokes clang via the `clang::Driver` API, there is some translation done on the CLI args to put them into `-cc1` style. This is done before exec-ing the clang binary. So once the clang binary is executed, which on Nix points to the wrapper responsible for injecting all of the Nix-specific CFLAGS/LDFLAGS, the arguments are all in this clang-specific style. Nix specifically disables its wrapper functionality in this case since most of the common flags it injects won't be supported in this compiler mode, see [here](https://github.com/NixOS/nixpkgs/blob/ea26e8f23ffe674172ab68c46337b5543c7ccbbb/pkgs/build-support/cc-wrapper/cc-wrapper.sh#L221-L227). A hacky workaround for this is to manually exec the pre-translated clang compile command, rather than letting the driver do the translation before exec-ing clang.
>     - You can track this down by running `strace -f -e execve jank check-health` to see the exact clang command being invoked.
> 3. In the jank cmake, `JANK_DEPS_LIBRARY_DIRS` does not contain what I would expect. On my Nix system it only contains `/lib` (which is empty) rather than a bunch of `/nix/store/...` directories where all of the target dependency shared libs are stored. This means that when the AOT compiler is invoked, it effectively has no library search path and fails to find the libs. As another hacky workaround, I've just injected `NIX_LDFLAGS` into this var, which does contain the proper linker search paths.
> 
> With these changes the AOT stage of `jank check-health` now succeeds.

Patryk27:
This commit fixes AOT on Nix - the changes boil down to pulling LLVM from Nix (conflicts with #565!) and adding a small `preConfigure` hook that sort-of emulates Nix's cc-wrapper (the one that gets disabled by LLVM transparently adding `-cc1`):

```nix
preConfigure = ''
  export JANK_EXTRA_AOT_FLAGS=" \
      $(cat ${llvm.clang}/nix-support/cc-cflags) \
      $(cat ${llvm.clang}/nix-support/libc-crt1-cflags) \
      $(cat ${llvm.clang}/nix-support/cc-ldflags) \
      $NIX_LDFLAGS \
  "
'';
``` 

Special thanks to @kylc (https://github.com/jank-lang/jank/pull/502#issuecomment-3315159644)! 

Closes #515    
Closes #565


jeaye:
@kylc Would you mind giving a review?

compiler+runtime/CMakeLists.txt
```diff
@@ -587,6 +587,7 @@ foreach(lib ${jank_lib_link_dirs_prop})
   list(APPEND jank_lib_linker_flags_list -L${lib})
 endforeach()
 list(APPEND jank_lib_linker_flags_list ${jank_linker_flags})
+list(APPEND jank_lib_linker_flags_list $ENV{JANK_EXTRA_AOT_FLAGS})
```
kylc: Consider adding `jank_extra_aot_flags` as a regular CMake variable instead of depending on it from the environment, and then passing it as `cmake -Djank_extra_aot_flags=...`. That would be more consistent with the rest of the CMake.

flake.nix
```diff
@@ -49,36 +45,67 @@
             pname = "jank";
             version = "git";
 
-            # Add only essential files so that the source hash is consistent.
```
kylc, line 52: Lots of formatting and other small changes in the PR. You should run `nix fmt flake.nix` and avoid changing unrelated lines.

flake.nix
```diff
@@ -49,36 +45,67 @@
...
             '';
 
+            preConfigure = ''
+              export JANK_EXTRA_AOT_FLAGS=" \
+                  $(cat ${llvm.clang}/nix-support/cc-cflags) \
+                  $(cat ${llvm.clang}/nix-support/libc-crt1-cflags) \
+                  $(cat ${llvm.clang}/nix-support/cc-ldflags) \
+                  $NIX_LDFLAGS \
+              "
+            '';
```
kylc: Nice, this is a clean workaround to get the extra Nix flags.

Note that if you move the JANK_EXTRA_AOT_FLAGS from the ENV to a CMake argument, you could do something like:
```
jankAotFlags =
  lib.concatStringsSep " "
    [(lib.readFile "${llvm.clang}/nix-support/cc-cflags")
     ...];

cmakeFlags = [
  ...,
  (lib.cmakeFeature "jank_extra_aot_flags" jankAotFlags)
];
```

flake.nix
```diff
@@ -105,7 +133,7 @@
             cmake
             ninja
             pkg-config
-            llvm-jank
+            llvm.clang
```
kylc: Building with CMake in the devshell fails on my NixOS machine. I need these two packages in the devshell as well:
```
llvm.libclang
llvm.libllvm
```

To reproduce, try:
```
./bin/configure -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo && ./bin/compile
```

flake.nix
```diff
@@ -105,7 +133,7 @@
             cmake
```
kylc: Compiling in the devshell is missing the extra AOT flags var. You could either export it if you keep it as an env var, or if you add it as a CMake argument then you can wrap the CMake executable in the devshell to implicitly include it:
```
jankCmake = pkgs.writeShellScriptBin "cmake" ''
  exec ${pkgs.cmake}/bin/cmake -Djank_extra_aot_flags=... "$@"
'';

packages = [ jankCmake ... ];

```

kylc:
Also it seems like the nixpkgs LLVM is too large to build in our CI environment. I guess it compiles more modules and tests than our custom build. We could either:
- Wait for binaries to be available (presumably when LLVM 22 is released)
- Try to trim down LLVM with some override flags (`doCheck = false`?)
- @jeaye could push up a local build to the cache by doing something like `nix build . --print-out-paths | cachix push jank-lang`

Patryk27:
I've been playing around with LLVM for a while here and I think warming up the cache is the best way forward - reducing targets (to just `x86`) and disabling tests doesn't seem to affect most of the "overhead".

Regarding rest of the comments - thanks, I'll submit the improved version later today!

compiler+runtime/CMakeLists.txt
Patryk27: I've tried, but it seems CMake has some issues with newlines -- I've done:

```cmake
if(DEFINED jank_extra_aot_flags)
  list(APPEND jank_lib_linker_flags_list ${jank_extra_aot_flags})
endif()
```
... and then:
```nix
(lib.cmakeFeature "jank_extra_aot_flags" jankExtraAotFlags) # also tried with lib.escapeShellArg
```
... and for the love of everything, I couldn't get it to work:
```
       > cmake flags: -GNinja -DCMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY=OFF -DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_LOCALEDIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/share/locale -DCMAKE_INSTALL_LIBEXECDIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/libexec -DCMAKE_INSTALL_LIBDIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/lib -DCMAKE_INSTALL_DOCDIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/share/doc/jank -DCMAKE_INSTALL_INFODIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/share/info -DCMAKE_INSTALL_MANDIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/share/man -DCMAKE_INSTALL_INCLUDEDIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/include -DCMAKE_INSTALL_SBINDIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/sbin -DCMAKE_INSTALL_BINDIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/bin -DCMAKE_INSTALL_NAME_DIR=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git/lib -DCMAKE_POLICY_DEFAULT_CMP0025=NEW -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_STRIP=/nix/store/x8mydcgbry214s802nzvy7fdljx404ym-gcc-wrapper-14.3.0/bin/strip -DCMAKE_RANLIB=/nix/store/x8mydcgbry214s802nzvy7fdljx404ym-gcc-wrapper-14.3.0/bin/ranlib -DCMAKE_AR=/nix/store/x8mydcgbry214s802nzvy7fdljx404ym-gcc-wrapper-14.3.0/bin/ar -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_INSTALL_PREFIX=/nix/store/wrlfycqf216sh22iajwmzwh99hz93gwi-jank-git -DCMAKE_C_COMPILER=/nix/store/9qpsh2fyh9hiabdbyn5mmwxi44vx8njh-clang-wrapper-22.0.0-unstable-2025-10-19/bin/clang -DCMAKE_CXX_COMPILER=/nix/store/9qpsh2fyh9hiabdbyn5mmwxi44vx8njh-clang-wrapper-22.0.0-unstable-2025-10-19/bin/clang++ -DCMAKE_SKIP_RPATH=ON -DFETCHCONTENT_SOURCE_DIR_LIBDWARF=/nix/store/ksad80kbisl4n76qwjc2k3ig8zcrn7fs-source -DFETCHCONTENT_SOURCE_DIR_ZSTD=/nix/store/w0fyq7vdpja3yawg943an3zr644cqx1v-source -Djank_unity_build:BOOL=TRUE -Djank_test:BOOL=TRUE -Djank_extra_aot_flags:STRING=\'-B/nix/store/ffrg0560kj0066s4k9pznjand907nlnz-gcc-14.3.0/lib/gcc/x86_64-unknown-linux-gnu/14.3.0 --gcc-toolchain=/nix/store/ffrg0560kj0066s4k9pznjand907nlnz-gcc-14.3.0 -B/nix/store/likdwk5nl7w6i8d50rxcfmmak3bcl18q-clang-22.0.0-unstable-2025-10-19-lib/lib -nostdlibinc -resource-dir=/nix/store/9qpsh2fyh9hiabdbyn5mmwxi44vx8njh-clang-wrapper-22.0.0-unstable-2025-10-19/resource-root -B/nix/store/daamdpmaz2vjvna55ccrc30qw3qb8h6d-glibc-2.40-66/lib/ -L/nix/store/ffrg0560kj0066s4k9pznjand907nlnz-gcc-14.3.0/lib/gcc/x86_64-unknown-linux-gnu/14.3.0 -L/nix/store/z7a34j3xnp66rpddayyxrxwsahxccbip-gcc-14.3.0-lib/lib -L/nix/store/ffrg0560kj0066s4k9pznjand907nlnz-gcc-14.3.0//lib -L/nix/store/mnd897flngivhxnw8sxznhrcwwc9vyk3-gcc-14.3.0-libgcc/lib -L/nix/store/likdwk5nl7w6i8d50rxcfmmak3bcl18q-clang-22.0.0-unstable-2025-10-19-lib/lib \'
       > CMake Warning:
       >   Ignoring extra path from command line:
       >
       >    "'"
       >
       > 
       > CMake Error: Unknown argument -L/nix/store/likdwk5nl7w6i8d50rxcfmmak3bcl18q-clang-22.0.0-unstable-2025-10-19-lib/lib
       > CMake Error: Run 'cmake --help' for all supported options.
```

flake.nix
Patryk27: kk, formatting changes rolled back

flake.nix
Patryk27: Sure, that sounds better - although note that this requires "emulating" the `$NIX_LDFLAGS` part as well:
```nix
preConfigure = ''
  export JANK_EXTRA_AOT_FLAGS=" \
      $(cat ${llvm.clang}/nix-support/cc-cflags) \
      $(cat ${llvm.clang}/nix-support/libc-crt1-cflags) \
      $(cat ${llvm.clang}/nix-support/cc-ldflags) \
      $NIX_LDFLAGS \  <-- this boi
  "
'';
```
... making the final expression somewhat long, because we have to consider `buildInputs` as well:
```nix
jankExtraAotFlags =
  lib.concatStringsSep " "
  [
    (lib.trim (lib.readFile "${llvm.clang}/nix-support/cc-cflags"))
    (lib.trim (lib.readFile "${llvm.clang}/nix-support/libc-crt1-cflags"))
    (lib.trim (lib.readFile "${llvm.clang}/nix-support/cc-ldflags"))
    "-L${lib.getLib llvm.libllvm.lib}/lib"
    "-L${lib.getLib pkgs.bzip2}/lib"
    "-L${lib.getLib pkgs.libzip}/lib"
    "-L${lib.getLib pkgs.openssl}/lib"
    "-L${lib.getLib pkgs.zlib}/lib"
    "-L${lib.getLib pkgs.zstd}/lib"
  ];
```
Still, I think that's better than relying on everything from `$NIX_LDFLAGS`, because that might pull extra changes we don't need (e.g. `-rpath` or whatever).

flake.nix
Patryk27: kk, fixed!

flake.nix
Patryk27: I've kept the envvar approach (https://github.com/jank-lang/jank/pull/567#discussion_r2463784855) - now it gets correctly passed to the dev-shell.

jeaye:
Thanks, guys. I am getting that cache warmed now, before triggering CI again.

jeaye:
Ok, cache has been populated, but looks like the health check fails when testing AOT building!

Patryk27:
Hah, interesting failure indeed - while it works on my machine™️, I do have some ideas what might be going on there, 🔍

Patryk27:
Ok, managed to reproduce it on Ubuntu (with Nix installed on top of it) - I think the final issue is that we don't specify the dynamic linker, but doing the obvious thing:
```nix
        jankExtraAotFlags =
          lib.concatStringsSep " "
          [
            "-Wl,-dynamic-linker=${pkgs.stdenv.cc.bintools.dynamicLinker}"
            # ...
          ];
```
... also doesn't seem to help, needs a bit more elbow grease.

Patryk27:
Ok, got it! -- the missing incantation seems to be:

```
 "-Wl,-rpath,${pkgs.stdenv.cc.libc}/lib"
```

... as in here:

https://github.com/NixOS/nixpkgs/blob/757b062770997b90b71d16b6ad50cf1dde2b5241/pkgs/development/libraries/glibc/common.nix#L222

With that, the AOT works on both NixOS and Ubuntu (checked in a VM).

compiler+runtime/CMakeLists.txt
kylc: I ran into a similar issue before; it's very confusing behavior. You can find a workaround here: https://discourse.nixos.org/t/cmakeflags-and-spaces-in-option-values/20170/2

kylc:
Nice work, LGTM!

jeaye:
Badass! I just updated it with main locally and removed libzip, which is no longer a dep. cachix should be warmed for this, so we'll run CI and if it's green we're golden.

Excellent work, Patryk and Kyle!

compiler+runtime/CMakeLists.txt
Patryk27: Ah, that's nice - I'll take a stab at that tomorrow, thanks!

jeaye:
image

We're golden!

compiler+runtime/CMakeLists.txt
jeaye: jank adds `CMAKE_CXX_FLAGS` to the build command for the PCH, but we don't do the same for AOT flags. Seems like that should be all we need, and it should *just work* for any system which already supports adding the necessary CMake bits.

Edit: meaning that we may need to set some extra `CMAKE_CXX_FLAGS` for Nix, but we'd have better out-of-the-box support on other platforms, too, without them needing to add their platform-specific flags in a jank-specific way.

Patryk27: `CMAKE_CXX_FLAGS` seems to be a nice approach indeed, let me know what you think now!

compiler+runtime/CMakeLists.txt
```diff
@@ -612,6 +612,7 @@ foreach(lib ${jank_lib_link_dirs_prop})
   list(APPEND jank_lib_linker_flags_list -L${lib})
 endforeach()
 list(APPEND jank_lib_linker_flags_list ${jank_linker_flags})
+list(APPEND jank_lib_linker_flags_list ${CMAKE_CXX_FLAGS})
```
Patryk27, line 615: oor maybe `separate_arguments(jank_lib_linker_flags_list UNIX_COMMAND ${CMAKE_CXX_FLAGS})` in place of the `set(jank_lib_linker_flags_list "")` above? (not a cmake expert, no idea if that's tangibly different / better)

jeaye: I think this is just fine.

compiler+runtime/CMakeLists.txt
jeaye: Yep, much better. I have the `JANK_EXTRA_*` env vars as escape hatches, but the goal is to not have people rely on them for distribution if we can avoid it.

jeaye:
Anything else you'd like to do with this PR before merging?

Patryk27:
Nah, I'm all set!

jeaye:
Nice work!

PR #568, opened 2025-10-26
Allow re-indexing module path on entry miss

djblue:
Since source files can be added to the module-path after the jank process starts, we need to reindex the module-path on failures to ensure a module is still missing.

```clojure
% jank repl --module-path="."
user=> (require 'hello)
Uncaught exception: unable to find module: hello
user=> (spit "hello.jank" "(println ::hi)")
nil
user=> (require 'hello)
:user/hi
nil
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
@@ -301,7 +301,7 @@ namespace jank::runtime::module
   static void
   register_module_path(native_unordered_map<jtl::immutable_string, loader::entry> &entries,
                        jtl::immutable_string const &paths,
-                       bool skip_jar = false)
+                       bool const skip_jar)
   {
     usize start{};
     usize i{ paths.find(loader::module_separator, start) };
@@ -356,7 +356,7 @@ namespace jank::runtime::module
     this->paths = paths;
 
     //util::println("module paths: {}", paths);
-    register_module_path(entries, paths);
+    register_module_path(entries, paths, false);
   }
 
   object_ref file_entry::to_runtime_data() const
@@ -617,8 +617,8 @@ namespace jank::runtime::module
       return first_find->second;
     }
 
-    // Skip jars after the initial indexing since we can consider them immutable
-    // for development purposes.
+    /* Skip jars after the initial indexing since we can consider them immutable
+     * for development purposes. */
     register_module_path(entries, paths, true);
 
     auto const &second_find(entries.find(module));
@@ -644,7 +644,7 @@ namespace jank::runtime::module
       return err(util::format("unable to find module: {}", module));
     }
 
-    auto const entry(found.unwrap());
+    auto const &entry(found.unwrap());
 
     if(ori == origin::source)
     {


compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -298,6 +298,45 @@ namespace jank::runtime::module
...
+        register_path(entries, paths);
+      }
+    }
+    else
+    {
+      while(i != jtl::immutable_string::npos)
+      {
+        auto const path(paths.substr(start, i - start));
+
+        if(!skip_jar || !path.ends_with(".jar"))
```
jeaye, line 323: This will prevent new jars from being added to the path, right? That prevents the use case of adding a new dependency at runtime.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -298,6 +298,45 @@ namespace jank::runtime::module
     }
   }
 
+  static void
+  register_module_path(native_unordered_map<jtl::immutable_string, loader::entry> &entries,
+                       jtl::immutable_string const &paths,
+                       bool skip_jar = false)
```
jeaye: Please don't use default params. Once one is added, everything after it needs to be defaulted as well. Either use an overload which doesn't take the param and then calls this one or just require the param on all calls.

Also, please make `skip_jar` const.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -586,27 +605,56 @@ namespace jank::runtime::module
...
-    auto const &entry(entries.find(patched_module));
-    if(entry == entries.end())
+    auto const &found(find_module(entries, paths, patched_module));
+
+    if(found.is_none())
     {
       return err(util::format("unable to find module: {}", module));
     }
 
+    auto const entry(found.unwrap());
```
jeaye: No need for a deep copy here, which will involve copying eight different strings. Please use `auto const &` instead.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -586,27 +605,56 @@ namespace jank::runtime::module
...
+  {
+    auto const &first_find(entries.find(module));
+
+    if(first_find != entries.end())
+    {
+      return first_find->second;
+    }
+
+    // Skip jars after the initial indexing since we can consider them immutable
+    // for development purposes.
```
jeaye: Please use `/* */` style comments.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
djblue: I think so, but I assume that adding a jar will be an explicit action from the user, somewhat like [`clojure.repl.deps/add-lib`](https://clojuredocs.org/clojure.repl.deps/add-lib). In which case, we would know the exact artifact to index and we could skip all the guess work.

jeaye:
Nice work, Chris!

PR #569, opened 2025-10-26
Switch to large code model

shantanu-sardesai:
While building the prototypes for the Macroexpand deep conf. I ended up out reaching the LLVM code model in which case the following changes were suggested.

This might be too simple a change I've made, but keeping this PR open so that what all things are actually required for it will be made.


jeaye:
Thanks!

PR #570, opened 2025-10-29
Incremental porting

shantanu-sardesai:
There will be a change required for jank in the `bigdec` test case in the Clojure test suite, I'll open a PR for it once this one is merged.


shantanu-sardesai:
@jeaye 
Shall I remove my [Track test status](https://github.com/jank-lang/jank/pull/570/commits/9429213a9316e1adfc8b0c0738f41d79584017a0) commit from this PR?

compiler+runtime/src/jank/clojure/core.jank
shantanu-sardesai, line 5433: Is there any benefit in converting this to:

```jank
(def decimal? (fn* [n] (cpp/jank.runtime.is_big_decimal n)))
```

And same for `bigdec`?

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1774,4 +1774,37 @@ namespace jank::runtime
       throw make_box(util::format("Expected string, got {}", object_type_str(o->type))).erase();
     }
   }
+
+  bool is_big_decimal(object_ref const o)
+  {
+    return o->type == object_type::big_decimal;
+  }
+
+  object_ref to_big_decimal(object_ref o)
```
jeaye: `o` can be `const`.

compiler+runtime/include/cpp/jank/runtime/core/math.hpp
```diff
@@ -277,4 +277,7 @@ namespace jank::runtime
 
   i64 parse_long(object_ref o);
   f64 parse_double(object_ref o);
+
+  bool is_big_decimal(object_ref o);
+  object_ref to_big_decimal(object_ref o);
```
jeaye: We can return a `big_decimal_ref` here instead, so we actually have type information to use for optimizations.

compiler+runtime/src/cpp/jank/runtime/obj/big_decimal.cpp
```diff
@@ -124,6 +124,26 @@ namespace jank::runtime::obj
...
 
+  big_decimal::big_decimal(native_big_integer const &val)
+    : data{ val.real() }
+  {
+  }
+
+  big_decimal::big_decimal(native_big_integer &&val)
+    : data{ val.real() }
+  {
+  }
```
jeaye: There is no point in taking an rvalue reference if there's nothing to move and if you don't even call `std::move`. Both of these rvalue ref ctors should not exist.

compiler+runtime/include/cpp/jank/runtime/obj/big_decimal.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <jank/runtime/object.hpp>
+#include <jank/runtime/obj/ratio.hpp>
```
jeaye: There is no need to include this here, since we're only using `ratio` as a reference. Please forward declare, instead. Always keep in mind the cost of headers.

jeaye (changes requested):
Thanks, Shantanu!

shantanu-sardesai:
This [PR](https://github.com/jank-lang/clojure-test-suite/pull/772) needs to be merged and we need to repoint the Clojure test suite for the CI to pass.

jeaye:
> This [PR](https://github.com/jank-lang/clojure-test-suite/pull/772) needs to be merged and we need to repoint the Clojure test suite for the CI to pass.

Merged. Ready for your repoint.

shantanu-sardesai:
Done.

Issue #479, opened 2025-09-11
Add `#cpp` support for integers, reals, and bools


jeaye:
This should be very straightforward, given our existing support for `#cpp "foo"`. However, one trick is that we'll want to explicitly cast the value to be a jank type, rather than the builtin C++ type (`int` and `double` respectively).

* `#cpp 100` => `(cpp/value "static_cast<jank::i64>(100)")`
* `#cpp 100.5` => `(cpp/value "static_cast<jank::f64>(100.5)")`
* `#cpp true` => `(cpp/value "true")`

PR #571, opened 2025-10-29
Extend `#cpp` tagged literal

shantanu-sardesai:
Resolves https://github.com/jank-lang/jank/issues/479

### DIFF
--- a/compiler+runtime/test/jank/reader-macro/cpp/real/pass-real.jank
+++ b/compiler+runtime/test/jank/reader-macro/cpp/real/pass-real.jank
@@ -1,5 +1,3 @@
 (let* [f #cpp 4.2]
   (if (cpp/== (cpp/value "4.2") f)
     :success))
-
-:success


compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -937,19 +937,47 @@ namespace jank::read::parse
...
+#pragma clang diagnostic ignored "-Wswitch-enum"
+    switch(str_end.kind)
     {
-      return error::parse_invalid_reader_tag_value(
-        "The form after '#cpp' must be a string literal.",
-        { start_token.start, latest_token.end });
+      case lex::token_kind::string:
+      case lex::token_kind::escaped_string:
+        {
+          auto string{ expect_object<obj::persistent_string>(str_result.expect_ok().unwrap().ptr) };
```
jeaye: All of these `auto` and be `auto const`.

compiler+runtime/test/jank/reader-macro/cpp/boolean/pass-boolean.jank
```diff
@@ -0,0 +1,4 @@
+#cpp true
```
jeaye: All of these tests don't actually check the *usability* of the values. As such, they provide very little value. Yes, you copied the string tests, but those were poor to start with.

Put the value in a `let*` and only *conditionally* evaluate to `:success` if everything is correct. There are _many_ examples of this in this test suite.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/read/parse.cpp
shantanu-sardesai: I'm **_still_** blind to these 🙃.

compiler+runtime/test/jank/reader-macro/cpp/real/pass-real.jank
```diff
@@ -1,3 +1,5 @@
-#cpp 4.2
+(let* [f #cpp 4.2]
+  (if (cpp/== (cpp/value "4.2") f)
+    :success))
 
 :success
```
jeaye: I don't think we want this extra `:success`.

jeaye (changes requested):
One bit missing. Very close.

shantanu-sardesai:
Done.

jeaye:
Nice work! No more features, though. Stability time!

Issue #271, opened 2025-02-16
Incorrect error message when passing non-metadatable to `with-meta`


frenchy64:
The meta is blamed instead of the object. Similar problem with metadata mutation.

```clojure
clojure.core=> (with-meta 1 {})
Exception: not metadatable: {}
```

PR #572, opened 2025-10-30
Improve error reporting for non-metadatable values

shantanu-sardesai:
Resolves https://github.com/jank-lang/jank/issues/271

### DIFF
--- a/compiler+runtime/src/cpp/jank/error/runtime.cpp
+++ b/compiler+runtime/src/cpp/jank/error/runtime.cpp
@@ -26,10 +26,4 @@ namespace jank::error
   {
     return make_error(kind::runtime_non_metadatable_value, message, source);
   }
-
-  error_ref
-  internal_runtime_failure(jtl::immutable_string const &message, read::source const &source)
-  {
-    return make_error(kind::internal_runtime_failure, message, source);
-  }
 }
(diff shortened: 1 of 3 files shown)


compiler+runtime/src/cpp/jank/runtime/core/meta.cpp
```diff
@@ -1,3 +1,4 @@
+#include "jank/error/runtime.hpp"
```
jeaye: Please use `<>` for includes. Also, put this at the bottom, not at the top of the include list.

compiler+runtime/src/cpp/jank/error/runtime.cpp
```diff
@@ -20,4 +20,16 @@ namespace jank::error
...
+  runtime_non_metadatable_value(jtl::immutable_string const &message, read::source const &source)
+  {
+    return make_error(kind::runtime_non_metadatable_value, message, source);
+  }
+
+  error_ref
+  internal_runtime_failure(jtl::immutable_string const &message, read::source const &source)
+  {
+    return make_error(kind::internal_runtime_failure, message, source);
+  }
```
jeaye: Looks like you added this but you're not using it. Let's remove it. We don't know the signature we'll want for this, when we do need it, but it almost certainly won't include a `source` param (since internal errors very rarely have source).

jeaye (changes requested):
Thanks, Shantanu!

compiler+runtime/src/cpp/jank/runtime/core/meta.cpp
shantanu-sardesai: Dang.

Issue #276, opened 2025-02-17
`empty` should preserve metadata to align with Clojure


frenchy64:
```jank
;; jank
clojure.core=> (meta (empty (with-meta [] {:foo 1})))
nil

;; clojure
user=> (meta (empty (with-meta [] {:foo 1})))
{:foo 1}
```

jeaye:
Related to some work @djblue was doing.

PR #573, opened 2025-10-30
Preserve metadata across `empty` call

shantanu-sardesai:
Resolves https://github.com/jank-lang/jank/issues/276.

```jank
user=> (meta (empty (with-meta [] {:foo 1})))
{:foo 1}

user=> (meta (empty (with-meta #{} {:foo 1})))
{:foo 1}

user=> (meta (empty (with-meta {} {:foo 1})))
{:foo 1}

user=> (meta (empty (with-meta '() {:foo 1})))
{:foo 1}
```


jeaye:
Let's couple this with a test in the clojure-test-suite to prove this works. We're not testing for meta preservation, but we should be.

shantanu-sardesai:
I'll create a PR in the Clojure test suite 👍.

shantanu-sardesai:
[Done](https://github.com/jank-lang/clojure-test-suite/pull/775). Although still thinking about what all other test cases that can be added.

jeaye:
So I think this just needs a repoint for clojure-test-suite now.

shantanu-sardesai:
Done. Although the `empty` test is commented out on the jank side with the reason for it. I could fix it here but wanted to keep the PR focused on the ticket.

jeaye:
Yeah, that's no problem. We're just doing what we can for now. You did the right thing.

PR #576, opened 2025-11-01
Start analyzing & evaluating `meta`

### LINKED ISSUE #197: `def` meta should be evaluated and support arbitrary values

shantanu-sardesai:
### Description

We should begin analyzing the metadata as well. At present, we store the metadata source directly without proper analysis or evaluation, resulting in the following code executing despite being semantically incorrect:

```jank
user=> (def ^{:pp #(println foo)} bar 0)
#'user/bar

user=> (meta bar)
nil

user=> (meta #'bar)
{:jank/source {:file "/var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-fYkERF", :start {:offset 27, :line 1, :col 28}, :end {:offset 30, :line 1, :col 31}}, :pp (fn* [] (println foo))}

user=> ((-> #'bar meta :pp))
Uncaught exception: Invalid call with 0 args to: (fn* [] (println foo))
```

There are two issues being highlighted here:
1. Although the `foo` symbol is not in scope during the definition of the `bar` var, jank does not emit any analysis failure warnings.
2. Due to lack of evaluation currently the `:pp` key in the metadata of the `bar` var contains the source code instead of a compiled function, preventing us from invoking the function.

### Goal

Within the scope of this PR I wish to start analyzing the metadata such that the above cases behave as would be expected:

```jank
user=> (def ^{:pp #(println foo)} bar 0)
─ analyze/unresolved-symbol ────────────────────────────────────────────────────────────────────────
error: Unable to resolve symbol 'foo'.

     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-hKkVT2
  1  │ (def ^{:pp #(println foo)} bar 0)
     │                      ^^^ Found here.

user=> (def foo 0)
#'user/foo

user=> (def ^{:pp #(println foo)} bar 0)
#'user/bar

user=> (meta #'bar)
{:jank/source {:file "/var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-hKkVT2", :start {:offset 27, :line 1, :col 28}, :end {:offset 30, :line 1, :col 31}}, :pp #object [user/user-fn-53 jit_function 0x10ec383f8]}

user=> ((-> #'bar meta :pp))
0
nil
```

---

### Fixes

Resolves https://github.com/jank-lang/jank/issues/197.

### DIFF
--- a/compiler+runtime/src/cpp/jank/analyze/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/analyze/processor.cpp
@@ -1335,28 +1335,28 @@ namespace jank::analyze
       qualified_sym = qualified_sym->with_meta(meta_with_doc);
     }
 
+    jtl::option<expression_ref> meta{};
+
     /* Lift this so it can be used during codegen. */
     /* TODO: I don't think lifting meta is actually needed anymore. Verify. */
     if(qualified_sym->meta.is_some())
     {
       current_frame->lift_constant(qualified_sym->meta.unwrap());
+      auto const meta_result{ analyze(qualified_sym->meta.unwrap(),
+                                      current_frame,
+                                      expression_position::value,
+                                      fn_ctx,
+                                      true) };
+
+      if(meta_result.is_err())
+      {
+        return meta_result;
+      }
+
+      meta = meta_result.expect_ok();
     }
 
-    auto const meta{
-      analyze(qualified_sym->meta.unwrap(), current_frame, expression_position::value, fn_ctx, true)
-    };
-
-    if(meta.is_err())
-    {
-      return meta;
-    }
-
-    return jtl::make_ref<expr::def>(position,
-                                    current_frame,
-                                    true,
-                                    qualified_sym,
-                                    meta.expect_ok(),
-                                    value_expr);
+    return jtl::make_ref<expr::def>(position, current_frame, true, qualified_sym, meta, value_expr);
   }
 
   processor::expression_result
--- a/compiler+runtime/src/cpp/jank/analyze/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/analyze/processor.cpp
@@ -1342,7 +1342,21 @@ namespace jank::analyze
       current_frame->lift_constant(qualified_sym->meta.unwrap());
     }
 
-    return jtl::make_ref<expr::def>(position, current_frame, true, qualified_sym, value_expr);
+    auto const meta{
+      analyze(qualified_sym->meta.unwrap(), current_frame, expression_position::value, fn_ctx, true)
+    };
+
+    if(meta.is_err())
+    {
+      return meta;
+    }
+
+    return jtl::make_ref<expr::def>(position,
+                                    current_frame,
+                                    true,
+                                    qualified_sym,
+                                    meta.expect_ok(),
+                                    value_expr);
   }
 
   processor::expression_result
--- a/compiler+runtime/src/cpp/jank/evaluate.cpp
+++ b/compiler+runtime/src/cpp/jank/evaluate.cpp
@@ -276,7 +276,7 @@ namespace jank::evaluate
   object_ref eval(expr::def_ref const expr)
   {
     auto var(__rt_ctx->intern_var(expr->name).expect_ok());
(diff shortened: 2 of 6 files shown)


shantanu-sardesai:
Currently the following doesn't work:

```jank
(let* [foo 42]
  (def ^{:pp #(println foo)} bar 0)
  (assert (= ((-> #'bar meta :pp)) 42)))
```

But before I dive deeper I thought best align on where to store the analysis results for `meta`.

compiler+runtime/test/jank/form/def/pass-meta.jank
```diff
@@ -8,4 +8,14 @@
 (assert (= one 'one))
 (assert (= (:doc (meta #'one)) nil))
 
+; Analysis with global variables
```
jeaye, line 11: This is not how new tests should be added. Each test should have a dedicated test file, describing what it's testing. Also, we now have possibilities for extra errors, which come from analyzing meta. That needs to be tested.

compiler+runtime/src/cpp/jank/evaluate.cpp
```diff
@@ -276,7 +276,7 @@ namespace jank::evaluate
   object_ref eval(expr::def_ref const expr)
   {
     auto var(__rt_ctx->intern_var(expr->name).expect_ok());
-    var->meta = expr->name->meta;
+    var->meta = some(eval(expr->meta.unwrap()));
```
jeaye: This is lazy and not safe (because we don't know that `meta` contains a value). Look at line 281 for not only the variable you should use, but how to safely grab this value.

compiler+runtime/test/jank/form/def/pass-meta.jank
```diff
@@ -8,4 +8,14 @@
...
+(def foo 42)
+(def ^{:pp #(println foo)} bar 0)
```
jeaye, line 13: This will fail if you wrap it in a function or `let*`, since IR gen has not been updated.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -1342,7 +1342,21 @@ namespace jank::analyze
       current_frame->lift_constant(qualified_sym->meta.unwrap());
     }
 
-    return jtl::make_ref<expr::def>(position, current_frame, true, qualified_sym, value_expr);
+    auto const meta{
+      analyze(qualified_sym->meta.unwrap(), current_frame, expression_position::value, fn_ctx, true)
```
jeaye: This unwrap is not safe.

jeaye (changes requested):
Good start, Shantanu. Your overall `unwrap` usage is lacking a lot of necessary discipline. The testing strategy will need some more detailed planning, too, since this opens up a lot of doors. Finally, IR gen has not been updated, so none of this will work when wrapped in a function or AOT compiled.

compiler+runtime/src/cpp/jank/evaluate.cpp
shantanu-sardesai: I've chosen to wrap the `unwrap` in an `is_some` block so that I won't even bother calling `eval` when there is no `meta` available.

jeaye:
@shantanu-sardesai Following up on this so it doesn't get lost. What's the status? Do you need anything from me?

jeaye:
Closing this since it's out of date, incomplete, and conflicting. We still need this feature, but let's start a new PR once we work on it.

Issue #204, opened 2025-01-30
Keyword literals should not be empty


frenchy64:
```clojure
clojure.core=> (:)
Exception: invalid call with 0 args to: :
clojure.core=> '(:)
(:)
clojure.core=> '(::)
(:clojure.core/)
```

shantanu-sardesai:
The first two cases are being handled gracefully by jank already while the last one causes a crash (will fix that):

```jank
user=> (:)
─ lex/invalid-keyword ──────────────────────────────────────────────────────────────────────────────
error: A keyword must contain a valid symbol after the ':'.

─────┬──────────────────────────────────────────────────────────────────────────────────────────────
     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-3acEX3
─────┼──────────────────────────────────────────────────────────────────────────────────────────────
  1  │ (:)
     │  ^ Found here.
─────┴──────────────────────────────────────────────────────────────────────────────────────────────

user=> '(:)
─ lex/invalid-keyword ──────────────────────────────────────────────────────────────────────────────
error: A keyword must contain a valid symbol after the ':'.

─────┬──────────────────────────────────────────────────────────────────────────────────────────────
     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-3acEX3
─────┼──────────────────────────────────────────────────────────────────────────────────────────────
  1  │ '(:)
     │   ^ Found here.
─────┴──────────────────────────────────────────────────────────────────────────────────────────────

user=> '(::)
Assertion failed! index < len
Stack trace (most recent call first):
#0  0x00000001004d57bb in jtl::do_assertion_panic(jtl::immutable_string_view const&) at /Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/build/jank-phase-1
#1  0x00000001005049e3 in jtl::immutable_string_view::operator[](unsigned long long) const at /Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/build/jank-phase-1
#2  0x00000001005f66cb in jank::read::lex::processor::next() at /Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/build/jank-phase-1
#3  0x00000001005f224b in jank::read::lex::processor::iterator::operator++() at /Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/build/jank-phase-1
[...]

PR #577, opened 2025-11-01
Improve `keyword` parsing

shantanu-sardesai:
Resolves https://github.com/jank-lang/jank/issues/204, https://github.com/jank-lang/jank/issues/578.

Now jank also handles the auto-resolved qualified keywords properly:

```jank
user=> '(::)
─ lex/invalid-keyword ──────────────────────────────────────────────────────────────────────────────
error: An auto-resolved keyword may not start with ')'.

     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-Q7azuw
  1  │ '(::)
     │   ^ Found here.
```

### DIFF
--- a/compiler+runtime/test/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/test/cpp/jank/read/lex.cpp
@@ -5,7 +5,6 @@
 #include <jank/runtime/context.hpp>
 #include <jank/runtime/core/to_string.hpp>
 #include <jank/runtime/obj/nil.hpp>
-#include <jank/util/fmt/print.hpp>
 
 /* This must go last; doctest and glog both define CHECK and family. */
 #include <doctest/doctest.h>
@@ -570,7 +569,7 @@ namespace jank::read::lex
         native_vector<jtl::result<token, error_ref>> const tokens(p.begin(), p.end());
         CHECK(tokens
               == make_results({ { make_error(kind::lex_invalid_ratio, 0, 3) },
-                                { make_error(kind::lex_invalid_symbol, 3, 3) } }));
+                                { make_error(kind::lex_invalid_symbol, 3, 2) } }));
       }
       SUBCASE("Failures - x/x.x")
       {
@@ -584,7 +583,7 @@ namespace jank::read::lex
         native_vector<jtl::result<token, error_ref>> const tokens(p.begin(), p.end());
         CHECK(tokens
               == make_results({ make_error(kind::lex_invalid_ratio, 0, 3),
-                                make_error(kind::lex_invalid_symbol, 3, 3) }));
+                                make_error(kind::lex_invalid_symbol, 3, 2) }));
       }
       SUBCASE("Failures - x/xex")
       {
@@ -651,8 +650,8 @@ namespace jank::read::lex
         CHECK(tokens
               == make_results({
                 make_error(kind::lex_invalid_number, 0, 2),
-                make_error(kind::lex_invalid_number, 3, 5),
-                make_error(kind::lex_invalid_number, 6, 9),
+                make_error(kind::lex_invalid_number, 3, 2),
+                make_error(kind::lex_invalid_number, 6, 3),
               }));
       }
 
@@ -676,10 +675,10 @@ namespace jank::read::lex
         CHECK(tokens
               == make_results({
                 make_error(kind::lex_invalid_number, 0, 6),
-                make_error(kind::lex_invalid_number, 7, 12),
-                make_error(kind::lex_invalid_number, 13, 16),
-                make_error(kind::lex_invalid_number, 17, 21),
-                make_error(kind::lex_expecting_whitespace, 21, 21),
+                make_error(kind::lex_invalid_number, 7, 5),
+                make_error(kind::lex_invalid_number, 13, 3),
+                make_error(kind::lex_invalid_number, 17, 4),
+                make_error(kind::lex_expecting_whitespace, 21, 0),
                 token{ 21, 2, token_kind::symbol, ".2"sv },
         }));
       }
@@ -729,10 +728,10 @@ namespace jank::read::lex
         CHECK(tokens
               == make_results({
                 token{  0, 3, token_kind::integer, 123ll },
-                make_error(kind::lex_invalid_number, 4, 8),
+                make_error(kind::lex_invalid_number, 4, 4),
                 token{  9, 3, token_kind::integer,   7ll },
                 token{ 13, 3, token_kind::integer, -12ll },
-                make_error(kind::lex_invalid_number, 17, 22),
+                make_error(kind::lex_invalid_number, 17, 5),
                 token{ 23, 5, token_kind::integer, 255ll }
         }));
       }
@@ -771,11 +770,11 @@ namespace jank::read::lex
         native_vector<jtl::result<token, error_ref>> const tokens(p.begin(), p.end());
         CHECK(tokens
               == make_results({ make_error(kind::lex_invalid_number, 0, 3),
-                                make_error(kind::lex_invalid_number, 4, 8),
-                                make_error(kind::lex_invalid_number, 9, 14),
-                                make_error(kind::lex_invalid_number, 15, 21),
-                                make_error(kind::lex_invalid_number, 22, 24),
-                                make_error(kind::lex_invalid_number, 25, 27) }));
+                                make_error(kind::lex_invalid_number, 4, 4),
+                                make_error(kind::lex_invalid_number, 9, 5),
+                                make_error(kind::lex_invalid_number, 15, 6),
+                                make_error(kind::lex_invalid_number, 22, 2),
+                                make_error(kind::lex_invalid_number, 25, 2) }));
       }
 
       SUBCASE("Invalid hex - 1x1")
@@ -795,7 +794,7 @@ namespace jank::read::lex
         CHECK(tokens
               == make_results({
                 token{ 0, 3,   token_kind::real,    2.0 },
-                make_error(kind::lex_invalid_number, 3, 3),
+                make_error(kind::lex_expecting_whitespace, 3, 0),
                 token{ 3, 2, token_kind::symbol, "x1"sv },
         }));
       }
@@ -853,11 +852,11 @@ namespace jank::read::lex
         native_vector<jtl::result<token, error_ref>> const tokens(p.begin(), p.end());
         CHECK(tokens
               == make_results({ make_error(kind::lex_invalid_number, 0, 3),
-                                make_error(kind::lex_invalid_number, 4, 8),
-                                make_error(kind::lex_invalid_number, 9, 14),
-                                make_error(kind::lex_invalid_number, 15, 22),
-                                make_error(kind::lex_invalid_number, 23, 25),
-                                make_error(kind::lex_invalid_number, 26, 29) }));
+                                make_error(kind::lex_invalid_number, 4, 4),
+                                make_error(kind::lex_invalid_number, 9, 5),
+                                make_error(kind::lex_invalid_number, 15, 7),
+                                make_error(kind::lex_invalid_number, 23, 2),
+                                make_error(kind::lex_invalid_number, 26, 3) }));
       }
 
       SUBCASE("Invalid arbitrary radix - 0r0")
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -1205,20 +1205,33 @@ namespace jank::read::lex
           {
             ++pos;
             auto const after_dd_colon(peek());
-            if(after_dd_colon.is_ok() && after_dd_colon.expect_ok().character == '/')
+
+            /* Invalid ":: " source found. */
+            if(after_dd_colon.is_err())
+            {
+              return error::lex_invalid_keyword(
+                "A auto-resolved keyword must contain a valid symbol after the '::'.",
```
jeaye: `An`, not `A` here, since `auto` starts with a vowel sound.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -1205,20 +1205,33 @@ namespace jank::read::lex
...
+                if(result.is_err() || !is_symbol_char(ch))
                 {
                   break;
                 }
                 pos += result.expect_ok().len;
               }
-              return error::lex_invalid_keyword("An auto-resolved keyword may not start with '/'.",
-                                                { token_start, pos });
+              return error::lex_invalid_keyword(
+                util::format("An auto-resolved keyword may not start with '{}'.", ch),
```
jeaye: We have the information here for a better error. We know that we have a `::` and that we have one of two cases:

1. `::/`, so we can provide an error saying the namespace cannot be empty for the keyword.
2. `::` followed by a non-symbol char, so we can provide an error saying there's an unexpected end to that keyword. This is an incomplete keyword.

I think trying to word it as what a keyword can't start with is less helpful to the user. The user isn't going to type `::)` because they think a keyword can start with that. They're going to do that because they forgot to provide the rest of the keyword.

jeaye (changes requested):
Looks good, just some error message improvements we can do.

compiler+runtime/src/cpp/jank/read/lex.cpp
shantanu-sardesai: > ::/, so we can provide an error saying the namespace cannot be empty for the keyword.

Just wondering, since this is an auto-resolved keyword, can they even provide a namespace? If they can't then a better error would be to just say that you can't provide one, no?

jeaye: Yes they can. Auto-resolved keyword namespaces match namespace aliases (that sentence is a tongue twister).

For example:
```clojure
(require '[jank.compiler :as c])
::c/foo ; :jank.compiler/foo
```

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -1810,6 +1811,33 @@ namespace jank::read::lex
...
+                make_error(kind::lex_invalid_keyword, 0, 2),
+              }));
+      }
+
+      SUBCASE("Comma after ::")
+      {
+        processor p{ "::," };
+        native_vector<jtl::result<token, error_ref>> const tokens(p.begin(), p.end());
+        CHECK(tokens == make_results({ make_error(kind::lex_invalid_keyword, 0, 2) }));
+      }
```
jeaye, line 1850: The lexer needs to be able to correctly and gracefully continue lexing on a bad token and we need to test that for each bad token. So please include a test like `:: 12` to ensure you get both an invalid keyword and a valid number with the value 12.

shantanu-sardesai: Done.

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -1810,6 +1811,44 @@ namespace jank::read::lex
...
+
+      SUBCASE("Graceful lexing after invalid ::")
+      {
+        processor p{ ":: 42" };
+        native_vector<jtl::result<token, error_ref>> const tokens(p.begin(), p.end());
+        CHECK(tokens
+              == make_results({
+                make_error(kind::lex_invalid_keyword, 0, 2),
+                token{ 0, 3, token_kind::integer, "42" }
```
jeaye: If this test passes, we have a lexer bug. This says that the integer token starts at offset 0 and is 3 bytes long. **These numbers matter.** When we fill in a test, we have to check them and make sure they make sense.

These do not make sense.

Also, the value for the integer should not be a string. It should be an integral value. My guess is you didn't run this locally and it'll fail in CI.

shantanu-sardesai: Ah, I thought the 0 represented the line number.

Weird, I did run it locally before committing, I'll check.

jeaye: I believe you, since CI also passed. :grin:  So that means we have a lexer bug with this change. We need to make sure to properly progress the position when returning an error.

shantanu-sardesai: This is weirder that I thought, I just updated the test case to:

```c++
token{ 3, 2, token_kind::integer, 42ll }
```

And it worked... Looking into it.

shantanu-sardesai: As I would expect, this fails:

```C++
CHECK(tokens.at(1).expect_ok() == token{ 0, 2, token_kind::integer, 42ll });
```

and this passes:

```C++
CHECK(tokens.at(1).expect_ok() == token{ 3, 2, token_kind::integer, 42ll });
```

> Corollary of the above point, no lexer bug 🙂.

Are we sure that the use of `==` to compare `tokens` (type: `native_vector<jtl::result<token, error_ref>>`) and output of `make_results` (type: `std::array<jtl::result<token, error_ref>, N>`) correct? Or something else is going weird/wrong or just me thinking incorrectly...

I saw it being used [here](https://github.com/shantanu-sardesai/jank/blob/2d822586734ed6b1a2e0045065c34e3317cf96c4/compiler+runtime/test/cpp/jank/read/lex.cpp#L666-L674) so I wrote my tests similarly.

jeaye: Yeah, looks like we have some bad `==` logic going on here! Try jumping to definition on the `==` usage here and working through the logic. You can also step through it with a debugger.

From what I can see, your usage is correct. Could be we've had an issue in there already and we're only spotting it now.

shantanu-sardesai: [Fixed the bug](https://github.com/jank-lang/jank/pull/577/commits/476b9e78aee163908a957073f974546635befb37). As I feared, it opened up (or maybe exposed) a can of worms...

```
===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:492:
TEST SUITE: lex
TEST CASE:  Ratio
  Failures - x.x/x

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:492:
TEST SUITE: lex
TEST CASE:  Ratio
  Failures - xex/x

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:601:
TEST SUITE: lex
TEST CASE:  Integer
  Invalid octal number

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:601:
TEST SUITE: lex
TEST CASE:  Integer
  Edge case: Invalid mixed radix

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:601:
TEST SUITE: lex
TEST CASE:  Integer
  Mixed valid and invalid numbers

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:601:
TEST SUITE: lex
TEST CASE:  Integer
  Invalid hex numbers

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:601:
TEST SUITE: lex
TEST CASE:  Integer
  Invalid hex - 2.0x1

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:601:
TEST SUITE: lex
TEST CASE:  Integer
  Invalid arbitrary radix

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:980:
TEST SUITE: lex
TEST CASE:  N suffix
  1N/34

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:1108:
TEST SUITE: lex
TEST CASE:  Real
  Scientific notation
  Missing exponent

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:1108:
TEST SUITE: lex
TEST CASE:  Real
  Scientific notation
  Signs after exponent found

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:1108:
TEST SUITE: lex
TEST CASE:  Real
  Scientific notation
  Extra dots

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:1108:
TEST SUITE: lex
TEST CASE:  Real
  Scientific notation
  Extra characters in exponent

===============================================================================
/Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/test/cpp/jank/read/lex.cpp:1297:
TEST SUITE: lex
TEST CASE:  Big decimal
  Invalid - M in middle

===============================================================================
[doctest] test cases:  114 |  109 passed |  5 failed | 0 skipped
[doctest] assertions: 1665 | 1651 passed | 14 failed |
[doctest] Status: FAILURE!
```

Shall I fix these in the PR as well? I'm assuming yes, since we don't want to merge without a green pipeline?

shantanu-sardesai: > Ah, I thought the 0 represented the line number.

Heh, seems like there were similar misunderstandings in the [tests](https://github.com/jank-lang/jank/pull/577/commits/b5cb3eea633b372503ff48c32b6d751a96bb4585) on what the numbers mean 🙂. Anyways, fixed all of them.

jeaye: Nice! Yeah, we want all of these to be fixed. Thanks for putting in that effort.

compiler+runtime/test/cpp/jank/read/lex.cpp
```diff
@@ -78,7 +78,12 @@ namespace jank::read::lex
 
         auto const &lhs_err(lhs.expect_err());
         auto const &rhs_err(rhs.expect_err());
-        return lhs_err->kind == rhs_err->kind && lhs_err->source == rhs_err->source;
+        auto const is_eq{ lhs_err->kind == rhs_err->kind && lhs_err->source == rhs_err->source };
+
+        if(!is_eq)
+        {
+          return false;
+        }
```
jeaye, line 86: From what I see we don't actually return a value if `is_eq`. Shouldn't there be a `return true;` at the end of the fn, if you want to do this?

jeaye (changes requested):

compiler+runtime/test/cpp/jank/read/lex.cpp
shantanu-sardesai: [We already have that](https://github.com/jank-lang/jank/blob/main/compiler+runtime/test/cpp/jank/read/lex.cpp#L96), no?

If `is_eq` then we wish to check exhaustively and at the end of the loop, since we never bailed out in between, we just return `true`.

jeaye: Ah, yes. It was hidden.

image

jeaye:
Thank you, Shantanu!

PR #580, opened 2025-11-03
fix: Locking on Lazy Sequences

### LINKED ISSUE #170: lazy-seq can be realized multiple times
ambrose: Is there a race condition on this line that could realize the lazy-seq multiple times? https://github.com/jank-lang/jank/blob/18dd00c596d32731030ab9290a94b0749523289c/compiler%2Bruntime/src/cpp/jank/runtime/obj/lazy_sequence.cpp#L119
jeaye: Yeah, we have a few objects lacking the necessary synchronization. This is one of them!

stmonty:
Fixes #170 
This is just a current draft of what I am thinking about for how to implement locking on lazy sequences.
I am using a `recursive_mutex` with the synchronized state so that there wouldn't be a deadlock when the `lazy_seq` tries to call back into itself while holding the lock.

To be honest still trying to wrap my head around the implementation of lazy data structures. I think I understand how to use them, but how to build them is something new to me. Especially when discussing concurrency on them.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/obj/lazy_sequence.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/lazy_sequence.cpp
@@ -20,8 +20,7 @@ namespace jank::runtime::obj
     *lock = *other_lock;
   }
 
-  lazy_sequence &
-  lazy_sequence::operator=(lazy_sequence const &other)
+  lazy_sequence &lazy_sequence::operator=(lazy_sequence const &other)
   {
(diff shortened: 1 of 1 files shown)


stmonty:
@frenchy64 would love to get your thoughts as you created the original issue and maybe you have some context for this!

jeaye:
Closing this, since the draft has been stalled for a couple of months now. If you make progress and want to reopen, you can. But if this isn't something you're working on anymore, please unassign the ticket.

PR #581, opened 2025-11-04
Port `biginteger`

shantanu-sardesai:
A low hanging fruit 🙂.

Are we going to differentiate between a `BigInt` and `BigInteger` in jank the same way that Clojure does?

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -5444,27 +5444,11 @@
 (defn bigint
   "Coerce to BigInt"
   [x]
-  ;; (cond
-  ;;   (instance? clojure.lang.BigInt x) x
-  ;;   (instance? BigInteger x) (clojure.lang.BigInt/fromBigInteger x)
-  ;;   (decimal? x) (bigint (.toBigInteger ^BigDecimal x))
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -5465,7 +5465,7 @@
   ;;   (ratio? x) (.bigIntegerValue ^clojure.lang.Ratio x)
   ;;   (number? x) (BigInteger/valueOf (long x))
   ;;   :else (BigInteger. x))
```
jeaye: We can delete the 7 lines above this now, right?

jeaye:
> Are we going to differentiate between a `BigInt` and `BigInteger` in jank the same way that Clojure does?

No, that's an implementation detail of Java. Only `BigInt` is Clojure.

compiler+runtime/src/jank/clojure/core.jank
shantanu-sardesai: Oops, missed them.

shantanu-sardesai:
I've used the same `big_integer` implementation for both the `bigint` & `biginteger`.

jeaye:
Thanks!

PR #583, opened 2025-11-07
Port `double`

shantanu-sardesai:
Do we plan on differentiate between a `float` and a `double`?

Requires a repoint.


shantanu-sardesai:
Do we plan on differentiating between a `float` (`jank::f32`) and a `double` (`jank::f64`)?

image

---

Currently the implementation of the Clojure `float` in the jank's core is:

```jank
(def float
  "Coerce to float"
  (fn* real [o]
    (cpp/jank.runtime.to_real o)))
```

And the signature for `to_real` is:

```c++
f64 to_real(object_ref o);
```

AFAICT we don't have a datatype in the `jank::runtime::obj` to hold a `float` 🤔.

jeaye:
As you said, we don't have an object type for float right now. I don't think we will for some time. We'll just have the float fns work the same on our doubles.

shantanu-sardesai:
With this PR we can:
- [x] Merge the `double?` PR in the Clojure test suite.
- [x] Repoint the Clojure test suite.
- [x] Merge this PR.
- Keep the `double` Clojure test suite PR open till jank can support catching exceptions of a specific type.

shantanu-sardesai:
Repointed, this can be merged.

jeaye:
Thanks, Shantanu!

Issue #563, opened 2025-10-23
The nix build seems to consider only linux (aarch64-darwin fails)


ghost:
### jank health check

N/A

### Description of the issue with reproduction steps

Using `nix develop ".#"` fails immediately on "aarch64-darwin" architecture (Apple Silicon).

As a workaround, did some small tweaks to `llvm.nix`.

```bash
>  git diff  
diff --git a/llvm.nix b/llvm.nix
index 99eb6140..8ab7c687 100644
--- a/llvm.nix
+++ b/llvm.nix
@@ -14,6 +14,11 @@
 }: let
   gccForLibs = stdenv.cc.cc;
 
+  # On Darwin, libc doesn't have a .dev output
+  libcInclude = if stdenv.isDarwin
+    then "${stdenv.cc.libc}/include"
+    else "${gcc.libc.dev}/include";
+
   # nix adds these flags to the system clang transparently via a wrapped
   # executable. Since we will be compiling clang's runtime with our custom
   # (unwrapped) clang/clang++ executables, we need to provide them manually.
@@ -23,7 +28,7 @@
 
     "-isystem ${gcc.cc}/include/c++/${gcc.version}"
     "-isystem ${gcc.cc}/include/c++/${gcc.version}/${targetPlatform.config}"
-    "-isystem ${gcc.libc.dev}/include"
+    "-isystem ${libcInclude}"
 
     # some of these flags won't be used all the time
     "-Wno-unused-command-line-argument"
@@ -66,7 +71,7 @@ in
       (lib.cmakeBool "LLVM_ENABLE_OCAMLDOC" false)
 
       # nix-specific changes
-      (lib.cmakeFeature "C_INCLUDE_DIRS" "${gcc.libc.dev}/include")
+      (lib.cmakeFeature "C_INCLUDE_DIRS" "${libcInclude}")
       (lib.cmakeFeature "LLVM_ENABLE_RUNTIMES" "libunwind;compiler-rt")
       # fix linking path for compiler-rt sanitize libs with
       # -fsanitize={address,undefined}
```

However, after reexecuting `nix develop ".#"`, I get:

```
warning: download buffer is full; consider increasing the 'download-buffer-size' setting
error: Cannot build '/nix/store/8z7f1p4yq9y8ldfdbndf5wx5ysid502y-llvm-jank-22.0.0-git.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
[...]

PR #585, opened 2025-11-09
Improving portability in nix-darwin environments

ghost:
This commit addresses platform-specific linker library requirements in jank's AOT compilation system by moving the platform detection logic from runtime C++ preprocessor checks to build-time CMake configuration.

- Created `src/cpp/jank/aot/processor.cpp.in` as a template
- Moved hardcoded library list from processor.cpp to CMake configuration
- Libraries are now configured at build time based on target platform
- On macOS: excludes `-lstdc++` (Clang uses libc++ implicitly)
- On Linux: includes `-lstdc++` (requires explicit linking)

- Added `JANK_DEFAULT_LINKER_LIBS` variable with platform detection
- Use `configure_file()` to generate processor.cpp from template
- Template uses `@DEFAULT_LIBS_INITIALIZER@` placeholder for library list
- Updated jank_lib sources to reference generated file from build directory

- Removed `-L/opt/homebrew/opt/llvm/lib/unwind` on macOS
- This path can cause linking issues and is not needed on macOS

Previously, the library list in processor.cpp was hardcoded with `-lstdc++` for all platforms. On macOS, Clang uses libc++ by default and explicitly linking libstdc++ can cause conflicts or warnings.

By using CMake's `configure_file()`, we generate the correct library list at configuration time, making the build system more portable and reducing the need for runtime platform checks.

- Builds successfully on macOS (arm64, LLVM 22)
- All `check-health` tests pass
- AOT compilation works correctly

**EDIT**:
Approach changed based on feedback. See commit message.

### DIFF
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -112,7 +112,7 @@ set(
   )
 
 if (NOT jank_local_clang)
-  list(APPEND jank_linker_flags -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++ -L/opt/homebrew/opt/llvm/lib/unwind)
+  list(APPEND jank_linker_flags -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++)
 endif ()
 
 # If the user opts into a specific build type, we will provide sane defaults. However, for
@@ -220,6 +220,20 @@ list(APPEND jank_common_compiler_flags
   -DJANK_CLANG_RESOURCE_DIR="${clang_resource_dir}"
 )
 
+# Platform-specific filtering of system include paths for JIT flags
+# On macOS, filter out SDK paths that should come from -isysroot at JIT time
+# This prevents baking absolute SDK paths into the binary's JIT flags
+set(clang_jit_include_flags "${clang_system_include_flags}")
+if(APPLE)
+  set(_filtered_jit_flags "")
+  foreach(flag ${clang_system_include_flags})
+    if(NOT flag MATCHES "/SDKs/")
+      list(APPEND _filtered_jit_flags "${flag}")
+    endif()
+  endforeach()
+  set(clang_jit_include_flags "${_filtered_jit_flags}")
+endif()
+
 set(
   jank_jit_compiler_flags
   # We don't actually want to see any warnings for JIT compiled code.
@@ -230,7 +244,7 @@ set(
   -Xclang -fno-validate-pch
   -Xclang -fno-pch-timestamp
   ${LLVM_DEFINITIONS_LIST}
-  ${clang_system_include_flags}
+  ${clang_jit_include_flags}
 )
 # ---- Clang JIT environment ----
 
--- a/compiler+runtime/src/cpp/jank/aot/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/aot/processor.cpp
@@ -228,7 +228,6 @@ int main(int argc, const char** argv)
     for(auto const &lib : { "-ljank-standalone",
                             /* Default libraries that jank depends on. */
                             "-lm",
-                            "-lstdc++",
                             "-lLLVM",
                             "-lclang-cpp",
                             "-lcrypto",
@@ -236,6 +235,13 @@ int main(int argc, const char** argv)
                             "-lzstd" })
     {
       compiler_args.push_back(strdup(lib));
+    }
+
+    /* On non-macOS platforms, explicitly link libstdc++.
+     * macOS uses libc++ implicitly via Clang. */
+    if constexpr(jtl::current_platform != jtl::platform::macos_like)
+    {
+      compiler_args.push_back(strdup("-lstdc++"));
     }
 
     for(auto const &define : util::cli::opts.define_macros)


compiler+runtime/CMakeLists.txt
```diff
@@ -115,6 +115,12 @@ if (NOT jank_local_clang)
   list(APPEND jank_linker_flags -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++ -L/opt/homebrew/opt/llvm/lib/unwind)
 endif ()
 
+# Platform-specific: Remove -L/opt/homebrew/opt/llvm/lib/unwind on macOS
+# as it's not needed and can cause linking issues
+if(APPLE AND NOT jank_local_clang)
+  list(REMOVE_ITEM jank_linker_flags "-L/opt/homebrew/opt/llvm/lib/unwind")
+endif()
+
```
jeaye: I don't think this is needed, if we just remove it from line 115.

compiler+runtime/CMakeLists.txt
```diff
@@ -374,6 +394,38 @@ include(cmake/dependency/boost-multiprecision.cmake)
...
+  string(APPEND DEFAULT_LIBS_INITIALIZER "    \"${lib}\",\n")
+endforeach()
+
+# Configure the processor.cpp file with platform-specific libraries
+configure_file(
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/jank/aot/processor.cpp.in"
+  "${CMAKE_CURRENT_BINARY_DIR}/src/cpp/jank/aot/processor.cpp"
+  @ONLY
+)
+# ---- Platform-specific configuration complete ----
```
jeaye: We have significantly complicated our AOT processor code here, by making it a `.in` file. The only thing we get out of this is to conditionally add `-lstdc++` when not on macOS. We can just do this in the C++ code instead.

```cpp
if constexpr(jtl::current_platform != jtl::platform::macos_like)
{
  compiler_args.push_back(strdup("-lstdc++"));
}
```

compiler+runtime/CMakeLists.txt
```diff
@@ -220,6 +226,20 @@ list(APPEND jank_common_compiler_flags
   -DJANK_CLANG_RESOURCE_DIR="${clang_resource_dir}"
 )
 
+# Platform-specific filtering of system include paths for JIT flags
+# On macOS, filter out SDK paths that should come from -isysroot at JIT time
+# This prevents baking absolute SDK paths into the binary's JIT flags
```
jeaye, line 225: I would like to understand this better. Would you mind explaining the exact scenario you had here?

You AOT compiled on one macOS, with one SDK version, then tried to run on a different macOS machine, with a different SDK version? And then it failed? Is the Clang JIT runtime adding the correct SDK for us, in that case, or are we getting it from somewhere else?

compiler+runtime/CMakeLists.txt
ghost: Fixed.

compiler+runtime/CMakeLists.txt
ghost: Awesome!

compiler+runtime/CMakeLists.txt
ghost: Thanks @jeaye! I should clarify the context better.

I'm working on making jank installable via Homebrew on macOS in a nix-darwin environment. Although jank builds cleanly on vanilla macOS, nix-darwin setting SDKROOT and other environment variables interfere with the build.

My SDK filtering changes might actually help both scenarios:
1. nix-darwin + Homebrew (my use case): Filters out nix-polluted SDK paths
2. Pure Nix on macOS (issue #563): Could help with the SDK path detection issues

The filtering ensures JIT uses -isysroot dynamically at runtime rather than hardcoded build-time SDK paths, which should make jank more portable across different macOS/Nix configurations.

jeaye:
CI is gummed up right now, due to a Github bug: https://github.com/actions/runner-images/issues/13341

This is the sort of thing we definitely want to make sure works in CI, though. Please consider these changes accepted, but stand by until we're able to run our macOS CI jobs again.

jeaye:
Thank you!

compiler+runtime/CMakeLists.txt
```diff
@@ -112,7 +112,7 @@ set(
   )
 
 if (NOT jank_local_clang)
-  list(APPEND jank_linker_flags -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++ -L/opt/homebrew/opt/llvm/lib/unwind)
+  list(APPEND jank_linker_flags -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++)
```
jeaye, line 115: @oscarvarto Removing this causes issues for people. What issue, exactly, did you run into with this provided?

shantanu-sardesai: I've created this [PR](https://github.com/jank-lang/jank/pull/607) that reverts this change, feel free to update/close it based on your discussions here.

ghost: @jeaye, @shantanu-sardesai  We briefly mentioned this above:
https://github.com/jank-lang/jank/pull/585#discussion_r2547449135

I think the solution should work for both Linux and macOS. My understanding is that unwind is needed on Linux only. So we need a mechanism to remove/skip it on macOS.

shantanu-sardesai: I think I should have mentioned, I'm on a mac 🙂.

shantanu-sardesai: Not sure how relevant this is but I still do get these warnings:

```
$ ./bin/watch ./bin/test
...
-- Build files have been written to: /Users/shantanusardesai/Desktop/code/projects/jank/compiler+runtime/build
[158/164] Linking CXX executable jank-phase-1
ld: warning: reexported library with install name '/opt/homebrew/opt/llvm/lib/unwind/libunwind.1.dylib' found at '/opt/homebrew/Cellar/llvm/HEAD-17efa57/lib/unwind/libunwind.1.0.dylib' couldn't be matched with any parent library and will be linked directly
[162/164] Linking CXX executable jank-test
ld: warning: reexported library with install name '/opt/homebrew/opt/llvm/lib/unwind/libunwind.1.dylib' found at '/opt/homebrew/Cellar/llvm/HEAD-17efa57/lib/unwind/libunwind.1.0.dylib' couldn't be matched with any parent library and will be linked directly
...
```

ghost: @shantanu-sardesai I recompiled with the flag restored, and it seems to be working for me as well. `jank check-health` gives me:

```
─ ✅ jank can jit compile llvm ir
ld: warning: search path '/lib' not found
ld: warning: reexported library with install name '/opt/homebrew/opt/llvm/lib/unwind/libunwind.1.dylib' found at '/opt/homebrew/Cellar/llvm/HEAD-d96a93f/lib/unwind/libunwind.1.0.dylib' couldn't be matched with any parent library and will be linked directly
```

but I guess that's not a problem.

@shantanu-sardesai Your PR with a comment is necessary!

shantanu-sardesai: Yeah, I have been getting those warning for quite some time, for the moment I guess they are fine.

jeaye: Ok, merged it back in. Thanks, guys.

PR #586, opened 2025-11-12
Update depth in submodule update in build.md

pfeodrippe:
Increased the depth for submodule update from 1 to 100 because it was giving an error locally when using depth of 1 as it was unable to fetch older commits.

### DIFF
--- a/compiler+runtime/doc/build.md
+++ b/compiler+runtime/doc/build.md
@@ -43,10 +43,10 @@ export CPPFLAGS="-I/opt/homebrew/opt/llvm/include ${CPPFLAGS}"
 Clone the repo as follows:
 
 ```bash
-git clone --depth 1 --single-branch --shallow-submodules --recurse-submodules https://github.com/jank-lang/jank.git
+git clone --recurse-submodules https://github.com/jank-lang/jank.git
 
 # If you didn't recurse submodules when cloning, you'll need to run this.
-git submodule update --init --recursive --depth 100 --jobs 8
+git submodule update --init --recursive --jobs 8
 ```
 
 ## Compiling Clang/LLVM


compiler+runtime/doc/build.md
```diff
@@ -46,7 +46,7 @@ Clone the repo as follows:
 git clone --depth 1 --single-branch --shallow-submodules --recurse-submodules https://github.com/jank-lang/jank.git
```
pfeodrippe: This one works as depth is only for jank itself and everyhing seems to be pulled for each submodule

compiler+runtime/doc/build.md
```diff
@@ -46,7 +46,7 @@ Clone the repo as follows:
...
 
 # If you didn't recurse submodules when cloning, you'll need to run this.
-git submodule update --init --recursive --depth 1 --jobs 8
+git submodule update --init --recursive --depth 100 --jobs 8
```
pfeodrippe: Or maybe we should remove depth from here to have the same behaviour as above? Maybe it would be better to just remove it, but let me know o/

jeaye: Yeah, let's remove depth. It causes lots of issues. Very handy in CI, but less so locally.

compiler+runtime/doc/build.md
jeaye: I'd say we can remove it from all places and let people add it themselves, if they so desire.

compiler+runtime/doc/build.md
pfeodrippe: Done, thanks \o

jeaye:
Thanks!

PR #587, opened 2025-11-14
fix: odd? on negative numbers works properly

### LINKED ISSUE #512: `odd?` incorrect for negative numbers
```clojure
(odd? -3) => false
```

stmonty:
Just a simple fix that works with negative numbers. Fixes #512


jeaye:
Thanks!

PR #591, opened 2025-11-17
Fix ODR violation for cpp/raw functions during AOT compilation

### LINKED ISSUE #582: ODR violation for cpp/raw functions during AOT compilation
### jank health check

```
─ system ───────────────────────────────────────────────────────────────────────────────────────────
─ ✅ operating system: linux
─ ✅ default triple: x86_64-unknown-linux-gnu

─ jank install ─────────────────────────────────────────────────────────────────────────────────────
─ ✅ jank version: jank-0.1-e671f1b0555a2a82371994784f0a6bb4b3519b43
─ ✅ jank resource dir: ../lib/jank/0.1 
─ ✅ jank resolved resource dir: /usr/local/bin/../lib/jank/0.1 (found)
─ ✅ jank user cache dir: /home/mauricio/.cache/jank/x86_64-unknown-linux-gnu-e593da5ef0bf37ba08dd893dc570a851271246c873d5d452dc8c6b42c63e4c0a (found)

─ clang install ────────────────────────────────────────────────────────────────────────────────────
─ ⚠  configured clang path: /home/runner/work/jank/jank/compiler+runtime/build/llvm-install/usr/local/bin/clang++ (not found)
─ ✅ runtime clang path: /usr/local/bin/../lib/jank/0.1/bin/clang++ (found)
─ ⚠  configured clang resource dir: /home/runner/work/jank/jank/compiler+runtime/build/llvm-install/usr/local/lib/clang/22 (not found)
─ ✅ runtime clang resource dir: /usr/local/lib/jank/0.1/lib/clang/22 (found)

─ jank runtime ─────────────────────────────────────────────────────────────────────────────────────
─ ✅ jank runtime initialized
─ ✅ jank pch path: /home/mauricio/.cache/jank/x86_64-unknown-linux-gnu-e593da5ef0bf37ba08dd893dc570a851271246c873d5d452dc8c6b42c63e4c0a (found)
─ ✅ jank can jit compile c++
─ ✅ jank can jit compile llvm ir
─ ✅ jank can aot compile working binaries

─ support ──────────────────────────────────────────────────────────────────────────────────────────
If you're having issues with jank, please either ask the jank community on the Clojurians Slack or report the issue on Github.

─ Slack: https://clojurians.slack.com/archives/C03SRH97FDK
─ Github: https://github.com/jank-lang/jank
```

### Description of the issue with reproduction steps

With this code in a file called `something.jank`:

```clojure
(ns something)

(cpp/raw "
inline int hello() {
  return 10;
}
")

(defn -main []
  (cpp/hello))
```

When I try to compile this code either with:

`jank --module-path . compile something`

or

`jank --module-path . compile-module something`

I get the following error:

```
input_line_5:2:12: error: redefinition of 'hello'
    2 | inline int hello() {
      |            ^
input_line_1:2:12: note: previous definition is here
    2 | inline int hello() {
      |            ^
─ runtime/unable-to-load-module ────────────────────────────────────────────────────────────────────
error: Unable to parse 'cpp/raw' expression.                                                        
```
```

### Stack trace

_No response_

### PR DESCRIPTION
**Pull request by pfeodrippe**
Fixes #582 .

We use a simple #ifndef/#define guard as a workaround, so all calls to `cpp/raw` that have exactly the same content (same string) will be considered equivalent and run only once. And it should be fine as we should use `cpp/raw` only in side-effects situations. If really needed, we can think about creating a `cpp/raw*` or `cpp/raw-once`, but it should be enough for now.

### CODE REVIEW

**Comment by pfeodrippe** [comment]
The nix CI failure seems to be related to some external caching from cachix

**Comment by pfeodrippe** [comment]
Let me know if we need something else here or if we'd like another approach o/

File: compiler+runtime/src/cpp/jank/analyze/processor.cpp [hunk 1]
```diff
@@ -3662,8 +3662,28 @@ namespace jank::analyze
...
 
     auto const string_obj(l->data.rest().first().unwrap());
+
+    /* Only apply guards if the argument is a string literal */
+    runtime::object_ref guarded_string_obj{ string_obj };
+    if(string_obj->type == runtime::object_type::persistent_string)
+    {
+      auto const raw_string{ expect_object<runtime::obj::persistent_string>(string_obj)->data };
+      auto const content_hash{ std::hash<jtl::immutable_string>{}(raw_string) };
+      auto const guard_name{ util::format("CPP_RAW_INLINE_{}_DECLARED", content_hash) };
```
jeaye: Let's call this `JANK_CPP_RAW_{}`.

compiler+runtime/test/bash/ahead-of-time/src/cpp_raw_inline/core.jank
```diff
@@ -0,0 +1,20 @@
...
+#_(cpp/raw "
+#ifndef CPP_RAW_INLINE_HELLO_DECLARED
+#define CPP_RAW_INLINE_HELLO_DECLARED
+
+ inline int cpp_rawinline_hello() {
+  return 10;
+}
+#endif
+")
+
```
jeaye: Please remove this.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -3662,8 +3662,28 @@ namespace jank::analyze
     }
 
     auto const string_obj(l->data.rest().first().unwrap());
+
+    /* Only apply guards if the argument is a string literal */
+    runtime::object_ref guarded_string_obj{ string_obj };
+    if(string_obj->type == runtime::object_type::persistent_string)
```
jeaye: We already do a check below to ensure that `cpp/raw` only receives a string. If we give it anything else, we have an error. We should just update the string in the `primitive_literal` expression that we get back later in this function. We have it bound as `auto const obj`.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -3662,8 +3662,28 @@ namespace jank::analyze
     }
 
     auto const string_obj(l->data.rest().first().unwrap());
+
+    /* Only apply guards if the argument is a string literal */
```
jeaye: This comment describes "what", but not "why". The code itself already says "what", and we don't want to duplicate what the code says. A go at describing "why" might be:

> We wrap all `cpp/raw` strings in unique preprocessor guards because jank
> currently does codegen twice when compiling and this can lead to ODR violations.

jeaye (changes requested):
Thanks for the PR! The overall approach is good, we just have some shuffling to do in order to make it fit into the analyzer better.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
pfeodrippe: All should be good now, thanks, man!! \o

jeaye:
Thanks for the fix!

PR #596, opened 2025-11-19
Fix integer equal with big_integer

djblue:
Before:
```clojure
% jank repl
user=> (= 2N 2)
true
user=> (= 2 2N)
false
```
After:
```clojure
% jank repl
user=> (= 2N 2)
true
user=> (= 2 2N)
true
```


jeaye:
Thanks!

PR #597, opened 2025-11-20
Fix `even?` and `odd?` for `big_integer`

djblue:
Before:
```clojure
% jank repl
user=> (even? 2)
true
user=> (even? 2N)
Uncaught exception: invalid object type: 3
```
After:
```clojure
% jank repl 
user=> (even? 2)
true
user=> (even? 2N)
true
```


jeaye:
Thanks!

Issue #349, opened 2025-07-13
Implement interop dtors for stack-allocated C++ values


jeaye:
We need to use the code from #329 to implement this.

jeaye:
This is done and tested, using C++ interop and our custom IR.

Issue #242, opened 2025-02-06
Add source mapping from IR


jeaye:
## Overview
When using a debugger, or viewing a stack trace, we want jank source information tied to our native code. This requires specifically emitting that information in the LLVM IR which we generate.

## Steps
- [ ] Add a flag to enable source mapping (`-g`)
- [ ] When that flag is enabled, generate source information for all expressions
  - Start with just function definitions, then add fn calls, then work on let bindings, etc
- [ ] Verify this is working in GDB, LLDB, etc
- [ ] Verify this is working for stacktrace generation

## References
The LLVM documentation for this is here: https://llvm.org/docs/SourceLevelDebugging.html

The LLVM build-your-own-language tutorial has a chapter on this: https://llvm.org/docs/tutorial/MyFirstLanguageFrontend/LangImpl09.html

I would also use Clang, with LLVM IR output enabled, to see what it adds. For example:

```bash
❯ cat source-info.c
#include <stdio.h>

int main()
{
    printf("meow");
    return 0;
}

❯ clang -O0 -ggdb -S -emit-llvm -o - source-info.c
<inspect output>
```

jeaye:
We support this now.

PR #598, opened 2025-11-23
Switch to C++ codegen by default

### LINKED ISSUE #354: Improve sorted container performance
As described in https://github.com/arximboldi/immer/issues/105, BppTree isn't actually working for us. The issues are, in order of priority:

1. BppTree doesn't support custom allocators, so it's not using jank's GC, so we're leaking all sorted maps and sets
2. BppTree doesn't support runtime functions for sorting, so we can't implement `clojure.core/sorted-set-by`
3. BppTree is super heavy to compile, which affects not only jank's compile-time but also its startup time, due to our JIT compilation of jank's headers

Our best bet is for immer to add support for sorted sets and maps, but that'll likely cost a fair amount of money to get Juanpe to do it. In the meantime, if we can't find an alternative, we should just not support sorted sets and maps, rather than "supporting" them by leaking all of their contents.

----

Edit: BppTree has been removed and we're using C++ stdlib containers with deep copying for sorted containers. This *works*, but we need something more efficient.

### LINKED ISSUE #372: Catch C++ codegen up and make it easy to use
As part of the interop work, I added the C++ code generator back in. It compiles now, but we need to catch it back up with all of the interop expressions and we need to add allow the user to choose which codegen to do.

- [ ] Add a `--codegen <ir,cpp>` flag, which defaults to `ir`; this can be used at any time, not just for AOT
- [ ] Implement the new expressions (`case*`, `letfn*`, and all interop)
- [ ] Update bash test suite to run `build/jank-test` with both codegen types
- [ ] Default to C++ codegen for `clojure.core`

jeaye:
On main, we have issues with the GC on Linux and macOS. It doesn't collect! After working to resolve these issues (in #549), I managed to get Linux working, but macOS would still crash due to premature collections. I found that the issues don't happen with C++ codegen, so I'm committing us to that for the alpha release. We can sort out why the IR leads to premature collections as we move forward, but I've already put a couple solid weeks into this and there's too much else to do.

Embracing C++ codegen makes sense for other reasons, too. It's far simpler than IR gen. It leads to faster runtime performance than IR gen. It celebrates that jank is C++ and it more closely ties jank's AST to C++. It also allows us to not worry about entire unimplemented IR features like destructors.

I still plan on keeping IR gen, but we can develop it in parallel, take our time with it, and enable it by default when it makes sense to do so. There should never be a behavioral difference between the two.

This closes #549.
This closes #372.
This replaces #349, so it's no longer needed in the alpha.
This closes #354.
This replaces #242, so it's no longer needed in the alpha.


pfeodrippe:
I have https://github.com/jank-lang/jank/pull/599 \o

I'm seeing some errors locally (as the CI is borked for some reason, it seems, I've just something like `v1`  in one of my branches instead of using the hashed), but maybe you are seeing them as well and it's expected (they were also happening before applying https://github.com/jank-lang/llvm-project/pull/1 )

```
TEST SUITE: jit
TEST CASE:  files

/Users/pfeodrippe/dev/jank-branches/issue-582-fix/compiler+runtime/test/cpp/jank/jit/processor.cpp:192: ERROR: CHECK( failures.empty() ) is NOT correct!
  values: CHECK( false )

failure: test/jank/cpp/new/builtin/pass-aliased.jank
        Exception thrown: Failed to evaluate C++ code.
failure: test/jank/cpp/constructor/complex/fail-template-instantiation.jank
        Test failure was expected, but it passed with :success
failure: test/jank/cpp/constructor/builtin/pass-aliased.jank
        Exception thrown: Failed to evaluate C++ code.
tested 576 jank files with 9 skips and 3 failures
```

pfeodrippe:
Created https://github.com/jank-lang/jank/pull/600 (a draft), but this is for extra stuff if it makes sense for the alpha release (it may need more iteration), don't bother if it's fine as it's o/ (e.g. inst tests)

jeaye:
> I have #599 \o
> 
> I'm seeing some errors locally (as the CI is borged for some reason, it seems, I've just something like `v1` in one of my branches instead of using the hashed), but maybe you are seeing them as well and it's expected (they were also happening before applying [jank-lang/llvm-project#1](https://github.com/jank-lang/llvm-project/pull/1) )
> 
> ```
> TEST SUITE: jit
> TEST CASE:  files
> 
> /Users/pfeodrippe/dev/jank-branches/issue-582-fix/compiler+runtime/test/cpp/jank/jit/processor.cpp:192: ERROR: CHECK( failures.empty() ) is NOT correct!
>   values: CHECK( false )
> 
> failure: test/jank/cpp/new/builtin/pass-aliased.jank
>         Exception thrown: Failed to evaluate C++ code.
> failure: test/jank/cpp/constructor/complex/fail-template-instantiation.jank
>         Test failure was expected, but it passed with :success
> failure: test/jank/cpp/constructor/builtin/pass-aliased.jank
>         Exception thrown: Failed to evaluate C++ code.
> tested 576 jank files with 9 skips and 3 failures
> ```

I ran into these on macOS locally as well. Three failures. I will look into them, but we can count them as unrelated to your LLVM changes.

pfeodrippe:
> I ran into these on macOS locally as well. Three failures. I will look into them, but we can count them as unrelated to your LLVM changes.

Yeah, checked locally and https://github.com/jank-lang/jank/pull/600/files#diff-c5647a31142ea401b3cdb72da13815ad411e00155f7be18be66466cac5cd6ae2 (only the changes at cpp_util.cpp) fix the aliased ones as well the jit/processor.cpp one (alias are always a problem ehuehuehueh) \o

For the template one, just modify it to a pass as done at https://github.com/jank-lang/jank/pull/600/files#diff-ce521936753298bfe26a40586f6ac2887d71294b003d119fbcec80682a1b036f , check some value for `.-a` and it should be good (we have to use `a(std::to_string(t))` so the value is not interpreted as a char)

jeaye:
> > I ran into these on macOS locally as well. Three failures. I will look into them, but we can count them as unrelated to your LLVM changes.
> 
> Yeah, checked locally and https://github.com/jank-lang/jank/pull/600/files#diff-c5647a31142ea401b3cdb72da13815ad411e00155f7be18be66466cac5cd6ae2 (only the changes at cpp_util.cpp) fix the aliased ones as well the jit/processor.cpp one (alias are always a problem ehuehuehueh) \o
> 
> For the template one, just modify it to a pass as done at https://github.com/jank-lang/jank/pull/600/files#diff-ce521936753298bfe26a40586f6ac2887d71294b003d119fbcec80682a1b036f , check some value for `.-a` and it should be good (we have to use `a(std::to_string(t))` so the value is not interpreted as a char)

Please keep that PR focused on just updating LLVM. Your help on jank and this Clang issue is very welcome, but I am not open to vibe-coded changes in jank.

The changes for inst were out of left field, the alias fix is not how we do things in jank (we don't use Clang's AST directly, we use CppInterOp), and the changed template test takes a test which is **supposed to fail** and changes it to make it pass.

I appreciate your excitement to help out, and your help overall. This is a boundary we must have in order to keep my life sane when it comes to maintaining this code, though.

pfeodrippe:
> > > I ran into these on macOS locally as well. Three failures. I will look into them, but we can count them as unrelated to your LLVM changes.
> > 
> > 
> > Yeah, checked locally and https://github.com/jank-lang/jank/pull/600/files#diff-c5647a31142ea401b3cdb72da13815ad411e00155f7be18be66466cac5cd6ae2 (only the changes at cpp_util.cpp) fix the aliased ones as well the jit/processor.cpp one (alias are always a problem ehuehuehueh) \o
> > For the template one, just modify it to a pass as done at https://github.com/jank-lang/jank/pull/600/files#diff-ce521936753298bfe26a40586f6ac2887d71294b003d119fbcec80682a1b036f , check some value for `.-a` and it should be good (we have to use `a(std::to_string(t))` so the value is not interpreted as a char)
> 
> Please keep that PR focused on just updating LLVM. Your help on jank and this Clang issue is very welcome, but I am not open to vibe-coded changes in jank.
> 
> The changes for inst were out of left field, the alias fix is not how we do things in jank (we don't use Clang's AST directly, we use CppInterOp), and the changed template test takes a test which is **supposed to fail** and changes it to make it pass.
> 
> I appreciate your excitement to help out, and your help overall. This is a boundary we must have in order to keep my life sane when it comes to maintaining this code, though.

For sure for sure o/ Yeah, I've reviewed the code, but didn't know about the preference, for example, about CppInterOp or that we wanted to keep it failing. Don't want to create AI slop for you for sure (or anyone) Sorry \o

PR #601, opened 2025-11-23
Allow users to customize pch includes

cjbarre:
## Problem

As a user, I should be able to guarantee the availability of standard system libs for my jank program at runtime.

However, currently, what is available depends on what jank's pre-compiled header includes.

This means that users can be exposed to transitive dependency issues.

### Example 

- Commit A: My program and Jank both `#include <float.h>` somewhere, all is good.
- Commit B: My program does `#include <float.h>`, but jank has swapped out a dependency and no longer needs `float.h`

My program is now broken because `float.h` is no longer included in Jank's PCH.

## Solution

This PR allows users to customize what gets included in the `prelude.hpp` file which ultimately becomes the PCH.

The user is now able to supply a new flag `--include-pch` to the jank binary when running jank commands.

### Example

`jank --include-pch "float.h" ...`

## Notes

I'm aware that there may be language and naming tweaks as well as tweaks to the approach.

[Original Slack Thread](https://clojurians.slack.com/archives/C03SRH97FDK/p1763444293023409)

### DIFF
--- a/compiler+runtime/src/cpp/jank/util/clang.cpp
+++ b/compiler+runtime/src/cpp/jank/util/clang.cpp
@@ -221,7 +221,6 @@ namespace jank::util
 
   jtl::option<jtl::immutable_string> find_pch(jtl::immutable_string const &binary_version)
   {
-
     /* For AOT compiled jank programs, the PCH is embedded within the executable as a resource.
      * In this case, we just grab a pointer to the data and set it up with Clang's VFS so
      * no copying or file IO is required. */


compiler+runtime/doc/build.md
```diff
@@ -63,6 +63,8 @@ export CC=$PWD/build/llvm-install/usr/local/bin/clang; export CXX=$PWD/build/llv
 
 Now configure and build jank as normal, but pass `-Djank_local_clang=on` when you configure.
 
+If you need to customize Jank's Pre-Compiled Header (PCH), then pass `-Djank_runtime_pch=off` and use `--include-pch` flags in your jank command.
+
```
jeaye, line 67: We can remove this, since there is no such CMake flag and `build.md` is not for documenting jank CLI flags.

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -234,24 +232,62 @@ namespace jank::util
...
+                    native_vector<jtl::immutable_string> const &pch_includes,
+                    jtl::immutable_string const &binary_version)
+  {
+    /* Copy the prelude.hpp template to a temporary directory */
+
+    std::filesystem::path customized_prelude_path = format("{}/jank_prelude_customized_{}.hpp",
+                                                           std::filesystem::temp_directory_path(),
+                                                           binary_version)
+                                                      .c_str();
+    auto const copy_options = std::filesystem::copy_options::overwrite_existing;
```
jeaye, line 267: Please always use direct initialization. i.e.

```cpp
auto const copy_options{ std::filesystem::copy_options::overwrite_existing };
```

Please do this for the entire PR (I won't comment on any more of them this review).

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -234,24 +232,62 @@ namespace jank::util
...
+
+    customized_prelude_file_out << "\n\n /* Hello user, you have provided the include(s) below via "
+                                   "the --include-pch flag(s) in your jank command. */ \n\n";
+
+    for(auto const &pch_include : pch_includes)
+    {
+      customized_prelude_file_out << format("#include <{}> \n", pch_include);
+    }
+
+    customized_prelude_file_out.close();
```
jeaye, line 282: There is no need for this. This is C++. We have destructors. :)

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -234,24 +232,62 @@ namespace jank::util
...
+                                                      .c_str();
+    auto const copy_options = std::filesystem::copy_options::overwrite_existing;
+    std::filesystem::copy(prelude_path, customized_prelude_path, copy_options);
+
+    /* Append user includes to customized prelude file */
+
+    std::ofstream customized_prelude_file_out(customized_prelude_path, std::ios::app);
+
+    customized_prelude_file_out << "\n\n /* Hello user, you have provided the include(s) below via "
+                                   "the --include-pch flag(s) in your jank command. */ \n\n";
```
jeaye, line 275: Let's just insert new lines without the cheeky message. :upside_down_face:

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -234,24 +232,62 @@ namespace jank::util
...
+   * 
+   * Given the original prelude.hpp path, the pch_includes from the CLI opts, and the binary version
+   * Return the filesystem path of the customized prelude header file.
+   */
+  static std::filesystem::path
+  customize_prelude(std::filesystem::path const &prelude_path,
+                    native_vector<jtl::immutable_string> const &pch_includes,
+                    jtl::immutable_string const &binary_version)
+  {
+    /* Copy the prelude.hpp template to a temporary directory */
```
jeaye, line 261: If `pch_includes` is empty, which is the normal case, we can just return `prelude_path` and not do any of the other work.

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -70,6 +70,14 @@ namespace jank::util::cli
                    opts.libs,
                    "Library identifiers, absolute or relative paths eg. -lfoo for libfoo.so or "
                    "foo.dylib. Can be specified multiple times.");
+    cli.add_option(
+      "--include-pch",
```
jeaye, line 74: Please call this `--include`, not `--include-pch`. That will fit the behavior of Clang.

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -70,6 +70,14 @@ namespace jank::util::cli
...
+      opts.pch_includes,
+      "Include a system header that you or your dependencies will need at runtime in your program. "
+      "Some system headers are included transitively by Jank, but those can change arbitrarily. "
+      "Use this to ensure stability in your program. "
+      "Can be specified multiple times. "
+      "Ex. --include-pch \"float.h\"");
```
jeaye, line 80: This description is not correct. jank allows for people to just `(cpp/raw "#include <foo.h>")` at any time. You should not have to muck with the PCH. In your specific use case, though, the PCH was affecting your setup. That is incredibly niche.

The typical use case for this flag is not even your niche case, it's actually for faster startup. Also, there's nothing saying the include needs to be a system header. Finally, please make sure to always keep jank lowercase for any material within the `jank-lang` Github org.

A suggested string to use instead:

> Appends an include to jank's pre-compiled header. Includes are wrapped in `<>` automatically. Can be used multiple times.

compiler+runtime/src/cpp/jank/util/environment.cpp
```diff
@@ -115,6 +115,15 @@ namespace jank::util
       sb(def);
     }
 
+    sb(".");
+
+    sb("pch_includes:");
```
jeaye, line 120: There's no need for this line.

flake.nix
```diff
@@ -98,6 +98,10 @@
               )
             '';
 
+            preBuild = ''
+              export HOME=$TMPDIR
+            '';
```
jeaye, line 103: Something like this warrants a comment to explain **why** it's there.

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -273,6 +309,9 @@ namespace jank::util
       include_path = install_path;
     }
 
+    /* Customize prelude.hpp with user supplied --include-pch flags */
```
jeaye, line 312: 1. Especially since the flag name itself is changing, it's better to not repeat it in so many (5) places. 
2. Please use complete sentences and proper grammar for all comments. Comments are documentation. That includes ending sentences with a period.

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -234,24 +232,62 @@ namespace jank::util
...
+   * pre-compiled header, the template for which, is the prelude.hpp file. The flag
+   * by which they do this is --include-pch. This function takes those inputs, such as
+   * "float.h" or "cfloat" and appends them as #include directives to a copy of the prelude.hpp file. 
+   * This ensures that when the user includes these headers in their program, they will be available. 
+   * It prevents the user from depending on transitive dependencies included by jank itself as a basis 
+   * for writing jank programs. 
+   * 
+   * Given the original prelude.hpp path, the pch_includes from the CLI opts, and the binary version
+   * Return the filesystem path of the customized prelude header file.
+   */
```
jeaye, line 255: This comment is incorrect, as I explained elsewhere in this review. This is not the purpose of the PCH and your incredibly niche issue is not the common case.

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -70,6 +70,14 @@ namespace jank::util::cli
                    opts.libs,
                    "Library identifiers, absolute or relative paths eg. -lfoo for libfoo.so or "
                    "foo.dylib. Can be specified multiple times.");
+    cli.add_option(
```
jeaye, line 73: This is a whole new feature, so we really need some tests for it. If you're open to providing those tests, we'll want to jump into the `test/bash` suite and provide tests which cover:

1. Running jank with one or more custom PCH includes and verifying the code is loaded (by using it)
2. Putting those includes outside of the existing include paths so that `-I` needs to be used (to test that we're building the PCH with the correct include flags)

We're not testing the CLI flag system for these; we're testing the PCH building with custom includes. You can reach out to me for guidance on writing these if you get lost.

However, if writing these tests isn't doable for you right now, we can handle the rest of the review and then leave this PR parked until someone else can pick it up and implement the necessary testing.

cjbarre:
Closing in favor of https://github.com/jank-lang/jank/pull/602

PR #602, opened 2025-11-25
Add clang args that fix the potential for OS and local Clang build std libs conflicting with each other.

cjbarre:


jeaye:
This clearly breaks Linux, but can try only doing it for macOS.

```cpp
if constexpr(jtl::current_platform == jtl::platform::macos_like)
{
}
```

PR #607, opened 2025-11-26
Link to LLVM's custom libundwind

shantanu-sardesai:
On MacOS when we don't explicitly link to `libunwind` we are unable to catch exceptions across JIT frames.

```
...
testing file test/jank/form/throw/throw-top-level.jank => success
...
testing file test/jank/form/throw/throw-return-position.jank => libc++abi: terminating due to uncaught exception of type jank::runtime::oref<jank::runtime::object>
./bin/test: line 10: 13840 Abort trap: 6           "${here}/../build/jank-test" "$@"
```

### DIFF
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -112,7 +112,9 @@ set(
   )
 
 if (NOT jank_local_clang)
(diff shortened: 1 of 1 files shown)


jeaye:
Thanks!

PR #610, opened 2025-11-27
Implement support for typed catch clauses

jianlingzhong:
This change updates the try special form to support multiple, typed catch clauses, enabling more granular exception handling and C++ interop.

Key changes:

Syntax: catch clauses now require an explicit type argument (e.g., (catch cpp/jank.runtime.object_ref e ...)).
Parser: Updated parse_try to handle multiple catch clauses with their corresponding types.
Codegen: Extended LLVM IR generation to support multiple catch blocks.
Tests: Updated existing tests to use the new explicit catch syntax and added new tests for multiple catches, nested try-catch-finally blocks, etc..

### DIFF
--- a/compiler+runtime/src/cpp/jank/analyze/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/analyze/processor.cpp
@@ -2782,8 +2782,9 @@ namespace jank::analyze
               }
             }
 
-            /* Check for duplicate catch types - C++ does not allow multiple catch clauses
-             * for the same type as the first one would always match. */
+            /* Check for duplicate catch types. */
+            /*TODO Add full error handling for duplicated catch types:
+             * Add a new error kind and notes pointing to the duplicated types*/
             for(auto const &existing_catch : ret->catch_bodies)
             {
               if(existing_catch.type.data == catch_type_ref->type.data)


compiler+runtime/include/cpp/jank/analyze/expr/try.hpp
```diff
@@ -36,7 +38,7 @@ namespace jank::analyze::expr
     void walk(std::function<void(jtl::ref<expression>)> const &f) override;
 
     do_ref body;
-    jtl::option<catch_> catch_body{};
+    native_vector<jtl::option<catch_>> catch_bodies{};
```
jeaye: There's no need for an option here, now that we're using a vector. Before, we could have 0 or 1 catch body, so we used an option. Now we can have 0 or N, which the vector affords us.

compiler+runtime/include/cpp/jank/detail/to_runtime_data.hpp
```diff
@@ -100,4 +102,15 @@ namespace jank::detail
...
+  template <typename T>
+  object_ref to_runtime_data(native_vector<jtl::option<T>> const &m)
+  {
+    runtime::detail::native_persistent_vector ret;
+    for(auto const &e : m)
+    {
+      (void)ret.push_back(to_runtime_data(e));
+    }
+    return make_box<obj::persistent_vector>(ret);
+  }
```
jeaye, line 116: This can be removed by not using an option for catch bodies anyway.

compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
```diff
@@ -69,8 +69,21 @@ namespace jank::analyze::cpp_util
...
+  jtl::string_result<jtl::ptr<void>> resolve_literal_type(jtl::immutable_string const &literal);
+
   jtl::ptr<void> resolve_type(jtl::immutable_string const &sym, u8 const ptr_count)
   {
+    /* Clang canonicalizes "char" to "signed char" on some platforms, which breaks exception
+     * handling since they are distinct types. We use resolve_literal_type to get the
+     * exact type for "char". */
+    if(sym == "char")
+    {
+      if(auto const res = resolve_literal_type("char"); res.is_ok())
```
jeaye: 1. This is a very heavy-weight option. Resolving literal types like this requires JIT compiling C++ code, which is the slowest thing jank can do. What about just using `Cpp::GetType("char")`?
2. Please use direct initialization.

compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
```diff
@@ -69,8 +69,21 @@ namespace jank::analyze::cpp_util
     return type;
   }
 
+  jtl::string_result<jtl::ptr<void>> resolve_literal_type(jtl::immutable_string const &literal);
+
```
jeaye: This should not be needed, since it's declared in the header file, which we include.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2701,69 +2700,83 @@ namespace jank::analyze
...
+            /* Verify we have (catch cpp/type <sym> ...) */
             auto const catch_list(runtime::list(item));
             auto const catch_body_size(catch_list->count());
-            if(catch_body_size == 1)
+            if(catch_body_size < 3)
             {
               return error::analyze_invalid_try(
-                "A symbol is required after 'catch', which is used as the binding to "
-                "hold the exception value.",
+                "Catch clause requires a type, a symbol, and a body.",
```
jeaye: This isn't correct English anymore. Error messages must be very clear, with correct grammar. Also, a body is not required for a catch clause. Maybe something like:

> Each 'catch' form requires the type of exception to catch and a symbol for the name of the exception value.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2701,69 +2700,83 @@ namespace jank::analyze
...
+            auto const catch_sym_form(catch_it.first().unwrap());
+            auto const catch_type(analyze(catch_type_form, current_frame, position, fn_ctx, true));
+            if(catch_type.is_err() || catch_type.expect_ok()->kind != expression_kind::cpp_type)
+            {
+              return error::analyze_invalid_try(
+                       "An exception type required after 'catch'",
+                       object_source(item),
+                       error::note{
+                         "An exception type is required before this form.",
+                         object_source(catch_list->data.rest().rest().first().unwrap()),
```
jeaye: Isn't this just `catch_sym_form` or one of the other locals we have already?

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2701,69 +2700,83 @@ namespace jank::analyze
...
             }
+            auto catch_it(catch_list->data.rest());
+            auto const catch_type_form(catch_it.first().unwrap());
+            catch_it = catch_it.rest();
+            auto const catch_sym_form(catch_it.first().unwrap());
+            auto const catch_type(analyze(catch_type_form, current_frame, position, fn_ctx, true));
+            if(catch_type.is_err() || catch_type.expect_ok()->kind != expression_kind::cpp_type)
+            {
+              return error::analyze_invalid_try(
+                       "An exception type required after 'catch'",
```
jeaye: This is not correct English, since it's missing 'is' and a period at the end.

> An exception type is required after 'catch'.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2701,69 +2700,83 @@ namespace jank::analyze
...
+                "Catch clause requires a type, a symbol, and a body.",
                 object_source(item),
                 latest_expansion(macro_expansions));
             }
+            auto catch_it(catch_list->data.rest());
+            auto const catch_type_form(catch_it.first().unwrap());
+            catch_it = catch_it.rest();
+            auto const catch_sym_form(catch_it.first().unwrap());
+            auto const catch_type(analyze(catch_type_form, current_frame, position, fn_ctx, true));
+            if(catch_type.is_err() || catch_type.expect_ok()->kind != expression_kind::cpp_type)
```
jeaye: With this code, we're swallowing the analysis error for the type and just giving the same "An exception is required" error. This is not correct, since we have two cases.

1. We failed to analyze the catch type (could be the type was not found, or failed to instantiate, or whatever)
2. The form we did analyze is not a type

In the first case, we should return the actual analysis error.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2701,69 +2700,83 @@ namespace jank::analyze
...
+                         "An exception type is required before this form.",
+                         object_source(catch_list->data.rest().rest().first().unwrap()),
+                       },
+                       latest_expansion(macro_expansions))
+                ->add_usage(read::parse::reparse_nth(item, 1));
+            }
 
-            auto const sym_obj(catch_list->data.rest().first().unwrap());
-            if(sym_obj->type != runtime::object_type::symbol)
+            if(catch_sym_form->type != runtime::object_type::symbol)
```
jeaye, line 2744: The error for this (which you didn't change in this PR) is also missing an 'is'.

> A symbol is required after 'catch' ...

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2701,69 +2700,83 @@ namespace jank::analyze
...
+            auto const catch_sym(runtime::expect_object<runtime::obj::symbol>(catch_sym_form));
+            if(!catch_sym->get_namespace().empty())
             {
-              return error::analyze_invalid_try("The symbol after 'catch' must be unqualified.",
-                                                object_source(sym_obj),
-                                                latest_expansion(macro_expansions));
+              return error::analyze_invalid_try(
+                "The binding symbol in 'catch' must be unqualified.",
+                object_source(item),
+                latest_expansion(macro_expansions));
```
jeaye, line 2763: We want a reparse on this, too.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2701,69 +2700,83 @@ namespace jank::analyze
...
-            static auto const object_ref_type{ cpp_util::resolve_literal_type(
-                                                 "jank::runtime::oref<jank::runtime::object>")
-                                                 .expect_ok() };
-
-            ret->catch_body = expr::catch_{ sym,
-                                            object_ref_type,
-                                            static_ref_cast<expr::do_>(do_res.expect_ok()) };
+            do_res.expect_ok()->frame = catch_frame;
+            ret->catch_bodies.emplace_back(
+              expr::catch_{ catch_sym,
```
jeaye: If you get rid of the option, you can also get rid of the `expr::catch_{` here and just pass these values directly.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-char.jank
```diff
@@ -0,0 +1,13 @@
+;; Test catching C++ char exception
+(cpp/raw "void throw_char_z() { throw 'z'; }")
```
jeaye: These functions need to be wrapped in a namespace which is specific to the test. You're adding symbols into the global native namespace here, which could then conflict with other tests. This has led to issues in the past and it's not how we do it anymore. This should be:

```clojure
(cpp/raw "namespace jank::cpp::try_::pass_catch_cpp_char
          {
            void throw_char_z() { throw 'z'; }
          }")
```

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-char.jank
```diff
@@ -0,0 +1,13 @@
...
+
+(assert
+ (= \z
+    (try
+      (cpp/throw_char_z)
+      (catch cpp/int i
+        (println "Wrong: caught int")
```
jeaye: There's no need for a `println` in any of these tests. The assertion will handle this for us.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-pointer.jank
```diff
@@ -0,0 +1,16 @@
+;; Test catching C++ void* exception
+(cpp/raw "using void_ptr = void*;")
+(cpp/raw "void throw_void_ptr() { throw (void_ptr)0x1234; }")
+
+;; We can't easily assert the value of a void* in Jank yet without casting,
+;; but we can verify we caught the right type.
```
jeaye: 1. We can assert this; just define the pointer value in C++, throw the pointer, catch it, and use `(cpp/== p cpp/my_ptr)`
2. Please always write jank as lowercase within the jank repos

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-pointer.jank
```diff
@@ -0,0 +1,16 @@
...
+(let [caught (atom false)]
+  (try
+    (cpp/throw_void_ptr)
+    (catch cpp/int i
+      (println "Wrong: caught int"))
+    (catch cpp/void_ptr p
```
jeaye: Why are we using an alias here instead of `cpp/void*`?

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-exception.jank
```diff
@@ -0,0 +1,11 @@
+;; Test catching std::exception base class
+(cpp/raw "void throw_runtime_error_as_exception() { throw std::runtime_error(\"derived error\"); }")
+
+(let [caught (atom false)]
+  (try
+    (cpp/throw_runtime_error_as_exception)
+    (catch cpp/std.exception e
+      (reset! caught true)))
```
jeaye: It'd be great to grab `(cpp/.what e)` instead of just `true` here. That way we can verify the exception is valid. I know from experience, with LLVM IR, that it's too easy for things to look like they're working when they're not. That's why our tests should check actual values, too.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-invalid-argument.jank
```diff
@@ -0,0 +1,13 @@
+;; Test catching std::invalid_argument
```
jeaye: This test is redundant with the `bad_cast` test. We don't need both.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-logic-error.jank
```diff
@@ -0,0 +1,13 @@
+;; Test catching std::logic_error
```
jeaye: This test is redundant with the `bad_cast` and `invalid_argument` tests. We don't need all three.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-out-of-range.jank
```diff
@@ -0,0 +1,13 @@
+;; Test catching std::out_of_range
```
jeaye: This test is redundant with the `bad_cast`, `invalid_argument`, and `logic_error` tests. We don't need all four.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-runtime-error.jank
```diff
@@ -0,0 +1,13 @@
+;; Test catching std::runtime_error
```
jeaye: This test is redundant with the `bad_cast`, `invalid_argument`, `logic_error`, and `out_of_range` tests. We don't need all five.

compiler+runtime/test/jank/cpp/try/pass-multiple-catch-cpp-classes.jank
```diff
@@ -0,0 +1,53 @@
+;; Test multiple catch with C++ class types
```
jeaye: How is this test different from the ones above?

compiler+runtime/test/jank/cpp/try/pass-multiple-catch-cpp-with-finally.jank
```diff
@@ -0,0 +1,37 @@
+;; Test multiple catch with C++ exceptions and finally blocks
```
jeaye: We have tests for finally already (not in this PR). I don't think we need to repeat those here.

compiler+runtime/test/jank/cpp/try/pass-multiple-catch-mixed-types.jank
```diff
@@ -0,0 +1,32 @@
+;; Test mixing Jank and C++ exception types in multiple catch
```
jeaye: These tests also seem redundant.

compiler+runtime/test/jank/cpp/try/pass-rethrow-cpp-exception.jank
```diff
@@ -0,0 +1,16 @@
...
+
+(let [a (atom [])]
+  (try
+    (try
+      (cpp/throw_runtime_error)
+      (catch cpp/std.runtime_error e
+        (swap! a conj :caught-inner)
+        (cpp/throw_current))) ;; Rethrow using helper function
+    (catch cpp/std.runtime_error e
+      (swap! a conj :caught-outer)))
```
jeaye: Let's check `(cpp/.what e)` to verify we have a valid exception here.

compiler+runtime/test/jank/form/try/pass-multiple-catch-nested.jank
```diff
@@ -0,0 +1,23 @@
+;; Test nested try with multiple catches
+(assert (= :inner-caught
+           (try
+             (try
+               (throw :inner-exception)
+               (catch cpp/jank.runtime.object_ref e1
+                 :inner-caught)
+               (catch cpp/jank.runtime.object_ref e2
+                 :should-not-reach))
```
jeaye: This should not be possible. We have two catches of the same type. That is not permitted in C++ and we need to catch this during analysis.

compiler+runtime/test/jank/form/try/pass-multiple-catch-with-finally.jank
```diff
@@ -0,0 +1,25 @@
+;; Test multiple catch with finally block - first matching catch executes
```
jeaye: We already have tests for this. Why is this one needed?

compiler+runtime/test/jank/form/try/pass-multiple-catch.jank
```diff
@@ -0,0 +1,25 @@
+;; Test basic multiple catch with same type - first one always matches
```
jeaye: We already have tests for this. Why is this one needed?

compiler+runtime/test/jank/form/try/pass-nested-fn-finally.jank
```diff
@@ -0,0 +1,15 @@
+(let* [steps (atom [])
```
jeaye, line 1: We already have tests for this. Why is this one needed?

compiler+runtime/test/jank/form/try/pass-nested-try-with-finally.jank
```diff
@@ -0,0 +1,14 @@
+;; Test nested try blocks with finally
```
jeaye: We already have tests for this. Why is this one needed?

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-bad-cast.jank
```diff
@@ -0,0 +1,17 @@
...
+(cpp/raw "
+#include <typeinfo>
+void throw_bad_cast() {
+  throw std::bad_cast();
+}")
+
+(let [caught (atom false)]
+  (try
+    (cpp/throw_bad_cast)
+    (catch cpp/std.runtime_error e
```
jeaye: We're missing a test which catches an exception by its base verifies no slicing happened. 

Details: https://wiki.sei.cmu.edu/confluence/display/cplusplus/ERR61-CPP.+Catch+exceptions+by+lvalue+reference

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -73,6 +75,16 @@ namespace jank::codegen
     std::unique_ptr<llvm::PassInstrumentationCallbacks> pic;
     std::unique_ptr<llvm::StandardInstrumentations> si;
     llvm::ModulePassManager mpm;
+
+    struct exception_handler_info
+    {
+      llvm::BasicBlock *dispatch_block;
+      llvm::PHINode *ex_phi;
+      llvm::PHINode *sel_phi;
```
jeaye, line 83: Always default initialize raw pointers. Always.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -126,6 +138,38 @@ namespace jank::codegen
...
 
+    void route_unhandled_exception(llvm::Value *ex_ptr,
+                                   llvm::Value *selector,
+                                   llvm::BasicBlock *current_block) const;
+
+    void register_catch_clause_rtti(jtl::option<expr::catch_> const &catch_clause,
+                                    llvm::LandingPadInst *landing_pad,
+                                    native_set<llvm::Value *> &registered_rtti);
+
+    void generate_catch_block(size_t catch_index,
```
jeaye, line 149: Please use `usize` instead of `size_t`.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1669,17 +1714,294 @@ namespace jank::codegen
     return ret;
   }
 
+  void llvm_processor::impl::route_unhandled_exception(llvm::Value *ex_ptr,
+                                                       llvm::Value *selector,
+                                                       llvm::BasicBlock *current_block) const
```
jeaye, line 1720: These pointers can all be const.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1669,17 +1714,294 @@ namespace jank::codegen
...
+      /* No finally in current try. Check for parent handler. */
+      route_unhandled_exception(current_ex_ptr, current_sel, fallback_block);
+    }
+  }
+
+  void
+  llvm_processor::impl::register_parent_catch_clauses(llvm::LandingPadInst *landing_pad,
+                                                      native_set<llvm::Value *> &registered_rtti)
+  {
+    for(size_t i = 0; i < ctx->exception_handlers.size() - 1; ++i)
```
jeaye, line 1999: This can be a range-based loop.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1669,17 +1714,294 @@ namespace jank::codegen
...
+    }
+    else
+    {
+      /* No finally in current try. Check for parent handler. */
+      route_unhandled_exception(current_ex_ptr, current_sel, fallback_block);
+    }
+  }
+
+  void
+  llvm_processor::impl::register_parent_catch_clauses(llvm::LandingPadInst *landing_pad,
```
jeaye, line 1996: This pointer can be const.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -2434,7 +2666,22 @@ namespace jank::codegen
       args_array,
       sret
     };
-    ctx->builder->CreateCall(target_fn, ctor_args);
+    if(!lpad_and_catch_body_stack.empty())
+    {
+      llvm::BasicBlock *normal_dest
+        = llvm::BasicBlock::Create(*llvm_ctx,
+                                   "invoke.cxx.normal",
+                                   ctx->builder->GetInsertBlock()->getParent());
```
jeaye, line 2690: Please always use direct initialization.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1669,17 +1714,294 @@ namespace jank::codegen
     return ret;
   }
 
+  void llvm_processor::impl::route_unhandled_exception(llvm::Value *ex_ptr,
```
jeaye, line 1718: Starting here, we have nearly 300 lines of code which are not documented. There are a few comments, but they all say WHAT, not WHY. We need to do more here.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1669,17 +1714,294 @@ namespace jank::codegen
...
+  }
+
+  void
+  llvm_processor::impl::register_catch_clause_rtti(jtl::option<expr::catch_> const &catch_clause,
+                                                   llvm::LandingPadInst *landing_pad,
+                                                   native_set<llvm::Value *> &registered_rtti)
+  {
+    auto const catch_type{ catch_clause.unwrap().type };
+    auto const exception_rtti{ Cpp::MangleRTTI(catch_type) };
+    if constexpr(jtl::current_platform == jtl::platform::macos_like)
```
jeaye, line 1753: Why only macOS?

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1669,17 +1714,294 @@ namespace jank::codegen
...
+    auto const current_ex_ptr{ exception_phi };
+    auto const current_sel{ selector_phi };
+
+    auto const typeid_fn_type{
+      llvm::FunctionType::get(ctx->builder->getInt32Ty(), { ctx->builder->getPtrTy() }, false)
+    };
+    auto const typeid_fn{ llvm_module->getOrInsertFunction("llvm.eh.typeid.for.p0",
+                                                           typeid_fn_type) };
+
+    std::vector<llvm::BasicBlock *> catch_blocks;
```
jeaye, line 1915: Pleas use `native_vector` unless you're interacting with an API which requires `std::vector`.

jeaye (changes requested):
Thanks for all the work, Jianling. This is not ready for any more review, I think.

To start with, we have a LOT of new code here and not much documentation to follow it. Also, a lot of the new code doesn't match jank's style, for things you know we do. CI is also failing, since the tests don't pass.

The testing changes include a large amount of over-testing, which I suspect comes from an LLM rather than a conceived testing plan. I tried to identify the redundancies, but honestly there's so much that I know I missed some.

I know this feature is complex. That is all the more reason why we need to carefully craft it, though.

compiler+runtime/include/cpp/jank/detail/to_runtime_data.hpp
jianlingzhong: we don't have a "to_runtime_data" for native_vector, so we still need it.

compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
jianlingzhong: we cannot. using Cpp::GetType("char") would fail the following test:

```
(cpp/raw "void throw_char_z() { throw 'z'; }")

(assert
 (= \z
    (try
      (cpp/throw_char_z)
      (catch cpp/int i
        (println "Wrong: caught int")
        \0)
      (catch cpp/char c c))))
```

and the IR generated is using a "signed char"

```
  invoke void @clojure_core_G__24202(ptr null, i32 0, ptr null, ptr null)
          to label %invoke.cxx.normal unwind label %lpad

lpad:                                             ; preds = %entry
  %1 = landingpad { ptr, i32 }
          catch ptr @_ZTIi
          catch ptr @_ZTIa  # <-- signed char
```

if we use `resolve_literal_type("char")`, then the IR we got is:

```
  invoke void @clojure_core_G__24203(ptr null, i32 0, ptr null, ptr null)
          to label %invoke.cxx.normal unwind label %lpad

lpad:                                             ; preds = %entry
  %1 = landingpad { ptr, i32 }
          catch ptr @_ZTIi
          catch ptr @_ZTIc # <-- char
```

compiler+runtime/include/cpp/jank/detail/to_runtime_data.hpp
jeaye: Ok, in that case we need to fix the code. The `(void)` you added is to silence an error telling you that you're not using the data structure correctly. This is a **persistent** vector. You cannot change it. So `push_back` returns a new vector, which you're ignoring. In this code, `ret` starts empty and remains empty through the loop.

If you create a transient vector, you can mutate that before turning it into a persistent vector.

jeaye:
@jianlingzhong What's the status of this? Looks like CI has some failing `try` tests.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-pointer.jank
jianlingzhong: looks like we need a bit more work to support the `cpp/void*` syntax:

```
─ analyze/invalid-try ──────────────────────────────────────────────────────────────────────────────
error: Unable to find 'void' within the global namespace.                                           
                                                                                                    
     │ test/jank/cpp/try/pass-catch-cpp-pointer.jank                                                
  6  │ }")                                                                                          
  7  │                                                                                              
  8  │ (let [caught (atom false)]                                                                   
     │ ^ Expanded from this macro.                                                                  
  9  │   (try                                                                                       
 10  │     (cpp/jank.cpp.try_.pass_catch_cpp_pointer.throw_void_ptr)                                
 11  │     (catch cpp/int i)                                                                        
 12  │     (catch cpp/void* p                                                                       
     │                      ^ An exception type is required before this form.                       
     │            ^^^^^^^^^ Used here.                                                              

```

compiler+runtime/test/jank/cpp/try/pass-multiple-catch-cpp-with-finally.jank
jianlingzhong: removed.

compiler+runtime/test/jank/cpp/try/pass-multiple-catch-cpp-classes.jank
jianlingzhong: removed.

compiler+runtime/test/jank/cpp/try/pass-multiple-catch-mixed-types.jank
jianlingzhong: kept this one since this the only one that tests the mix of c++ and jank object exception types.

compiler+runtime/test/jank/form/try/pass-multiple-catch-nested.jank
jianlingzhong: removed.

compiler+runtime/test/jank/form/try/pass-multiple-catch.jank
jianlingzhong: removed.

compiler+runtime/test/jank/form/try/pass-nested-fn-finally.jank
jianlingzhong: this is the only one that tests try/catch across function boundaries.

compiler+runtime/test/jank/form/try/pass-nested-try-with-finally.jank
jianlingzhong: removed.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-bad-cast.jank
jianlingzhong: we will have test/jank/cpp/try/pass-catch-cpp-no-slicing.jank in the new pr.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
jianlingzhong: they cannot. they are used later in:

`parent_handler->ex_phi->addIncoming(xxx, xxx)`

and `addIncoming` does not like const pointers.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
jianlingzhong: only macos requires explicit registration: https://github.com/jank-lang/jank/blob/badd214cae7bf501b1cbf900ff29ebc624dfb8f9/compiler%2Bruntime/src/cpp/jank/codegen/llvm_processor.cpp#L1850

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
jianlingzhong: added some comments.

jianlingzhong:
closing in favor of #625

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-pointer.jank
jeaye: > looks like we need a bit more work to support the `cpp/void*` syntax:
> 
> ```
> ─ analyze/invalid-try ──────────────────────────────────────────────────────────────────────────────
> error: Unable to find 'void' within the global namespace.                                           
>                                                                                                     
> ─────┬──────────────────────────────────────────────────────────────────────────────────────────────
>      │ test/jank/cpp/try/pass-catch-cpp-pointer.jank                                                
> ─────┼──────────────────────────────────────────────────────────────────────────────────────────────
>   6  │ }")                                                                                          
>   7  │                                                                                              
>   8  │ (let [caught (atom false)]                                                                   
>      │ ^ Expanded from this macro.                                                                  
>   9  │   (try                                                                                       
>  10  │     (cpp/jank.cpp.try_.pass_catch_cpp_pointer.throw_void_ptr)                                
>  11  │     (catch cpp/int i)                                                                        
>  12  │     (catch cpp/void* p                                                                       
>      │                      ^ An exception type is required before this form.                       
>      │            ^^^^^^^^^ Used here.                                                              
> ─────┴──────────────────────────────────────────────────────────────────────────────────────────────
> ```

Please report this as a bug then, rather than just work around it. We need to know all of these issues so we can fix them.

jianlingzhong: i was not sure if this is valid syntax in clojure. i can fix it in #625 .

Issue #611, opened 2025-11-27
Missing logic for promotion to a hash map


shantanu-sardesai:
```jank
user=> (into {} (map #(vector % %) (range 50)))
Uncaught exception: Unable to expand array map to size 9. Be sure to check the size prior to insertion and promote to hash map if needed.
```

shantanu-sardesai:
@jeaye 

I'm assuming this is an alpha release ticket.

jeaye:
Yeah, I think so.

jeaye:
Read, and understand, the actual error message before you assume where the change should go. Array maps are currently putting the burden of promotion on the runtime functions. This is done so that we can use typed returns.

PR #612, opened 2025-11-28
Promote to hash map in `conj_in_place` in transient array map

shantanu-sardesai:
Resolves https://github.com/jank-lang/jank/issues/611.


shantanu-sardesai:
A test worthy case?

jeaye:
> A test worthy case?

We can unit test this in `test/cpp/jank/runtime/obj/transient_array_map.cpp`. We have other `obj` tests there which you can reference.

shantanu-sardesai:
> > A test worthy case?
> 
> We can unit test this in `test/cpp/jank/runtime/obj/transient_array_map.cpp`. We have other `obj` tests there which you can reference.

Done, also added the same tests for `persistent_array_map`.

jeaye:
Nice work!

PR #616, opened 2025-12-06
test(module-loader): bash tests

### LINKED ISSUE #363: Add bash tests for module loading
The module loader was built up before we had our bash test suite. We need to go back over that functionality in order to build up a test suite which will give us confidence that we can load and reload modules correctly in even the weirdest cases.

- [ ] Create a test plan
- [ ] Execute on the plan

These tests should include loading modules from source and binaries, reloading modules, out of date binaries, dependent modules, etc.

Samy-33:
# Bash tests

## .jank based modules

- [x] Overrides all .cljc, .cpp based modules if the module name is the same
- [x] Source

  - [x] loaded from filesystem: Tested through `test/bash/module/cljc`
  - [x] loading from a jar: Tested through `test/bash/module/jar` directory

## .cljc based modules

- [x] Loadable with correct reader conditional
- [x] Overriden by .jank based modules

## .cpp based modules

- [x] correctly loads
- [x] Overrides both .jank and .cljc based modules

## compiled modules (.o)

- [x] Source
  - [x] Should be loadable from filesystem
  - [x] Not loadable from a jar
- [x] If source file doesn't exist, fail to load
- [x] If source is more recent than the compiled module, use the source

## Module reloading

- [x] use :reload to reload a single module
  - [x] New var sare available after reloading
  - [x] Var available  but removed from source before reloading are still available
- [x] use :reload-all reloads the module as well as all it's dependencies
- [x] Reloading the modules should follow: https://github.com/clojure/tools.namespace?tab=readme-ov-file#reloading-code-motivation

## Dependency resolution

- [ ] All permutations of these types of module should be loadable.
  - [ ] .jank -> .cljc -> .cpp -> .o
  - [ ] .jank -> .cljc -> .o -> .cpp
  - And so on..

fixes #363

### DIFF
--- a/compiler+runtime/test/bash/module/default-loads/pass-test
+++ b/compiler+runtime/test/bash/module/default-loads/pass-test
@@ -1,4 +1,0 @@
-#!/usr/bin/env bash
-set -xeuo pipefail
-
-jank --module-path src run-main default-loads.core | grep '#{clojure.core clojure.core-native default-loads.core jank.compiler-native}'
--- a/compiler+runtime/include/cpp/jank/runtime/context.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/context.hpp
@@ -140,11 +140,6 @@ namespace jank::runtime
     native_unordered_map<jtl::immutable_string, native_vector<jtl::immutable_string>>
       module_dependencies;
 
-    /* TODO: Remove this in favor of calling module's `jank_load` functions
-     * on demand. At the moment it is being used to load all the compiled modules
-     * in ahead of time compiled binaries at startup (which is not an ideal way to
-     * achieve that). */
-    folly::Synchronized<native_deque<jtl::immutable_string>> loaded_modules_in_order;
     jtl::immutable_string binary_cache_dir;
     module::loader module_loader;
 
@@ -161,6 +156,11 @@ namespace jank::runtime
     /*** XXX: Everything here is thread-safe. ***/
     folly::Synchronized<native_unordered_map<obj::symbol_ref, ns_ref>> namespaces;
     folly::Synchronized<native_unordered_map<jtl::immutable_string, obj::keyword_ref>> keywords;
+
+    /* TODO: Remove this in favor of calling module's `jank_load` functions
+     * on demand. At the moment it is being used to load all the compiled modules
+     * in ahead of time compiled binaries at startup (which is not an ideal way to
+     * achieve that). */
     folly::Synchronized<native_deque<jtl::immutable_string>> loaded_modules_in_order;
     static thread_local native_list<thread_binding_frame> thread_binding_frames;
 
--- a/compiler+runtime/test/bash/module/default-loads/src/default_loads/core.jank
+++ b/compiler+runtime/test/bash/module/default-loads/src/default_loads/core.jank
@@ -1,3 +1,0 @@
-(ns default-loads.core)
-
-(defn -main [] (print (loaded-libs)))


compiler+runtime/src/cpp/jank/environment/check_health.cpp
```diff
@@ -390,12 +390,13 @@ namespace jank::environment
                                       [=] { util::cli::opts = saved_opts; }
       };
 
+      runtime::__rt_ctx->module_loader.set_is_loaded("/clojure.core");
```
jeaye: What're we trying to accomplish with this change?

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -954,11 +949,11 @@ namespace jank::runtime::module
       return res;
     }
 
-    loader::set_is_loaded(module);
```
jeaye, line 964: What's the reasoning for not having the loader manage this?

compiler+runtime/test/bash/module/compiled-modules/pass-test
```diff
@@ -0,0 +1,91 @@
...
+            [babashka.process :as proc]
+            [clojure.string :as string]
+            [clojure.test :as t :refer [deftest is testing use-fixtures]]))
+
+(def this-nsym (ns-name *ns*))
+
+(def module-a "jank-test.a")
+(def module-a-path "./src/jank_test/a.cljc")
+(defn compiled-a-path []
+  (first (fs/glob "./target" "**/a.o")))
```
jeaye: This is not needed if we compile the module with `-o some/path/a.o` and then just consistently use that path. That would make things much more reliable. Selecting the first result of a glob could land us in some trouble if there are multiple, since the ordering is not guaranteed to be based on date.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
Samy-33: I assume you're not just asking about _marking a module loaded_ but also _short circuiting if a module is loaded_ in the loader.

This is because if loader manages it, there would be two places that take the same responsibility i.e.
1. Loader itself
2. clojure.core

Major problem is with `:reload` implementation, if we short circuit, module will not be reloaded.

Clojure updates `*loaded-libs*` only in `clojure.core`. We should do the same IMO.

jeaye: If it's the responsibility of each `jank_load_foo` to mark its module as loaded, we have codegen changes which must be done, right? Since every module we AOT compile will need to include code which does that?

Samy-33: `jank_load_foo` wraps all the forms of the module. That includes the `(ns ...)` form as well. Which in turn expands and calls `load`. `load` includes the namespace into `*loaded-libs*`.

Tested: https://asciinema.org/a/Fmf3cFJTYQrGHGNRHRCX6CHBQ

We automatically include the namespace in the `*loaded-libs*`

jeaye: There's some some reasoning here that is incomplete or incorrect.

1. (info) You're saying we rely on `jank_load_foo` to use `ns`, which will mark a module as loaded
2. (info) `clojure.core` doesn't use `ns`, so we're specifically calling `set_is_loaded` for it
  a. (bug?) But in `check_health` we're calling `set_is_loaded` BEFORE calling `load_module`, which, I would think, could lead to short circuiting, since the loader will already consider `clojure.core` as loaded
3. (bug?) Although we're specifically calling `set_is_loaded` in `check_health`, we're not doing it in other place we're loading `clojure.core`, like the 5 places in `main.cpp`

It looks to me like `set_is_loaded` is only needed when we're loading a native module which doesn't use `ns`. But this PR is not doing a complete job of that, from what I can tell.

If this is something that's so easy to mess up, perhaps there's a better API design which would encourage correct usage.

Samy-33:
2) (bug?) no. Native `loader` doesn't short circuit anymore if a module is
loaded. It'll load the module again. Reason to mark `clojure.core` loaded
before actually loading it comes from how `ns` behaves. `ns` marks a module
loaded before it evaluates the rest of its body. So, doing the same for
`clojure.core`

3) That is a fair point. We should mark `clojure.core` as loaded at all
those places.

Let me know, if that makes sense.

jeaye:
> 3) That is a fair point. We should mark `clojure.core` as loaded at all those places. Let me know, if that makes sense.

Thanks for the explanation. Makes sense, but I think we should not have the manual `load_module` + `set_is_loaded` in half a dozen places. If we're doing it that often, there should be a fn for it.

Samy-33:
Agreed on having a separate function.

Samy-33:
I think, our worry about marking a module loaded within `load_module` is unnecessarily amplified.

1. For `clojure.core`, we need to evaluate it and even if we don't mark it loaded, that will not be a problem because it anyway will not be reloaded (why would one want that?). Now suppose, in case it is reloaded no issues on jank side. It's something the user did that results into re-evaluation of `clojure.core`. Nothing wrong.

2. For other modules that are loaded using `load_module`
     - If the module doesn't have `(ns )` then the module will not be marked loaded, but we don't lose anything. Clojure has the same behavior.
     - If the module has a `(ns )` then the module will anyway be marked loaded.

I've added a special function to load `clojure.core` to mimic the behavior of Clojure. For the other modules that the user loads using `context::load_module`, it doesn't matter if the module is marked loaded or not. Clojure's behavior also contributes to this point.

Samy-33:
I missed talking about native modules in my previous comment. As of now, they behave exactly the same way as `jank/cljc` based modules. If it doesn't specify a `(ns ..)` it'll be considered a `deviant`. The file will be loaded and `jank_load_` evaluated but nothing to be marked as loaded as only `namespaces` are marked loaded.

Same goes for a native module. A native module should have a call to `(ns..)` to be marked loaded. I know this may come as a surprise to the user. But, this is exactly the same behavior of `jank/cljc` based module and Clojure.

We might want to focus on a good documentation to ensure users know this. Later, if we decide to, we can think of a better API that diverges from Clojure but is an overall better experience for the user.

compiler+runtime/src/tmp.cljc
```diff
@@ -0,0 +1,5 @@
+(println "Hello")
```
jeaye: We don't need this file, right?

Samy-33: Removed.

Samy-33:
Rebased this branch over main.

Couple of changes
* Because we are relying on cpp codegen as well now, module loader needs to prefer `.cpp` over `.jank/.cljc` modules.
* In jank phase 2, use `jank_load_clojure_core` instead of the runtime C++ API.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -385,7 +403,6 @@ int main(int const argc, char const **argv)
 
 #ifdef JANK_PHASE_2
     jank_load_clojure_core();
-    __rt_ctx->module_loader.set_is_loaded("/clojure.core");
```
jeaye, line 389: How is `clojure.core` considered loaded, in phase 2? `jank_load_clojure_core` won't mark it as loaded.

Since I'm not getting this, I think it's clear we need to more documentation around this, in the form of a comment.

Samy-33: We don't need to mark clojure.core as loaded. There's no point.

1. Aligns with Clojure (doesn't mark clojure.core as loaded)
2. Namespaces loaded using `clojure.core/load` are marked loaded, whereas clojure.core is loaded using context::load_module.
3. clojure.core is never reloaded, unless forcefully done by requiring `clojure.core` again. (Which is not a supported use-case).

jeaye: I don't agree with this reasoning. `clojure.core` *is* loaded, by default. It's a special module for phase 2, but it's not special for phase 1. Even for phase 2, it's only special because we already have the `jank_load` function. That's it. If someone tries to check which modules are loaded, `clojure.core` should show up. Otherwise, we have this weird disconnect between modules, where the end user never knows which of them will actually be considered loaded.

This applies to more than just `clojure.core`, since the whole infrastructure for compiling core libs for phase 2 is inherently generalized to `build/core-libs` (not `build/clojure.core`). So we're going to end up doing this for other modules, too (like `clojure.string`).

> clojure.core is never reloaded, unless forcefully done by requiring clojure.core again. (Which is not a supported use-case).

Why is this not a supported use case? It works in Clojure.

compiler+runtime/test/bash/module/default-loads/pass-test
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+
+jank --module-path src run-main default-loads.core | grep '#{clojure.core clojure.core-native default-loads.core jank.compiler-native}'
```
jeaye: This is a very fragile test, since it's going to break whenever we default-load another ns. This is not asserting behavior, it's asserting state.

When writing a test, we need to ask ourselves one key question:

> Does this test improve my confidence in the system enough to cover the cost of the test?

In this case, my answer is no.

Samy-33: Given that `loaded-libs` is a public function, we should ensure its behavior of including the default namespaces as well as the ones that the user is loading.

How do we test this behavior if not via asserting the state?

jeaye:
Nice work, Saket! Thank you for sticking with this one to push it across the finish line.

PR #617, opened 2025-12-07
Update namespaces to support metadata

### LINKED ISSUE #605: `ns` doesn't handle meta properly
```bash
$ jank repl
user=> (+ 1 2 3)
6

user=> (ns foo {:a 1})
Uncaught exception: not nameable: [:a 1]

user=> (ns ^{:a 1} foo)
nil

foo=> (meta *ns*)
nil
```

Reported here: https://clojurians.slack.com/archives/C03SRH97FDK/p1764103529118349

djblue:
Fixes #605

```clojure
% jank repl
user=> (+ 1 2 3)
6

user=> (ns foo {:a 1})
nil

foo=>  (ns ^{:a 1} foo)
nil

foo=>  (meta *ns*)
{:jank/source {:file "/tmp/jank-repl-aeq2nT", :start {:offset 12, :line 1, :col 13}, :end {:offset 15, :line 1, :col 16}}, :a 1}
```

### DIFF
--- a/compiler+runtime/src/jank/clojure/core.jank
+++ b/compiler+runtime/src/jank/clojure/core.jank
@@ -1126,7 +1126,7 @@
   ([]
    0)
   ([x]
-   x)
+   (cpp/jank.runtime.number x))
   ([l r]
    (cpp/jank.runtime.add l r))
   ([l r & args]
@@ -1156,7 +1156,7 @@
   ([]
    1)
   ([x]
-   x)
+   (cpp/jank.runtime.number x))
   ([l r]
    (cpp/jank.runtime.mul l r))
   ([l r & args]
@@ -4455,8 +4455,7 @@
   "Returns a number one greater than num. Supports arbitrary precision.
   See also: inc"
   [x]
-  ;; (. clojure.lang.Numbers (incP x))
-  (throw "TODO: port inc'"))
+  (cpp/jank.runtime.promoting_inc x))
 
 ;; ;;math stuff
 ;; (defn ^:private nary-inline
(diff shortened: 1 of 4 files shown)


compiler+runtime/include/cpp/jank/runtime/behavior/metadatable.hpp
```diff
@@ -6,6 +6,11 @@
 
 namespace jank::runtime::behavior
 {
+  template <typename T>
```
djblue: I wonder if we should split `metadatable` into two separate concepts? Some objects in Clojure have metadata, but don't support `with-meta`, so it feels like we need a concept that says: this objects supports reading metadata, but says nothing about how that metadata is set.

Also, I know this should be in its own file, putting it here temporarily.

jeaye: I don't dig that separation. We should just use `metadatable`. If changing the meta is an error for namespaces, the `with_meta` function can throw.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -5220,6 +5193,70 @@
...
 
+(defmacro ns
+  "Sets *ns* to the namespace named by name (unevaluated), creating it
+  if needed.  references can be zero or more of: (:refer-clojure ...)
+  (:require ...) (:use ...) (:import ...) (:load ...) (:gen-class)
+  with the syntax of refer-clojure/require/use/import/load/gen-class
+  respectively, except the arguments are unevaluated and need not be
+  quoted. (:gen-class ...), when supplied, defaults to :name
+  corresponding to the ns name, :main true, :impl-ns same as ns, and
+  :init-impl-ns true. All options of gen-class are
```
jeaye, line 5204: Is this an updated version of `ns` from Clojure? Is that why it's longer now? Or did we strip out this stuff originally?

djblue: @jeaye It doesn't look like the [docstring](https://github.com/clojure/clojure/blame/master/src/clj/clojure/core.clj#L5817-L5837) has been updated recently, I just copied it verbatim from `clojure.core`. Do you want me to truncate the docstring back to the one before this PR?

jeaye: Nah, that's ok. We'll do a pass on the docstrings next year. I have big plans for them.

jeaye:
Nice work!

Issue #205, opened 2025-01-30
REPL returns `nil` or throws when no value provided


frenchy64:
Worth keeping mind when implementing `*1`.

```clojure
;; jank
clojure.core=> ;comment
nil

clojure.core=> *1
unbound@0x1094c7360 for var #'clojure.core/*1

clojure.core=> #_a
nil



;; clj
user=> ;comment
user=> *1
1
user=> #_a


<nothing happens>
```

Similiar oddity for `\<enter>`

```clojure
;; jank
clojure.core=> \
clojure.core=>...
nil


;; clj
user=> \

Syntax error reading source at (REPL:2:1).
EOF while reading character
```

Or even just pressing enter:

```clojure
;; jank
clojure.core=>
Exception: unable to map file /var/folders/s5/309818k92r92zn9c_nzvmnzc0000gn/T/jank-repl-e833.jank due to error: unable to map file



;; clj
user=>
user=>
```

jeaye:
`\` and then enter is a hack we have in place right now to allow multi-line REPL code. It should be replaced with logic to detect unexpected EOF and wait for more input.

The other notes here I think are valid bugs.

frenchy64:
Pressing enter now throws instead of `nil`.

```
clojure.core=>
Exception: unable to map file /var/folders/s5/309818k92r92zn9c_nzvmnzc0000gn/T/jank-repl-1ca7.jank due to error: unable to map file
```

Issue #558, opened 2025-10-22
Support catching specific exception types


jeaye:
Now that #345 is merged, we can add support for specific exception types. In jank, an exception type MUST be specified.

- [ ] Add support for exception types via `cpp/foo.bar` or `(cpp/type "foo::bar")` (just analyze the form and require it to be a `expr::cpp_type`)
- [ ] Add support for multiple `catch` forms, to catch specific types
- [ ] Add support for `:default` type which catches everything (like `catch(...)` in C++)
    - In this case, the exception value will always be `nil`
- [ ] Prevent incorrect scenarios like duplicate types (including through aliases) or `:default` multiple times

The default scenario will look now like this:

```clojure
(try
  (foo)
  (catch cpp/jank.runtime.object_ref o
    (println "caught" o)))
```

The catch-all scenario will look like this:

```clojure
(try
  (foo)
  (catch :default _
    (println "caught something")))
```

PR #618, opened 2025-12-13
REPL history support

shantanu-sardesai:
This PR adds:
- REPL history var support.
- No-op operations don't print anything in the REPL.

---

Resolves https://github.com/jank-lang/jank/issues/205.

### DIFF
--- a/compiler+runtime/src/cpp/main.cpp
+++ b/compiler+runtime/src/cpp/main.cpp
@@ -167,7 +167,7 @@ namespace jank
     auto const third_res_var{ __rt_ctx->find_var("clojure.core", "*3") };
     auto const error_var{ __rt_ctx->find_var("clojure.core", "*e") };
 
-    context::binding_scope scope{ obj::persistent_hash_map::create_unique(
+    context::binding_scope const scope{ obj::persistent_hash_map::create_unique(
       std::make_pair(first_res_var, jank_nil),
       std::make_pair(second_res_var, jank_nil),
       std::make_pair(third_res_var, jank_nil),


shantanu-sardesai:
@jeaye 

Clojure thread locally binds a bunch of vars inside it's `with-bindings` macro (seen [here](https://github.com/shantanu-sardesai/clojure/blob/35b579797d5a200a61d92edd237bfca583d847dc/src/clj/clojure/main.clj#L82-L99)) in the `main.clj` file. My PR only requires bindings for the REPL history vars so I have skipped the other ones.

shantanu-sardesai:
Currently I'm not binding the `*e` since I'm still thinking about how to hold the value for that var 🤔.

shantanu-sardesai:
Adding support for `no-op` type operations (mainly comments as of now) required some weird shenanigans, I wonder if it could be done better 🤔.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -173,11 +174,22 @@ namespace jank::runtime
...
+      }
+
       ret = evaluate::eval(expr);
 
       forms.emplace_back(form.expect_ok().unwrap().ptr);
     }
 
+    if(no_op)
+    {
+      return jtl::none;
```
jeaye: Just `none` should work fine.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -1,6 +1,7 @@
 #include <filesystem>
 #include <fstream>
 
+#include <jank/runtime/obj/persistent_hash_map.hpp>
```
jeaye: This include does not belong above LLVM includes. It belongs with the other `obj` includes.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -173,11 +174,22 @@ namespace jank::runtime
       auto const expr(analyze::pass::optimize(
         an_prc.analyze(form.expect_ok().unwrap().ptr, analyze::expression_position::statement)
           .expect_ok()));
+
+      if(no_op)
+      {
+        no_op = form.expect_ok().is_none();
```
jeaye: How could this be none here? We just did an `unwrap` of it on the previous line. It *has* to exist.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -159,12 +159,13 @@ namespace jank::runtime
     return eval_string(file.expect_ok().view());
   }
 
-  object_ref context::eval_string(jtl::immutable_string_view const &code)
+  jtl::option<object_ref> context::eval_string(jtl::immutable_string_view const &code)
```
jeaye, line 162: What does the `option` buy us? Why not just check for nil?

compiler+runtime/test/cpp/jank/jit/processor.cpp
```diff
@@ -19,6 +19,7 @@
 #include <jank/runtime/core/equal.hpp>
 #include <jank/analyze/processor.hpp>
 #include <jank/jit/processor.hpp>
+#include <jank/runtime/obj/nil.hpp>
```
jeaye: This include doesn't belong here.

compiler+runtime/src/cpp/jank/runtime/context.cpp
shantanu-sardesai: The idea behind converting the `object_ref` to `jtl::option<object_ref>` is to allow access to the cases where the parser has returned a `jtl::none` ([here](https://github.com/shantanu-sardesai/jank/blob/c6b783b56735536256a367417aa2aef023fa7594/compiler+runtime/src/cpp/jank/read/parse.cpp#L150), when we only have a [comment added](https://github.com/shantanu-sardesai/jank/blob/c6b783b56735536256a367417aa2aef023fa7594/compiler+runtime/src/cpp/jank/read/parse.cpp#L176) in the REPL as shown by one of the examples in [this ticket](https://github.com/jank-lang/jank/issues/205).) and such cases I am deeming as "no-op" cases.

There are two issues with overloading the use of `jank_nil` here:
- We can't distinguish between an actual result of `jank_nil` (ex. `(seq '())`, which comes post evaluation) and an actual "no-op" scenario (ex. `#_(comment)`, which comes during parsing.) when we are printing the result [here](https://github.com/shantanu-sardesai/jank/blob/c6b783b56735536256a367417aa2aef023fa7594/compiler+runtime/src/cpp/main.cpp#L213).
- _Personal opinion_, `jtl::option` felt like a more "official" approach to this issue than a runtime object like the `jank_nil`, although I'm also not a fan of how much change this required.

Am I missing something?

compiler+runtime/src/cpp/jank/runtime/context.cpp
shantanu-sardesai: True, although that makes me wonder how was `user=> ; comment` working fine in the REPL 🤔.

shantanu-sardesai: Besides that I've fixed the order of the ops.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -161,6 +164,19 @@ namespace jank
...
+    auto const third_res_var{ __rt_ctx->find_var("clojure.core", "*3") };
+    auto const error_var{ __rt_ctx->find_var("clojure.core", "*e") };
+
+    __rt_ctx
+      ->push_thread_bindings(
+        obj::persistent_hash_map::create_unique(std::make_pair(first_res_var, jank_nil),
+                                                std::make_pair(second_res_var, jank_nil),
+                                                std::make_pair(third_res_var, jank_nil),
+                                                std::make_pair(error_var, jank_nil)))
+      .expect_ok();
```
jeaye: There's no corresponding pop here. We don't really want to push, right? We want to actually just alter the roots of these vars? If Clojure JVM pushes for these, this is an unbounded memory leak, unless it has a corresponding pop.

compiler+runtime/src/cpp/jank/runtime/context.cpp
jeaye: Great explanation. I agree, we can keep the option.

compiler+runtime/src/cpp/main.cpp
shantanu-sardesai: Clojure uses the `binding` macro to set these vars, as can be seen [here](https://github.com/shantanu-sardesai/clojure/blob/35b579797d5a200a61d92edd237bfca583d847dc/src/clj/clojure/main.clj#L82-L99). I guess it's one of the advantages of implementing the setup around the REPL in Clojure instead of Java. Since the `binding` macro does have a corresponding "pop" in the `finally` block I'll add one for my usage as well.

shantanu-sardesai: Done.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -193,14 +207,24 @@ namespace jank
...
+        }
       }
       JANK_CATCH(jank::util::print_exception)
 
       input.clear();
       util::println("");
       le.setPrompt(get_prompt("=> "));
     }
+
+    __rt_ctx->pop_thread_bindings().expect_ok();
```
jeaye: This is not a `finally`. It will not run if an exception is thrown before we get here. We have a try/catch above, but there's still code after that which can throw. In general, we'd put this into a `scope_exit` instead, right after we push. However, we have `scope_binding` specifically for this use case. Please use that instead.

shantanu-sardesai: Done (I'm assuming you meant `binding_scope`). I really like the idea of `scope_exit` conceptually feels like the [defer](https://go.dev/tour/flowcontrol/12) statement of go. Although the `defer` statement is very specific to a use case, I suppose 🤔.

compiler+runtime/src/cpp/main.cpp
```diff
@@ -161,6 +162,17 @@ namespace jank
     std::string path_tmp{ tmp / "jank-repl-XXXXXX" };
     mkstemp(path_tmp.data());
 
+    auto const first_res_var{ __rt_ctx->find_var("clojure.core", "*1") };
+    auto const second_res_var{ __rt_ctx->find_var("clojure.core", "*2") };
+    auto const third_res_var{ __rt_ctx->find_var("clojure.core", "*3") };
+    auto const error_var{ __rt_ctx->find_var("clojure.core", "*e") };
+
+    context::binding_scope scope{ obj::persistent_hash_map::create_unique(
```
jeaye: CI will fail, since this can be const.

shantanu-sardesai: When will this become part of muscle memory 🙂. Done!

shantanu-sardesai:
@jeaye 

Will this PR also need to wait on https://github.com/jank-lang/jank/issues/558 since we also need to add support for `*e`.

jeaye:
> @jeaye
> 
> Will this PR also need to wait on #558 since we also need to add support for `*e`.

I think we can merge this without support for `*e`, but we should either leave the ticket open or create a new one to track `*e` support.

jeaye:
@shantanu-sardesai Did you test this locally? For me, it does not work, since `*1` and friends are never actually defined (and marked dynamic) anywhere.

shantanu-sardesai:
> Did you test this locally?

Yup, I did.

```jank
$ jank repl
user=> (str "Hello, " "World!")
"Hello, World!"

user=> (println *1 *2 *3)
Hello, World! nil nil
nil

user=> (println *1 *2 *3)
nil Hello, World! nil
nil

user=> (println *1 *2 *3)
nil nil Hello, World!
nil

user=> (println *1 *2 *3)
nil nil nil
nil
```

> since `*1` and friends are never actually defined (and marked dynamic) anywhere.

Sure they are, [here](https://github.com/shantanu-sardesai/jank/blob/badd214cae7bf501b1cbf900ff29ebc624dfb8f9/compiler+runtime/src/jank/clojure/core.jank#L6611-L6625) in the `clojure.core`. I've kept them there (as they were before) since it's how Clojure also does this. Do we need to take a different approach?

PR #619, opened 2025-12-15
fix: Use Global CLI Opts for Direct-Call

stmonty:
~~The runtime context was default-constructing its CLI options, never copying the parsed values. As a result
 ` __rt_ctx->opts.direct_call` was false, so the direct-call code path never runs. Initializing `context::opts` from
  `util::cli::opts` on construction fixes this and seemed to make direct-call work again when tested locally.~~

Now that the CLI options are a global, we don't need the runtime context to hold onto their copy. Because the runtime context was holding onto a copy before CLI options were being parsed, the copy on the runtime context had the default flags and therefore it was not possible to get direct-calling. This fixes this.

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
@@ -41,6 +41,8 @@ namespace jank::runtime
   object_ref add(obj::ratio_ref l, obj::integer_ref r);
   obj::ratio_ref add(obj::integer_ref l, obj::ratio_ref r);
 
+  object_ref promoting_add(object_ref const l, object_ref const r);
+
   object_ref sub(object_ref l, object_ref r);
   object_ref sub(obj::integer_ref l, object_ref r);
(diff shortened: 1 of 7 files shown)


compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -44,6 +44,7 @@ namespace jank::runtime
     /* We want to initialize __rt_ctx ASAP so other code can start using it. */
     : binary_version{ (__rt_ctx = this, util::binary_version()) }
     , binary_cache_dir{ util::binary_cache_dir(binary_version) }
+    , opts{ util::cli::opts }
```
jeaye: We should not be copying the opts.

The issue wasn't that we were never initializing the copy of the opts. The issue was that we had a copy of the opts in the first place.

jeaye (changes requested):

jeaye (approved):
Good find! That might explain why direct calls didn't work when I tried 'em.

Issue #622, opened 2025-12-21
jank prints the last form result in a file


shantanu-sardesai:
Not a big problem but jank prints the evaluation result of the last for in a file to the console which Clojure doesn't do.

The `example.cljc` file:

```cljc
42
```

Clojure:

```sh
$ clj -M example.cljc

```

jank:

```sh
$ jank run example.cljc
42
```

shantanu-sardesai:
It was kinda funny as I was trying to see if the [man-or-boy test](https://rosettacode.org/wiki/Man_or_boy_test#Clojure) passes for jank and just saw the number `-67`... I thought we had a bug in the compiler for a few secs before realisation hit me 😆!

PR #623, opened 2025-12-21
Stop printing results in `run`

shantanu-sardesai:
Resolves https://github.com/jank-lang/jank/issues/622.


jeaye:
This'll cause a conflict with #618, I think.

shantanu-sardesai:
I'll rebase.

PR #625, opened 2025-12-22
Implement support for multiple typed catch clauses

jianlingzhong:
## Basic Functionality
Core mechanics of exception handling and return values.

*   **[pass-try-basic.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-basic.jank)**
    *   **Purpose**: Tests that `try` blocks return the last expression in the body (success case) or the catch block (error case). Also verifies `try` works as an expression in other forms.
    *   **Uniqueness**: Fundamental baseline test.
*   **[pass-try-empty-bodies.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-empty-bodies.jank)**
    *   **Purpose**: Tests edge cases with empty `try`, `catch`, and `finally` bodies to ensure the parser handles them gracefully (usually returning `nil`).
    *   **Uniqueness**: Validates parser resilience against valid but degenerate code.
*   **[pass-unboxed-position.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-unboxed-position.jank)**
    *   **Purpose**: Verifies that `try` expressions correctly handle unboxed return types (e.g., returning a raw primitive number) when used in a context expecting a value.
    *   **Uniqueness**: Critical for the compiler's type analysis and boxing/unboxing logic.

## Control Flow & Execution Order
Verifying that code executes in the strict sequence defined by the language.

*   **[pass-try-catch-finally-execution.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-catch-finally-execution.jank)**
    *   **Purpose**: asserts the strict `try -> catch -> finally` execution order using side effects (vector accumulation).
    *   **Uniqueness**: The definitive integration test for the full lifecycle of a handled exception.
*   **[pass-catch-ordering.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-catch-ordering.jank)**
    *   **Purpose**: Tests that the *first* matching `catch` clause is selected, not the "best" match. Crucial for C++ inheritance (e.g., catching `Base` before `Derived`).
    *   **Uniqueness**: Verifies linear scan semantics of catch clauses.
*   **[pass-finally-behavior.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-finally-behavior.jank)**
    *   **Purpose**: Verifies `finally` blocks execute for side effects but do *not* affect the return value of the `try` expression (unless they throw).
    *   **Uniqueness**: Distinguishes `finally` (cleanup) from `catch` (recovery).
*   **[pass-throw-from-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-throw-from-finally.jank)**
    *   **Purpose**: Tests that throwing an exception from a `finally` block overrides any in-flight exception or return value.
    *   **Uniqueness**: Verifies "abrupt completion" semantics of `finally`.

## C++ Interop & Types
Coverage for catching native C++ types.

*   **[pass-catch-cpp-primitives.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-primitives.jank)**
    *   **Purpose**: Catching fundamental C++ types: `char`, `double`, `float`, `int`, `unsigned int`.
    *   **Uniqueness**: Ensures the runtime and usages of `throw/catch` work for non-object primitives.
*   **[pass-catch-cpp-string-literal.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-string-literal.jank)**
    *   **Purpose**: Catching `const char*` (string literals).
    *   **Uniqueness**: Verifies pointer matching specifically for C-strings.
*   **[pass-catch-cpp-pointer.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-pointer.jank)**
    *   **Purpose**: Catching `void*` and strictly defined pointer types.
    *   **Uniqueness**: Ensures pointer identity is preserved across the catch boundary.
*   **[pass-catch-cpp-std-exceptions.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-exceptions.jank)**
    *   **Purpose**: Catching standard library exceptions (`std::runtime_error`, `std::bad_cast`) and validating polymorphic catching (catching via base `std::exception`).
    *   **Uniqueness**: Verifies RTTI-based hierarchy matching for C++ objects.
*   **[pass-catch-cpp-no-slicing.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-no-slicing.jank)**
    *   **Purpose**: Catches a derived object by a base class value/reference to ensure the underlying object is not "sliced" or corrupted.
    *   **Uniqueness**: Critical for C++ object fidelity.

## Complex Semantics & Nesting
Advanced structural tests.

*   **[pass-nested-try-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-nested-try-catch.jank)**
    *   **Purpose**: Standard nesting: `try` within `try`. Verifies exceptions bubble up to the nearest enclosing matching catch.
    *   **Uniqueness**: Verifies scope definition and exception propagation.
*   **[pass-nested-no-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-nested-no-catch.jank)**
    *   **Purpose**: Nested `try` blocks where intermediate levels have *only* `finally` blocks and no `catch`.
    *   **Uniqueness**: Stress-tests stack unwinding through layers that perform cleanup but do not stop the exception.
*   **[pass-nested-fn-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-nested-fn-finally.jank)**
    *   **Purpose**: Exceptions propagating across function boundaries, triggering `finally` blocks in both the callee and caller.
    *   **Uniqueness**: Verifies that unwinding correctly traverses the call stack (not just lexical scope).
*   **[pass-try-in-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-in-catch.jank)**
    *   **Purpose**: A full `try` block nested *inside* a `catch` handler.
    *   **Uniqueness**: Validates that the catch block itself is a valid code context for new exception scopes (recovery logic).
*   **[pass-try-in-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-in-finally.jank)**
    *   **Purpose**: A full `try` block nested *inside* a `finally` handler.
    *   **Uniqueness**: Validates complex cleanup logic that might itself need error handling.

## Mixed Environment
Interaction between Jank and C++ exceptions.

*   **[pass-catch-mixed.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-mixed.jank)**
    *   **Purpose**: A single `try` block with both `catch cpp/...` and `catch jank.runtime.object_ref`.
    *   **Uniqueness**: Verifies that the runtime can dispatch to either backend logic from a single site.
*   **[pass-catch-rethrow-alloc.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-catch-rethrow-alloc.jank)**
    *   **Purpose**: Catching an exception object and immediately re-throwing *that same object*.
    *   **Uniqueness**: Verifies object identity and lifetime during re-throw (distinct from `throw;` which re-throws the *active* exception).
*   **[pass-rethrow.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-rethrow.jank)**
    *   **Purpose**: Tests `throw` with no arguments (C++ `throw;`) to re-raise the currently active exception.
    *   **Uniqueness**: Verifies access to thread-local exception state.
*   **[pass-uncaught-cpp-exception-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-uncaught-cpp-exception-finally.jank)**
    *   **Purpose**: Verifies `finally` executes even when a C++ exception is *not* caught by any local catch clause.
    *   **Uniqueness**: Ensures cleanup reliability for unhandled native exceptions.
*   **[pass-destructor-unwinding.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-destructor-unwinding.jank)**
    *   **Purpose**: Confirms that C++ destructors for local stack objects are invoked when a Jank `try/catch` unwinds the stack.
    *   **Uniqueness**: Mandatory for RAII correctness/memory safety.

## Recursion & Scoping
Interaction with `recur` and closures.

*   **[pass-capture-within-let.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-capture-within-let.jank)**
    *   **Purpose**: `try` nested inside `let`, verifying it can see outer bindings.
    *   **Uniqueness**: Basic lexical scoping check.
*   **[pass-closure-capture.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-closure-capture.jank)**
    *   **Purpose**: Catch blocks capturing outer function parameters and closure variables.
    *   **Uniqueness**: Essential for functional programming patterns where handlers rely on context.
*   **[pass-recur-local-to-try.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-recur-local-to-try.jank)**
    *   **Purpose**: Using `recur` to jump to a loop head *inside* the `try` block.
    *   **Uniqueness**: Verifies `recur` doesn't get confused by the try protection boundary.
*   **[pass-recur-local-to-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-recur-local-to-catch.jank)**
    *   **Purpose**: Using `recur` to jump to a loop head *inside* the `catch` block.
    *   **Uniqueness**: Verifies `recur` works in exception handler scope.
*   **[pass-recur-local-to-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-recur-local-to-finally.jank)**
    *   **Purpose**: Using `recur` to jump to a loop head *inside* the `finally` block.
    *   **Uniqueness**: Verifies `recur` works in cleanup scope.

## Compile-Time Validation (Negative Tests)
Ensuring invalid code is rejected with meaningful errors.

*   **[fail-catch-after-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-catch-after-finally.jank)**: Enforces strict [(catch ...)* (finally ...)?](file:///Users/jianlingzhong/personal/jank/compiler+runtime/dump_units.py#53-90) ordering.
*   **[fail-duplicate-catch-type.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-duplicate-catch-type.jank)**: Enforces that you can't have two `catch` clauses for the exact same native type.
*   **[fail-multiple-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-multiple-finally.jank)**: Enforces max one `finally` block.
*   **[fail-other-form-after-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-other-form-after-finally.jank)**: Enforces `finally` is the last element.
*   **[fail-recur-in-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-recur-in-catch.jank)**: (Compiler Constraint) Currently validates restrictions on recurring *out* of a catch block if unsupported, or semantics thereof.
*   **[fail-recur-in-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-recur-in-finally.jank)**: (Compiler Constraint) Validates restrictions on recurring *out* of a finally block.
*   **[fail-recur-in-try.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-recur-in-try.jank)**: (Compiler Constraint) Validates restrictions on recurring *across* the try boundary (e.g., from try to outside loop).


compiler+runtime/include/cpp/jank/codegen/llvm_processor.hpp
```diff
@@ -108,6 +108,7 @@ namespace jank::codegen
     llvm_processor(analyze::expr::function_ref const expr,
                    jtl::immutable_string const &module,
                    compilation_target target);
+    llvm_processor(analyze::expr::function_ref expr, std::unique_ptr<reusable_context> ctx);
```
jeaye: Looks like this is not defined anywhere. I think it can be deleted.

compiler+runtime/include/cpp/jank/runtime/context.hpp
```diff
@@ -114,6 +114,7 @@ namespace jank::runtime
     jtl::immutable_string unique_namespaced_string() const;
     jtl::immutable_string unique_namespaced_string(jtl::immutable_string const &prefix) const;
     jtl::immutable_string unique_string() const;
+    jtl::immutable_string unique_munged_string() const;
```
jeaye: I removed this function. Looks like you added it back. Please remove it and use `unique_string` and `munge` separately.

compiler+runtime/include/cpp/jank/runtime/core/make_box.hpp
```diff
@@ -9,9 +9,19 @@
...
 namespace jank::runtime
 {
+  template <typename T>
+  requires(!behavior::object_like<T> && !std::is_same_v<std::remove_const_t<T>, char>)
+  [[gnu::flatten, gnu::hot, gnu::visibility("default")]]
+  inline auto make_box(T * const d)
+  {
+    return make_box<obj::native_pointer_wrapper>(
+      const_cast<void *>(reinterpret_cast<void const *>(d)));
+  }
```
jeaye: I don't think we should have a `make_box` which swallows any old pointer. This is too magical to be a good API. If someone wants a `native_pointer_wrapper`, they can use `make_box<native_pointer_wrapper>`.

compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
```diff
@@ -308,15 +319,16 @@ namespace jank::analyze::cpp_util
...
-          {
-            alias_name += "*";
-          }
-          return alias_name;
-        }
+        /* Type aliases already include pointer/reference semantics in their definition.
+         * Don't add extra `*` for pointer type aliases like `using void_ptr = void*;`
+         * as that would make it `void_ptr*` (void**) instead of `void_ptr` (void*).
+         *
+         * Use printQualifiedName with the default policy to get the fully qualified name. */
```
jeaye: This comment is clearly written by AI and it's not a good comment.

1. It describes the change made to the code (we used to add `*` for pointers)
2. It says exactly WHAT the following code will do; comments should focus on WHY

There is no need for a comment here unless there is a WHY to explain.

compiler+runtime/src/cpp/jank/analyze/local_frame.cpp
```diff
@@ -94,13 +94,6 @@ namespace jank::analyze
       res.first->second.has_boxed_usage = true;
       /* To start with, we assume it's only boxed. */
       res.first->second.has_unboxed_usage = false;
-
-      /* Native values which are captured get auto-boxed, so we need to adjust the type
-       * of the binding. */
-      if(!cpp_util::is_any_object(res.first->second.type))
-      {
-        res.first->second.type = cpp_util::untyped_object_ptr_type();
-      }
```
jeaye, line 103: Why was this removed? It seems unrelated to exeption handling.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -1619,6 +1619,13 @@ namespace jank::analyze
     }
 
     auto const params(runtime::expect_object<runtime::obj::persistent_vector>(params_obj));
+    /* Using recur from this function is fine, even if we're in a try, since the jump target
+     * is this function. */
+    __rt_ctx
+      ->push_thread_bindings(runtime::obj::persistent_hash_map::create_unique(
+        std::make_pair(__rt_ctx->no_recur_var, runtime::jank_false)))
+      .expect_ok();
+    util::scope_exit const finally{ []() { __rt_ctx->pop_thread_bindings().expect_ok(); } };
```
jeaye: We have `binding_scope` for this, which handles the push + pop for us, given a scope.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2350,12 +2357,20 @@ namespace jank::analyze
                           jtl::option<expr::function_context_ref> const &fn_ctx,
                           bool const)
   {
-    /* When we analyze a loop, we actually just set a flag and then analyze a let.
```
jeaye, line 2353: Why was this first sentence changed?

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2350,12 +2357,20 @@ namespace jank::analyze
...
+    /* Using recur from this loop is fine, even if we're in a try, since the jump target
+     * is this loop. */
+    __rt_ctx
+      ->push_thread_bindings(runtime::obj::persistent_hash_map::create_unique(
+        std::make_pair(__rt_ctx->no_recur_var, runtime::jank_false)))
+      .expect_ok();
+    util::scope_exit const finally{ [&]() {
+      loop_details = old_loop_details;
+      __rt_ctx->pop_thread_bindings().expect_ok();
+    } };
```
jeaye: We can use `binding_scope` here.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2696,69 +2710,128 @@ namespace jank::analyze
...
+                       "The binding symbol in 'catch' must be unqualified.",
+                       object_source(item),
+                       latest_expansion(macro_expansions))
+                ->add_usage(read::parse::reparse_nth(item, 2));
+            }
+            auto const catch_type_ref(static_ref_cast<expr::cpp_type>(catch_type.expect_ok()));
+            /* If we're catching a C++ class/struct by value, we want to promote it to a reference
+             * to avoid object slicing and to enable polymorphism.
+             * However, we must NOT promote types in the jank::runtime namespace (like object_ref),
+             * as these are smart pointers expected to be passed by value in the runtime. */
```
jeaye, line 2811: Is there any reason why we can't catch `object_ref` by reference? Just because you catch it by reference doesn't mean you can't then pass it by value. This whole branch seems unnecessary.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -1003,7 +1003,7 @@ namespace jank::runtime::module
     else
     {
       __rt_ctx->jit_prc.eval_string(
-        util::format("void* {}(); {}();", load_function_name, load_function_name));
+        util::format("extern \"C\" void* {}(); {}();", load_function_name, load_function_name));
```
jeaye: The return type here should be `void`. All `jank_load` fns now return `void`.

compiler+runtime/test/bash/error-reporting/bin/jank/test/error_reporting.clj
```diff
@@ -62,7 +63,7 @@
     (doseq [test tests]
       (print "testing dir" (b.f/file-name (:dir test)) "=> ")
       (let [expected (try
-                       (slurp (:output-file test))
+                       (strip-ansi-codes (string/trim (slurp (:output-file test))))
```
jeaye: This should not be needed. On line 52, we strip ansi codes before writing. If you manually edited the file, instead of using the `generate` command, you may run into this issue, but not with correct operation.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-exceptions.jank
```diff
@@ -0,0 +1,40 @@
+;; Test catching various C++ std exception types
```
jeaye, line 1: This test file doesn't need to exist. We already test for base catching in the splicing test. We already test for mixed catches in a different test.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-pointer.jank
```diff
@@ -0,0 +1,16 @@
+;; Test catching C++ void* exception
+(cpp/raw "namespace jank::cpp::try_::pass_catch_cpp_pointer {
+  using void_ptr = void*;
+  void_ptr const thrown_ptr = (void_ptr)0x1234;
```
jeaye, line 5: Please use direct initialization.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-no-slicing.jank
```diff
@@ -0,0 +1,34 @@
+;; Test to verify that object slicing does not occur when catching by base type
+;; See ERR61-CPP: Catch exceptions by lvalue reference
+
+(cpp/raw "
+#include <string>
+namespace jank::cpp::try_::pass_catch_no_slicing
```
jeaye, line 5: The ns should be `pass_catch_cpp_no_slicing`.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-no-slicing.jank
```diff
@@ -0,0 +1,34 @@
+;; Test to verify that object slicing does not occur when catching by base type
+;; See ERR61-CPP: Catch exceptions by lvalue reference
+
+(cpp/raw "
+#include <string>
```
jeaye, line 4: There's no need for this include.

compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-exceptions.jank
```diff
@@ -0,0 +1,40 @@
...
+(cpp/raw "
+#include <stdexcept>
+#include <typeinfo>
+#include <vector>
```
jeaye, line 6: There should be no need for these includes.

compiler+runtime/test/jank/cpp/try/pass-destructor-unwinding.jank
```diff
@@ -0,0 +1,40 @@
...
+
+  struct DestructorTracker {
+    DestructorTracker() {}
+    ~DestructorTracker() {
+      destructor_count++;
+    }
+  };
+
+  void throw_with_unwinding() {
+    DestructorTracker tracker;
```
jeaye, line 16: This could be a better test. We're only testing the C++ compiler here, since `tracker` exists in C++ land instead of jank land. What we really want to test here is that dtors run for values within both jank land and C++ land. In particular:

1. Values within a `let` that's within a `try`
2. Values within a `let` within a `fn` when the exception escapes the fn for a higher `catch`

compiler+runtime/test/jank/form/try/pass-closure-capture.jank
```diff
@@ -0,0 +1,31 @@
+;; Test closure capture in try-catch blocks
```
jeaye, line 1: There is no closure capturing going on here. `try` no longer creates a wrapper fn, since we have codegen (IR and C++) to generate `try` directly. So this whole test is not testing what it's saying it's testing. I think the whole thing can be deleted.

compiler+runtime/test/jank/form/try/pass-nested-try-catch.jank
```diff
@@ -0,0 +1,25 @@
...
+          :inner-caught))
+      (catch cpp/jank.runtime.object_ref outer-e
+        :outer-caught))))
+
+;; Test 2: Inner catch rethrows to outer catch
+(assert
+ (= :success
+    (try
+      (try
+        (throw :success)
```
jeaye, line 19: I think it makes sense to throw something else, so we know the `:success` comes from line 23.

compiler+runtime/test/jank/form/try/pass-recur-local-to-finally.jank
```diff
@@ -0,0 +1,14 @@
...
+(try
+  :ignored
+  (finally
+    (loop [i 0]
+      (if (< i 3)
+        (do
+          (swap! a conj i)
+          (recur (inc i)))
+        :success))))
+(assert (= @a [0 1 2]))
```
jeaye, line 13: You check the value of `a` here, but not for the loop within catch test or loop within try test. I think it makes sense to do it for those tests, too.

compiler+runtime/test/jank/form/try/pass-try-basic.jank
```diff
@@ -0,0 +1,29 @@
+;; Basic try/catch form logic tests
```
jeaye, line 1: This test is overly general. What is *basic* execution, when we have so many different specific tests? Does this test actually cover something distinct from the other tests? If so, let's remove everything else and call it that. If not, let's delete this test.

compiler+runtime/test/jank/form/try/pass-try-catch-finally-execution.jank
```diff
@@ -0,0 +1,12 @@
+;; Comprehensive tests for try-catch-finally execution order
```
jeaye, line 1: This is already covered by so many other tests.

compiler+runtime/test/jank/form/try/pass-unboxed-position.jank
```diff
@@ -1,10 +1,12 @@
-; A box will be required.
```
jeaye, line 1: This test no longer makes sense and should be deleted. It pertained to an older unboxing mechanism we used to have.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -75,24 +75,7 @@ namespace jank::codegen
...
-      {
-        return existing->second;
-      }
-
-      auto const &native_name{ runtime::munge(__rt_ctx->unique_string("const")) };
-      lifted_constants.emplace(o, native_name);
-      return native_name;
-    }
-
+    [[maybe_unused]]
```
jeaye: Looks like a bad merge was done in this file. On main:

1. `lift_constant` is needed
2. `gen_constant_type` is used
3. `boxed_local_name` and other boxing-related code has been deleted
4. We should be using `unique_string` instead of `unique_namespaced_string`
5. Perhaps others.

This merge is a mess. I'm not sure what you're using for resolving conflicts, but perhaps the tool is the problem.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -909,40 +949,53 @@ namespace jank::codegen
...
+      ret = munged_name;
+    }
+    else if(is_unboxed_type)
+    {
+      ret = handle{ "", munged_name };
+    }
+    else
+    {
+      ret = handle{ detail::boxed_local_name(munged_name), munged_name };
+    }
```
jeaye: I suspect this is part of the bad merge.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1351,16 +1407,31 @@ namespace jank::codegen
...
     util::format_to(body_buffer, "{");
     if(expr->finally_body.is_some())
     {
-      util::format_to(body_buffer, "jank::util::scope_exit const finally{ [&](){ ");
+      util::format_to(body_buffer, "auto const finally{ [&](){ ");
       gen(expr->finally_body.unwrap(), fn_arity);
       util::format_to(body_buffer, "} };");
+      /* We use a local struct for RAII to ensure the finally block is called even
+     * if the try block returns. We can't use jank::util::scope_exit because it
+     * swallows exceptions. */
```
jeaye: RAII guarantees that dtors get called, even if the try block returns. Why do we need a bool for tracking if this is active? This feels very backward to me, but perhaps I'm missing something.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1541,15 +1614,15 @@ namespace jank::codegen
       cpp_util::get_qualified_type_name(
         Cpp::GetTypeWithoutCv(Cpp::GetNonReferenceType(expr->conversion_type))),
       (expr->policy == conversion_policy::into_object ? "into_object" : "from_object"),
-      value_tmp.unwrap().str(true));
+      value_tmp.unwrap().str(expr->policy == conversion_policy::from_object));
```
jeaye: None of this box tracking stuff is needed. It's all part of the bad merge.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -2028,67 +2113,41 @@ namespace jank::codegen
     {
       profile::timer const timer{ util::format("cpp gen {}", root_fn->name) };
 
-      /* Module targeting works in a special way, with the goal of
-       * cutting down the generated code size. Instead of each function
-       * having its own lifted vars/constants, we have one namespace for
```
jeaye, line 2033: This is all needed. Bad merge.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -2839,6 +3313,38 @@ namespace jank::codegen
     return ret;
   }
 
+  /* We generate a static wrapper function to return the address of the static member variable.
+   * This allows us to access the variable from the JIT, as we cannot take the address of
+   * a static member directly if it is not defined in the current module. */
+  llvm::Function *llvm_processor::impl::ensure_static_wrapper(jtl::ptr<void> scope)
```
jeaye: This is something which should be handled by CppInterop, I think. That's what we're already using to reach into C++ from IR.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -2222,6 +2676,26 @@ namespace jank::codegen
       return alloc;
     }
 
+    /* For static variables that are non-copyable, we can't use MakeAotCallable (which copies).
```
jeaye, line 2694: `MakeAotCallable` should not be making copies of member variables. It should be referencing them. If it's not, we should fix CppInterop. `MakeAotCallable` is something we created just for jank, anyway.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -384,12 +434,7 @@ namespace jank::codegen
                            && !(Cpp::IsArrayType(Cpp::GetNonReferenceType(input_type))) };
     auto const is_arg_ptr{ Cpp::IsPointerType(input_type)
                            || (Cpp::IsArrayType(Cpp::GetNonReferenceType(input_type))) };
-    //util::println(
-    //  "convert_object input_type = {}, output_type = {}, conversion_type = {}, policy = {}",
-    //  Cpp::GetTypeAsString(input_type),
-    //  Cpp::GetTypeAsString(output_type),
-    //  Cpp::GetTypeAsString(conversion_type),
-    //  conversion_policy_str(policy));
```
jeaye, line 392: This is useful to keep. That's why I have it around.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -477,9 +547,6 @@ namespace jank::codegen
     /* No need to call a function to erase a typed object. Just find the
      * offset to its base member and shift our pointer accordingly. */
     auto const base_offset{ cpp_util::offset_to_typed_object_base(ret_type) };
-    //util::println("convert_object ret_type = {}, needs base adjustment by {} bytes",
-    //              Cpp::GetTypeAsString(ret_type),
-    //              base_offset);
```
jeaye, line 482: Please keep this and all other helpful printlns that existed before your changes.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -525,19 +593,14 @@ namespace jank::codegen
   reusable_context::reusable_context(jtl::immutable_string const &module_name,
                                      std::unique_ptr<llvm::LLVMContext> llvm_ctx)
     : module_name{ module_name }
-    , ctor_name{ unique_munged_string("jank_global_init") }
-    //, llvm_ctx{ std::make_unique<llvm::LLVMContext>() }
-    //, llvm_ctx{ reinterpret_cast<std::unique_ptr<llvm::orc::ThreadSafeContext> *>(
-    //              reinterpret_cast<void *>(
-    //                &static_cast<clang::Interpreter &>(*__rt_ctx->jit_prc.interpreter)))
-    //              ->getContext() }
+    , ctor_name{ __rt_ctx->unique_munged_string() }
```
jeaye: Why lose the `jank_global_init` prefix? Without it, we have no idea what the fn is in IR land.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -525,19 +593,14 @@ namespace jank::codegen
...
     , lam{ std::make_unique<llvm::LoopAnalysisManager>() }
     , fam{ std::make_unique<llvm::FunctionAnalysisManager>() }
     , cgam{ std::make_unique<llvm::CGSCCAnalysisManager>() }
     , mam{ std::make_unique<llvm::ModuleAnalysisManager>() }
     , pic{ std::make_unique<llvm::PassInstrumentationCallbacks>() }
   {
-    auto m{ std::make_unique<llvm::Module>(unique_munged_string(module_name).c_str(), *llvm_ctx) };
+    auto m{ std::make_unique<llvm::Module>(__rt_ctx->unique_munged_string().c_str(), *llvm_ctx) };
```
jeaye: Again, why remove the prefix?

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -698,10 +761,10 @@ namespace jank::codegen
     locals[make_box<obj::symbol>("virtual/this")] = this_arg;
     locals[make_box<obj::symbol>(root_fn->name)] = this_arg;
 
-    for(usize i{}; i < arity.params.size(); ++i)
+    usize i{};
+    for(auto const &param : arity.params)
     {
-      auto &param(arity.params[i]);
-      auto arg(llvm_fn->getArg(i + 1));
+      auto arg(llvm_fn->getArg(++i));
```
jeaye, line 779: This is not a helpful change unless `i` is used outside the loop (and I don't think it is). Otherwise, you've increased the scope of the value, introduced the possibility of more bugs, and took up the same number of lines of code.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -964,9 +1048,58 @@ namespace jank::codegen
       for(auto const &arg_expr : expr->arg_exprs)
       {
         auto arg_handle{ gen(arg_expr, arity) };
-        if(llvm::isa<llvm::AllocaInst>(arg_handle))
+        auto const arg_type{ cpp_util::expression_type(arg_expr) };
+
+        bool should_box{ !cpp_util::is_any_object(arg_type) };
+
+        /* Robustness check: if LLVM type suggests primitive, force boxing */
+        if(!should_box)
```
jeaye, line 1068: We should never be doing implicit `jank_box` calls. If an implicit trait conversion is required, we'll do that, but that's a very different use case.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1187,6 +1320,21 @@ namespace jank::codegen
     {
       llvm::IRBuilder<>::InsertPointGuard const guard{ *ctx->builder };
 
+      for(auto const &capture : expr->captures())
+      {
+        auto const name{ capture.first };
+        llvm::Type *field_type{};
+        if(locals.contains(name) && llvm::isa<llvm::AllocaInst>(locals[name].data))
+        {
+          field_type = llvm::cast<llvm::AllocaInst>(locals[name].data)->getAllocatedType();
```
jeaye, line 1341: You've done three lookups into `locals` here. Just look up once using `find`.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1671,17 +1820,403 @@ namespace jank::codegen
...
+   * 1. Checking if there is an enclosing (parent) exception handler.
+   * 2. If yes, transferring control to that parent's dispatch block so it can attempt to match the exception.
+   * 3. If no, calling `_Unwind_Resume` to continue propagating the exception up the call stack
+   *    (to the caller of the current function).
+   *
+   * Crucially, this maintains the chain of exception handling, ensuring that an exception
+   * is only "swallowed" if it is actually matched by a catch clause. */
+  void llvm_processor::impl::route_unhandled_exception(llvm::Value *ex_ptr,
+                                                       llvm::Value *selector,
+                                                       llvm::BasicBlock *current_block) const
```
jeaye, line 1849: If these can be `const` (i.e. `llvm::Value * const ex_ptr`), please make them so. Same for all parameters.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -1671,17 +1820,403 @@ namespace jank::codegen
...
+  {
+    /* Get parent handler (if exists) */
+    reusable_context::exception_handler_info const *parent_handler{};
```
jeaye, line 1852: We don't need this outside of the `if`. No mutation is needed, if you remove the duplicate check. You check:

1. `!ctx->exception_handlers.empty()`
2. `parent_handler`, but this can only be set if `!ctx->exception_handlers.empty()`

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -3597,7 +4103,33 @@ namespace jank::codegen
...
+      for(auto const &capture : captures)
+      {
+        auto const name{ capture.first };
+        /* If the variable is in locals and it's an alloca, use its actual type.
+         * Otherwise use ptr (for deferred inits, global values, etc.). */
+        llvm::Type *field_type{};
+        if(locals.contains(name) && llvm::isa<llvm::AllocaInst>(locals[name].data))
+        {
+          auto const alloca_type{
+            llvm::cast<llvm::AllocaInst>(locals[name].data)->getAllocatedType()
```
jeaye, line 4173: We have three lookups here again.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -3597,7 +4103,33 @@ namespace jank::codegen
     }
     else
     {
-      std::vector<llvm::Type *> const capture_types{ captures.size(), ctx->builder->getPtrTy() };
+      /* Build context with correct types for each capture. */
+      std::vector<llvm::Type *> capture_types;
+      capture_types.reserve(captures.size());
+      for(auto const &capture : captures)
```
jeaye, line 4164: What led to all of this? Are we now capturing native values instead of just object ptrs?

jeaye (changes requested):
This has come a long way, Jianling, and it's a hugely complex feature. I have reviewed all of the tests and the LLVM IR gen. The IR gen code is definitely coming along, in terms of coherence and cohesion. I don't have design feedback for it right now.

I stopped reviewing the C++ gen about half way, since most of the changes in the diff are from a bad merge; they are undoing changes I made in the past couple of months. I think that you may want to do something like:

```bash
git checkout main -- src/cpp/jank/codegen/processor.cpp
```

And then re-apply your actual C++ gen changes which are specific to exception handling. Once the diff for that file contains only changes you've made, relevant to this PR, I will take another look.

Finally, we need to get CI passing. From a quick glimpse, it looks like some of the failures are things I've already fixed on main. Once you have finished your cleanup, all comments are addressed, and CI is green, I will check this out locally, probably do some cleanup of my own, and then get this merged.

jeaye:
As a separate note, please make sure there are tests for these cases:

* pass - throw + catch a native value (i.e. `(try (throw cpp/std.string.npos) (catch cpp/size_t e e))`)
* pass - return exception value into let (i.e. `(let* [v (try (throw (cpp/std.runtime_error.) (catch cpp/std.runtime_error e e)))`)
* fail - no exception type provided (i.e. `(try (catch e))`)

There might be already. It's possible I missed them.

jianlingzhong:
superseded by #671 , closing this one.

Issue #639, opened 2025-12-31
Rework tagged literals


shantanu-sardesai:
### Description

Currently in jank there is an issue in how tagged literals work, the following code that works in Clojure **_won't_** work in jank.

```clj
user=> (import 'java.net.URL)
java.net.URL

user=> (binding [*data-readers* {'url #(URL. %)}]          ;; Extending tag literals.
         (read-string "#url \"file:/etc/hosts\""))
#object[java.net.URL 0x6dd93a21 "file:/etc/hosts"]

user=> (binding [*data-readers* {'inst #(URL. %)}]         ;; Overwriting existing tag literals.
         (read-string "#inst \"file:/etc/hosts\""))
#object[java.net.URL 0x6b3871d6 "file:/etc/hosts"]
```

The reason it won't work in jank is due to the fact that we've [hard coded](https://github.com/shantanu-sardesai/jank/blob/badd214cae7bf501b1cbf900ff29ebc624dfb8f9/compiler+runtime/src/cpp/jank/read/parse.cpp#L999-L1010) the methods for `#inst` & `#uuid`, while extending the data readers is just not possible since jank doesn't even consult the `*data-readers*` var at the moment.

For example, Clojure consults the `*data-readers*` [here](https://github.com/clojure/clojure/blob/6a4ba6aedc8575768b2fff6d9c9c7e6503a0a93a/src/jvm/clojure/lang/LispReader.java#L1437-L1438) which can also be seen from the examples above. Also unlike jank, Clojure stores the `default-data-readers` in the core, [here](https://github.com/clojure/clojure/blob/6a4ba6aedc8575768b2fff6d9c9c7e6503a0a93a/src/clj/clojure/core.clj#L7988-L7994).

### Goal

- [x] jank should consult the `*data-readers*` for definitions of tagged literals.
- [x] ~~The current parser methods need to be stored in the `default-data-readers` and used inside jank.~~
  - See [this comment](https://github.com/jank-lang/jank/issues/639#issuecomment-3706551429) for explaination.
- [ ] Missing definitions for tagged literals should not be a compile time error but instead should be a runtime error (for ex. `#js {}`), [here](https://github.com/jank-lang/jank/blob/768f8310ce0f3d61b01f2df0f0e66ab9c9df1984/compiler%2Bruntime/src/cpp/jank/read/parse.cpp#L1013-L1016).
[...]

PR #650, opened 2026-01-04
Reworked tagged literals

shantanu-sardesai:
There are some fixes in this PR to the parser besides the changes required for reworking the tagged reader parser.

- Reader suppression support, which goes against the lazy nature of the parser.
- Lexer errors in the eagerly parsed collections (`map`, `set`, etc.) can cause a stack overflow.

---

Resolves https://github.com/jank-lang/jank/issues/639.

### DIFF
--- a/book/src/differences-from-clojure.md
+++ b/book/src/differences-from-clojure.md
@@ -16,7 +16,8 @@ Clojure.
 ## Parser
 * No load operation for `data_readers.(cljc|jank)` at start-up to extend
   supported tags
-* No parser error is thrown in case of unsupported reader conditionals.
+* Only lexer errors and unbalanced forms are rejected in unsupported reader
+  conditionals
 
 ## clojure.core
 * Baked into the `jank` binary, not shipped separately


shantanu-sardesai:
Still thinking on how to make reader suppression work with jank.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -395,107 +412,96 @@ namespace jank::read::parse
...
+            return err(e);
+          }
+
+          key = make_box<obj::opaque_box>(e.data, "jank::error::base const *");
         }
       }
 
-      return jtl::ok();
-    });
+      if(item->is_err())
```
shantanu-sardesai, line 476: One behaviour difference here, from the existing implementation, is that the parser will report all the errors (mainly the duplicate keys error) for the `key` **_before_** any errors for the `value` will be reported.

shantanu-sardesai:
As per our discussion, I'm closing this one in favour of [this](https://github.com/jank-lang/jank/pull/688) implementation.

PR #651, opened 2026-01-06
[docs] Add arglists metadata for the clojure core functions defined using a def (as we don't have defn available yet)

### LINKED ISSUE #643: [core/docs] Add arglists metadata for the clojure core functions defined using a `def` (as we don't have `defn` available yet)
Add arglists metadata for the clojure core functions defined using a `def` (as we don't have `defn` available yet). In Clojure (_Jank's JVM version_ lol), we have it like https://github.com/clojure/clojure/blob/master/src/clj/clojure/core.clj#L126 .

E.g. below
```clojure
(def nnext
  "Same as (next (next x))"
  (fn* nnext [o]
    (next (next o))))
``` 

would become

``` clojure
(def ^{:arglists '([o])
       :doc "Same as (next (next x))"}
  nnext
  (fn* nnext [o]
       (next (next o))))
```

Without it, dev tools won't show the args correctly.

I can open a PR for it if there is nothing impeding us of doing it, it should be straightforward \o

pfeodrippe:
Fixes #643

Just documentation,  also used https://github.com/clojure/clojure/blob/master/src/clj/clojure/core.clj as source for some of the arglists.


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -91,124 +96,150 @@
...
   :data)
 
-(def ex-cause
-  "Returns the cause of ex if ex is a Throwable.
-  Otherwise returns nil."
+(def ^{:arglists '([ex])
+       :doc "Returns the cause of ex if ex is a Throwable.
+  Otherwise returns nil."}
+  ex-cause
   :cause)
```
pfeodrippe, line 115: These are keywords here in jank, but it should be called just as functions as we do in the JVM , so adding here

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -466,27 +530,36 @@
...
+  apply*
   (fn* apply* [source args]
     (cpp/jank.runtime.apply_to source args)))
 
-(def ^{:macro true} let
+(def ^{:arglists '([bindings & body])
+       :macro true}
+  let
   (fn* let [&form &env & decl]
     `(let* ~@decl)))
```
pfeodrippe: Macros like this are redefined later anyway (after the bootstrapping), but it's up to you if we want to maintain this is

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -4406,9 +4484,6 @@
   ;; (. c (isInstance x))
   (throw "TODO: port instance?"))
 
-(def ^{:private true :dynamic true}
-  assert-valid-fdecl (fn [fdecl]))
```
pfeodrippe, line 4410: This is defined later and not really used in-between

compiler+runtime/src/jank/clojure/core.jank
jeaye: Let's not have any meta we don't need to have. Meta is *expensive*.

compiler+runtime/src/jank/clojure/core.jank
jeaye: Yep, nothing wrong with that.

compiler+runtime/src/jank/clojure/core.jank
pfeodrippe: Good o/ So I will remove from private functions as well, thanks for the pointer o/

pfeodrippe: And it’s expensive in terms of what? Just to understand it better, not perf, I hope ahahaha if it’s perf or if it increases the used memory by a lot, then maybe we could have this info in the dev tool that needs it instead (be it in the linter or REPL)

jeaye: Compile-time, startup time, and memory usage will all be impacted by more meta, in that order.

pfeodrippe: Done \o Thanks

jeaye:
Thanks, Paulo!

PR #658, opened 2026-01-10
Wire in ratio support to /

### LINKED ISSUE #629: Wire in ratio support to `/`
Currently the `/` operator in the jank compiler is not hooked into the runtime ratio capability due to which following cases in jank don't behave as would be expected of a Clojure dialect:

```jank
user=> (/ 5 2)
2    ;; Expected a ratio (5/2) instead of a rounded integer.
```

djblue:
Fixes #629

### DIFF
--- a/compiler+runtime/include/cpp/jank/analyze/cpp_util.hpp
+++ b/compiler+runtime/include/cpp/jank/analyze/cpp_util.hpp
@@ -53,6 +53,7 @@ namespace jank::analyze::cpp_util
   bool is_non_static_member_function(jtl::ptr<void> scope);
   bool is_nullptr(jtl::ptr<void> type);
   bool is_implicitly_convertible(jtl::ptr<void> from, jtl::ptr<void> to);
+  bool is_pointer_to_void_conversion(jtl::ptr<void> from, jtl::ptr<void> to);
 
   jtl::ptr<void> untyped_object_ptr_type();
   jtl::ptr<void> untyped_object_ref_type();
--- a/compiler+runtime/include/cpp/jank/runtime/convert/builtin.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/convert/builtin.hpp
@@ -238,10 +238,11 @@ namespace jank::runtime
 
   /* C strings. */
   template <typename T>
-  requires(jtl::is_any_same<std::decay_t<T>, char *, char const *>)
+  requires(
+    jtl::is_any_same<std::decay_t<T>, char *, char const *, signed char *, signed char const *>)
   struct convert<T>
   {
-    static obj::persistent_string_ref into_object(char const * const o)
+    static obj::persistent_string_ref into_object(T const o)
     {
       if(o == nullptr)
       {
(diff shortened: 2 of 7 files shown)


compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -376,11 +376,42 @@ namespace jank::runtime
...
+              else if constexpr(std::is_same_v<LeftType, native_big_integer>
+                                && std::is_same_v<RightType, i64>)
+              {
+                return obj::ratio::create(typed_l_data, native_big_integer(typed_r->data));
+              }
+              else
+              {
+                return obj::ratio::create(typed_l_data, typed_r->data);
+              }
+            }
```
jeaye, line 400: These 16 lines can just be the following, I think:

```cpp
return obj::ratio::create(native_big_integer(typed_l_data), native_big_integer(typed_r->data));
```

If we need to copy-construct a `native_big_integer` because we already have one, that shouldn't be a big problem. I'd expect it to be elided by Clang.

jeaye (changes requested):
Thanks for the PR, Chris!

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
jeaye: Or maybe this entire function can just be that?

djblue: I think because `div` needs to support more than `i64` and `native_big_integer`, we probably still need to check for int like:

```clojure
% clj                                          
Clojure 1.12.2
user=> (/ 10 1.0)
10.0
```

jeaye: Oh, ok. Let's just do this then:

```cpp
return obj::ratio::create(typed_l_data, typed_r->data);
```

Then add the overloads needed for `ratio::create`, to handle the various cases. You're basically implementing your own overload resolution here, with `if constexpr`, and I want to make sure we don't need to do that anywhere else. Might as well let Clang do it for us.

jeaye: Following up on this. Looks like you made the first change, but I want to make sure that all of the cases you're considering (like int / float) are all handled.

djblue: @jeaye I think since ratio only works with `i64` or `native_big_integer`, it would be a bit odd to add methods for constructing ratios for other number types. I could see dispatching to the other overloads of `div`, but the number of combinations of left and right types could get lengthy and they would all eventually land at the `/` cpp operator overloads anyway.

jeaye: Oh, that's a good point. So are you feeling good with this change as it is? Ready to merge?

djblue: I think the PR is good to merge 👍

jeaye:
Thanks, Chris!

Issue #422, opened 2025-08-27
Lexing (non hex) Unicode Characters Produces Errors


E-A-Griffin:
### jank health check

─ system ───────────────────────────────────────────────────────────────────────────────────────────
─ ✅ operating system: macos
─ ✅ default triple: arm64-apple-darwin24.6.0

─ jank install ─────────────────────────────────────────────────────────────────────────────────────
─ ✅ jank version: jank-0.1-a228bd5a2eaae21f008730b8bc12628c1d4ba062
─ ✅ jank assertions are enabled; performance will be impacted
─ ✅ jank resource dir: ../lib/jank/0.1  (not found)
─ ✅ jank resolved resource dir: /Users/emma/programming/jank/compiler+runtime/build/../lib/jank/0.1  (ignored for dev build)
─ ✅ jank user cache dir: /Users/emma/.cache/jank/arm64-apple-darwin24.6.0-8114833cac20af8f4006aac6e0400331d68409d58412ed4ac5586b8968d7a042  (not found)
─ ✅ jank pch path: /Users/emma/.cache/jank/arm64-apple-darwin24.6.0-8114833cac20af8f4006aac6e0400331d68409d58412ed4ac5586b8968d7a042 (found)

─ clang install ────────────────────────────────────────────────────────────────────────────────────
─ ✅ configured clang path: /Users/emma/programming/jank/compiler+runtime/build/llvm-install/usr/local/bin/clang++ (found)
─ ✅ configured clang resource dir: /Users/emma/programming/jank/compiler+runtime/build/llvm-install/usr/local/lib/clang/21 (found)

─ jank runtime ─────────────────────────────────────────────────────────────────────────────────────
─ ✅ jank runtime initialized
─ ✅ jank can jit compile c++
─ ✅ jank can jit compile llvm ir

─ support ──────────────────────────────────────────────────────────────────────────────────────────
If you're having issues with jank, please either ask the jank community on the Clojurians Slack or report the issue on Github.

─ Slack: https://clojurians.slack.com/archives/C03SRH97FDK
─ Github: https://github.com/jank-lang/jank

### Description of the issue with reproduction steps

While `jank` can properly lex (parse, perform semantic analysis, etc.) hex unicode characters (e.g. `\u00a0`) it returns an error when trying to lex unicode text directly (e.g. `\λ`)

```clojure
user=> \λ
─ lex/invalid-unicode ──────────────────────────────────────────────────────────────────────────────
error: Unfinished character.

[...]

PR #660, opened 2026-01-14
bugfix - Lex/parse (non-hex, non-octal) unicode characters

E-A-Griffin:
This closes https://github.com/jank-lang/jank/issues/422

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -28,41 +28,32 @@ namespace jank::read::parse
     }
     else if(s.size() == 2)
     {
-      auto const is_char0_utf8 {
-        static_cast<unsigned char>(s[0]) >= 0xC0 && static_cast<unsigned char>(s[0]) <= 0xDF
-      };
-      auto const is_char1_utf8{
-        static_cast<unsigned char>(s[1]) >= 0x80 && static_cast<unsigned char>(s[1]) <= 0xBF
-      };
+      auto const is_char0_utf8{ static_cast<unsigned char>(s[0]) >= 0xC0
+                                && static_cast<unsigned char>(s[0]) <= 0xDF };
+      auto const is_char1_utf8{ static_cast<unsigned char>(s[1]) >= 0x80
+                                && static_cast<unsigned char>(s[1]) <= 0xBF };
       return is_char0_utf8 && is_char1_utf8;
     }
     else if(s.size() == 3)
     {
-      auto const is_char0_utf8{
-        static_cast<unsigned char>(s[0]) >= 0xE0 && static_cast<unsigned char>(s[0]) <= 0xEF
-      };
-      auto const is_char1_utf8{
-        static_cast<unsigned char>(s[1]) >= 0x80 && static_cast<unsigned char>(s[1]) <= 0xBF
-      };
-      auto const is_char2_utf8{
-        static_cast<unsigned char>(s[2]) >= 0x80 && static_cast<unsigned char>(s[2]) <= 0xBF
-      };
+      auto const is_char0_utf8{ static_cast<unsigned char>(s[0]) >= 0xE0
+                                && static_cast<unsigned char>(s[0]) <= 0xEF };
+      auto const is_char1_utf8{ static_cast<unsigned char>(s[1]) >= 0x80
+                                && static_cast<unsigned char>(s[1]) <= 0xBF };
+      auto const is_char2_utf8{ static_cast<unsigned char>(s[2]) >= 0x80
+                                && static_cast<unsigned char>(s[2]) <= 0xBF };
       return is_char0_utf8 && is_char1_utf8 && is_char2_utf8;
     }
     else if(s.size() == 4)
     {
-      auto const is_char0_utf8{
-        static_cast<unsigned char>(s[0]) >= 0xF0 && static_cast<unsigned char>(s[0]) <= 0xF7
-      };
-      auto const is_char1_utf8{
-        static_cast<unsigned char>(s[1]) >= 0x80 && static_cast<unsigned char>(s[1]) <= 0xBF
-      };
-      auto const is_char2_utf8{
-        static_cast<unsigned char>(s[2]) >= 0x80 && static_cast<unsigned char>(s[2]) <= 0xBF
-      };
-      auto const is_char3_utf8{
-        static_cast<unsigned char>(s[3]) >= 0x80 && static_cast<unsigned char>(s[3]) <= 0xBF
-      };
+      auto const is_char0_utf8{ static_cast<unsigned char>(s[0]) >= 0xF0
+                                && static_cast<unsigned char>(s[0]) <= 0xF7 };
+      auto const is_char1_utf8{ static_cast<unsigned char>(s[1]) >= 0x80
+                                && static_cast<unsigned char>(s[1]) <= 0xBF };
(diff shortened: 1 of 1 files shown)


E-A-Griffin:
@jeaye 
Screenshot 2026-01-13 at 9 55 53 PM
I'm not seeing a way to request your review like I normally do through GitHub, let me know if there's another way I can besides tagging you like this

E-A-Griffin:
An interesting note from doing this work, `jank` supports 4 byte unicode characters while it appears that Clojure doesn't. If we'd like to match Clojure and not support 4 byte unicode characters, I can make that change. Clojure is odd though in that you can use 4 byte unicode symbols
```clj
Clojure 1.12.3
(def 𐅦 nil)
#'user/𐅦
user=> \𐅦
Syntax error reading source at (REPL:4:0).
Unsupported character: \𐅦
user=> 𐅦
nil
```

```clj
./build/jank repl
user=> \𐅦
\𐅦

user=> (def 𐅦 nil)
#'user/𐅦

user=> 𐅦
nil
```
https://graphemica.com/%F0%90%85%A6

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -592,13 +592,14 @@ namespace jank::read::lex
           require_space = false;
 
           auto const ch(peek());
-          ++pos;
 
           if(ch.is_err())
           {
+            pos++;
```
jeaye: Please use `++pos;` here. In C++, since `++` can be overloaded (and often is), the performance difference between pre and post increment can be significant. Furthermore, post-increment is a red flag, due to it being logically complex (this value isn't updated now, but it will be later). Whenever that shows up, it always begs the question of whether it can be simplified.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -20,6 +20,46 @@ namespace jank::read::parse
 {
   using namespace jank::runtime;
 
+  static bool is_utf8_char(jtl::immutable_string_view s)
```
jeaye: String views are dangerous, in C++. They can implicitly be created by C strings, lvalue strings, and also rvalue strings. The latter case is undefined behavior. For this function, there is no need for a view. We can just take a `jtl::immutable_string const &` and no copy will be done from our call. The rest of the code can remain the same.

As a side note, always make your parameters `const` if they can be.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -20,6 +20,46 @@ namespace jank::read::parse
...
+  {
+    if(s.size() == 1)
+    {
+      return static_cast<unsigned char>(s[0]) <= 0x7F;
+    }
+    else if(s.size() == 2)
+    {
+      bool const is_char0_utf8_code0
```
jeaye: Please always use direct initialization. Also, there's no need for spelling out types here.

```cpp
auto const is_char0_utf8_code0{ static_cast<unsigned char>(s[0]) >= 0xC0 && static_cast<unsigned char>(s[0]) <= 0xDF };
```

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -512,6 +552,18 @@ namespace jank::read::parse
                                    start_token,
                                    start_token };
       }
+      else if(sv[0] == '\\' && sv.size() >= 3 && sv.size() <= 5)
```
jeaye: Let's move these magic numbers into `static constexpr auto const` (say that 5x fast) values, with helpful names, so it's clear to the future reader a year from now.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -20,6 +20,46 @@ namespace jank::read::parse
...
+    {
+      bool const is_char0_utf8_code0
+        = static_cast<unsigned char>(s[0]) >= 0xF0 && static_cast<unsigned char>(s[0]) <= 0xF7;
+      bool const is_char1_utf8_code1
+        = static_cast<unsigned char>(s[1]) >= 0x80 && static_cast<unsigned char>(s[1]) <= 0xBF;
+      bool const is_char2_utf8_code2
+        = static_cast<unsigned char>(s[2]) >= 0x80 && static_cast<unsigned char>(s[2]) <= 0xBF;
+      bool const is_char3_utf8_code3
+        = static_cast<unsigned char>(s[3]) >= 0x80 && static_cast<unsigned char>(s[3]) <= 0xBF;
+      return is_char0_utf8_code0 && is_char1_utf8_code1 && is_char2_utf8_code2
```
jeaye: I appreciate the descriptive names. Perhaps we can shorten this by using `is_char0_utf8`, without the `_code0` suffix, since you're always repeating the number with `charN` and `codeN`.

jeaye:
> An interesting note from doing this work, `jank` supports 4 byte unicode characters while it appears that Clojure doesn't. If we'd like to match Clojure and not support 4 byte unicode characters, I can make that change. Clojure is odd though in that you can use 4 byte unicode symbols

Great catch. I would prefer to keep supporting 4 byte characters. This level of Clojure parity is not required to "be Clojure".

E-A-Griffin:
> This looks like a Unicode rendering issue with the reporter, which might be something you're interested in tackling as it's related.

@jeaye I think I saw this fixed during intermediate testing but now that it properly parses unicode I'm not sure how to reproduce it, any ideas?

jeaye:
> > This looks like a Unicode rendering issue with the reporter, which might be something you're interested in tackling as it's related.
> 
> @jeaye I think I saw this fixed during intermediate testing but now that it properly parses unicode I'm not sure how to reproduce it, any ideas?

The test reporter uses the lexer, so it could be that your changes fixed the issue! If your test from your original ticket works now, I'd say we're good.

For example, with this source:

```cpp
(let [c \λ
      d {c}])
```

You should see this error:

```
❯ jank run test.jank
─ parse/odd-entries-in-map ─────────────────────────────────────────────────────────────────────────
error: Odd number of entries in map.

     │ test.jank
  1  │ (let [c \λ
  2  │       d {c}])
     │          ^ No value for this key.
```

E-A-Griffin:
@jeaye 
```clj
user=> (let [c \λ d {c}])
─ parse/odd-entries-in-map ─────────────────────────────────────────────────────
error: Odd number of entries in map.                                            
                                                                                
─────┬──────────────────────────────────────────────────────────────────────────
     │ /var/folders/z1/z53r7vq50vj2852z43cdzsw40000gn/T/jank-repl-hbYE42        
─────┼──────────────────────────────────────────────────────────────────────────
  1  │ (let [c \λ d {c}])                                                       
     │                ^ No value for this key.                                  
─────┴──────────────────────────────────────────────────────────────────────────

```
looks good to me!

E-A-Griffin:
```clj
user=> (let [c \λ] \
user=>...    d {c} \
user=>... )
─ parse/odd-entries-in-map ─────────────────────────────────────────────────────
error: Odd number of entries in map.                                            
                                                                                
─────┬──────────────────────────────────────────────────────────────────────────
     │ /var/folders/z1/z53r7vq50vj2852z43cdzsw40000gn/T/jank-repl-hbYE42        
─────┼──────────────────────────────────────────────────────────────────────────
  1  │ (let [c \λ]                                                              
  2  │ d {c}                                                                    
     │    ^ No value for this key.                                              
─────┴──────────────────────────────────────────────────────────────────────────
```
with multi-lines just to be sure

jeaye:
Thank you, Emma!

PR #665, opened 2026-01-16
fix: Add zlib as Debian Dependency

stmonty:
Trying to compile a project on a fresh Ubuntu install with Leiningen failed to link.
This fixes it by adding zlib as a dependency.

```
  /usr/bin/ld: cannot find -lz: No such file or directory
─ system/failure ───────────────────────────────────────────────────────────────────────────────────
error: Clang failed with errors.error:  linker command failed with exit code 1 (use -v to see       
       invocation)
```


jeaye:
Thanks!

PR #671, opened 2026-01-20
Implement support for multiple typed catch clauses

### LINKED ISSUE #558: Support catching specific exception types

jianlingzhong:
This PR was written with assistance from Google Gemini models. 

This closes #558.

# Tests included for `try ... catch ... finally`. 

## Basic Functionality
Core mechanics of exception handling and return values.

*   **[pass-try-basic.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-basic.jank)**
    *   **Purpose**: Tests that `try` blocks return the last expression in the body (success case) or the catch block (error case). Also verifies `try` works as an expression in other forms.
    *   **Uniqueness**: Fundamental baseline test.
*   **[pass-try-empty-bodies.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-empty-bodies.jank)**
    *   **Purpose**: Tests edge cases with empty `try`, `catch`, and `finally` bodies to ensure the parser handles them gracefully (usually returning `nil`).
    *   **Uniqueness**: Validates parser resilience against valid but degenerate code.
*   **[pass-unboxed-position.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-unboxed-position.jank)**
    *   **Purpose**: Verifies that `try` expressions correctly handle unboxed return types (e.g., returning a raw primitive number) when used in a context expecting a value.
    *   **Uniqueness**: Critical for the compiler's type analysis and boxing/unboxing logic.

## Control Flow & Execution Order
Verifying that code executes in the strict sequence defined by the language.

*   **[pass-try-catch-finally-execution.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-catch-finally-execution.jank)**
    *   **Purpose**: asserts the strict `try -> catch -> finally` execution order using side effects (vector accumulation).
    *   **Uniqueness**: The definitive integration test for the full lifecycle of a handled exception.
*   **[pass-catch-ordering.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-catch-ordering.jank)**
    *   **Purpose**: Tests that the *first* matching `catch` clause is selected, not the "best" match. Crucial for C++ inheritance (e.g., catching `Base` before `Derived`).
    *   **Uniqueness**: Verifies linear scan semantics of catch clauses.
*   **[pass-finally-behavior.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-finally-behavior.jank)**
    *   **Purpose**: Verifies `finally` blocks execute for side effects but do *not* affect the return value of the `try` expression (unless they throw).
    *   **Uniqueness**: Distinguishes `finally` (cleanup) from `catch` (recovery).
*   **[pass-throw-from-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-throw-from-finally.jank)**
    *   **Purpose**: Tests that throwing an exception from a `finally` block overrides any in-flight exception or return value.
    *   **Uniqueness**: Verifies "abrupt completion" semantics of `finally`.

## C++ Interop & Types
Coverage for catching native C++ types.

*   **[pass-catch-cpp-primitives.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-primitives.jank)**
    *   **Purpose**: Catching fundamental C++ types: `char`, `double`, `float`, `int`, `unsigned int`.
    *   **Uniqueness**: Ensures the runtime and usages of `throw/catch` work for non-object primitives.
*   **[pass-catch-cpp-string-literal.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-string-literal.jank)**
    *   **Purpose**: Catching `const char*` (string literals).
    *   **Uniqueness**: Verifies pointer matching specifically for C-strings.
*   **[pass-catch-cpp-pointer.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-pointer.jank)**
    *   **Purpose**: Catching `void*` and strictly defined pointer types.
    *   **Uniqueness**: Ensures pointer identity is preserved across the catch boundary.
*   **[pass-catch-cpp-std-exceptions.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-std-exceptions.jank)**
    *   **Purpose**: Catching standard library exceptions (`std::runtime_error`, `std::bad_cast`) and validating polymorphic catching (catching via base `std::exception`).
    *   **Uniqueness**: Verifies RTTI-based hierarchy matching for C++ objects.
*   **[pass-catch-cpp-no-slicing.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-cpp-no-slicing.jank)**
    *   **Purpose**: Catches a derived object by a base class value/reference to ensure the underlying object is not "sliced" or corrupted.
    *   **Uniqueness**: Critical for C++ object fidelity.

## Complex Semantics & Nesting
Advanced structural tests.

*   **[pass-nested-try-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-nested-try-catch.jank)**
    *   **Purpose**: Standard nesting: `try` within `try`. Verifies exceptions bubble up to the nearest enclosing matching catch.
    *   **Uniqueness**: Verifies scope definition and exception propagation.
*   **[pass-nested-no-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-nested-no-catch.jank)**
    *   **Purpose**: Nested `try` blocks where intermediate levels have *only* `finally` blocks and no `catch`.
    *   **Uniqueness**: Stress-tests stack unwinding through layers that perform cleanup but do not stop the exception.
*   **[pass-nested-fn-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-nested-fn-finally.jank)**
    *   **Purpose**: Exceptions propagating across function boundaries, triggering `finally` blocks in both the callee and caller.
    *   **Uniqueness**: Verifies that unwinding correctly traverses the call stack (not just lexical scope).
*   **[pass-try-in-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-in-catch.jank)**
    *   **Purpose**: A full `try` block nested *inside* a `catch` handler.
    *   **Uniqueness**: Validates that the catch block itself is a valid code context for new exception scopes (recovery logic).
*   **[pass-try-in-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-try-in-finally.jank)**
    *   **Purpose**: A full `try` block nested *inside* a `finally` handler.
    *   **Uniqueness**: Validates complex cleanup logic that might itself need error handling.

## Mixed Environment
Interaction between Jank and C++ exceptions.

*   **[pass-catch-mixed.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-catch-mixed.jank)**
    *   **Purpose**: A single `try` block with both `catch cpp/...` and `catch jank.runtime.object_ref`.
    *   **Uniqueness**: Verifies that the runtime can dispatch to either backend logic from a single site.
*   **[pass-catch-rethrow-alloc.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-catch-rethrow-alloc.jank)**
    *   **Purpose**: Catching an exception object and immediately re-throwing *that same object*.
    *   **Uniqueness**: Verifies object identity and lifetime during re-throw (distinct from `throw;` which re-throws the *active* exception).
*   **[pass-rethrow.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-rethrow.jank)**
    *   **Purpose**: Tests `throw` with no arguments (C++ `throw;`) to re-raise the currently active exception.
    *   **Uniqueness**: Verifies access to thread-local exception state.
*   **[pass-uncaught-cpp-exception-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-uncaught-cpp-exception-finally.jank)**
    *   **Purpose**: Verifies `finally` executes even when a C++ exception is *not* caught by any local catch clause.
    *   **Uniqueness**: Ensures cleanup reliability for unhandled native exceptions.
*   **[pass-destructor-unwinding.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/cpp/try/pass-destructor-unwinding.jank)**
    *   **Purpose**: Confirms that C++ destructors for local stack objects are invoked when a Jank `try/catch` unwinds the stack.
    *   **Uniqueness**: Mandatory for RAII correctness/memory safety.

## Recursion & Scoping
Interaction with `recur` and closures.

*   **[pass-capture-within-let.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-capture-within-let.jank)**
    *   **Purpose**: `try` nested inside `let`, verifying it can see outer bindings.
    *   **Uniqueness**: Basic lexical scoping check.
*   **[pass-closure-capture.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-closure-capture.jank)**
    *   **Purpose**: Catch blocks capturing outer function parameters and closure variables.
    *   **Uniqueness**: Essential for functional programming patterns where handlers rely on context.
*   **[pass-recur-local-to-try.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-recur-local-to-try.jank)**
    *   **Purpose**: Using `recur` to jump to a loop head *inside* the `try` block.
    *   **Uniqueness**: Verifies `recur` doesn't get confused by the try protection boundary.
*   **[pass-recur-local-to-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-recur-local-to-catch.jank)**
    *   **Purpose**: Using `recur` to jump to a loop head *inside* the `catch` block.
    *   **Uniqueness**: Verifies `recur` works in exception handler scope.
*   **[pass-recur-local-to-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/pass-recur-local-to-finally.jank)**
    *   **Purpose**: Using `recur` to jump to a loop head *inside* the `finally` block.
    *   **Uniqueness**: Verifies `recur` works in cleanup scope.

## Compile-Time Validation (Negative Tests)
Ensuring invalid code is rejected with meaningful errors.

*   **[fail-catch-after-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-catch-after-finally.jank)**: Enforces strict [(catch ...)* (finally ...)?](file:///Users/jianlingzhong/personal/jank/compiler+runtime/dump_units.py#53-90) ordering.
*   **[fail-duplicate-catch-type.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-duplicate-catch-type.jank)**: Enforces that you can't have two `catch` clauses for the exact same native type.
*   **[fail-multiple-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-multiple-finally.jank)**: Enforces max one `finally` block.
*   **[fail-other-form-after-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-other-form-after-finally.jank)**: Enforces `finally` is the last element.
*   **[fail-recur-in-catch.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-recur-in-catch.jank)**: (Compiler Constraint) Currently validates restrictions on recurring *out* of a catch block if unsupported, or semantics thereof.
*   **[fail-recur-in-finally.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-recur-in-finally.jank)**: (Compiler Constraint) Validates restrictions on recurring *out* of a finally block.
*   **[fail-recur-in-try.jank](file:///Users/jianlingzhong/personal/jank/compiler+runtime/test/jank/form/try/fail-recur-in-try.jank)**: (Compiler Constraint) Validates restrictions on recurring *across* the try boundary (e.g., from try to outside loop).

### DIFF
--- a/compiler+runtime/src/cpp/jank/util/scope_exit.cpp
+++ b/compiler+runtime/src/cpp/jank/util/scope_exit.cpp
@@ -7,7 +7,6 @@ namespace jank::util
 {
   scope_exit::scope_exit(scope_exit::function_type &&f)
     : func{ std::move(f) }
-    , should_propagate_exceptions{ false }
   {
   }
 


compiler+runtime/include/cpp/jank/util/scope_exit.hpp
```diff
@@ -9,8 +9,16 @@ namespace jank::util
...
+    scope_exit(function_type &&f, bool should_propagate_exceptions);
+    ~scope_exit() noexcept(false);
+
+    void release()
+    {
+      active = false;
+    }
 
     function_type func;
+    bool should_propagate_exceptions;
```
jeaye: Always initialize primitive values. This will be undefined, otherwise.

compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
```diff
@@ -76,6 +72,17 @@ namespace jank::analyze::cpp_util
 
   jtl::ptr<void> resolve_type(jtl::immutable_string const &sym, u8 const ptr_count)
   {
+    /* Clang canonicalizes "char" to "signed char" on some platforms, which breaks exception
+     * handling since they are distinct types. We use resolve_literal_type to get the
+     * exact type for "char". */
+    if(sym == "char")
+    {
+      if(auto const res{ resolve_literal_type("char").expect_ok() })
```
jeaye: This is a very costly operation, since `resolve_literal_type` requires JIT compiling C++ code. We should put this into a static variable so we only do this once and then the resulting type is cached.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2736,69 +2740,129 @@ namespace jank::analyze
...
+                       "The binding symbol in 'catch' must be unqualified.",
+                       object_source(item),
+                       latest_expansion(macro_expansions))
+                ->add_usage(read::parse::reparse_nth(item, 2));
+            }
+            auto const catch_type_ref(static_ref_cast<expr::cpp_type>(catch_type.expect_ok()));
+            /* If we're catching a C++ class/struct by value, we want to promote it to a reference
+             * to avoid object slicing and to enable polymorphism.
+             * However, we must NOT promote types in the jank::runtime namespace (like object_ref),
+             * as these are smart pointers expected to be passed by value in the runtime. */
```
jeaye, line 2813: Why not just catch all non-pointer types by reference? This won't matter in C++ at all. Does it complicate the IR?

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1309,7 +1309,14 @@ namespace jank::codegen
...
       case analyze::expression_position::tail:
-        util::format_to(body_buffer, "return {};", ret);
+        if(expr->needs_box)
+        {
+          util::format_to(body_buffer, "return jank::runtime::make_box({});", ret);
+        }
+        else
+        {
+          util::format_to(body_buffer, "return {};", ret);
+        }
```
jeaye: We do not want any boxing changes for C++ codegen.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1762,65 +1769,57 @@ namespace jank::codegen
...
 
     util::format_to(body_buffer, "{");
     if(expr->finally_body.is_some())
     {
-      util::format_to(body_buffer, "jank::util::scope_exit const finally{ [&](){ ");
+      util::format_to(body_buffer, "auto const finally{ [&](){ ");
       gen(expr->finally_body.unwrap(), fn_arity);
       util::format_to(body_buffer, "} };");
+      util::format_to(body_buffer, R"(
+      jank::util::scope_exit guard{ finally, true };
```
jeaye: Let's use generated names for both `finally` and `guard`, so they cannot conflict with the generated code.

```cpp
auto const finally_tmp{ runtime::munge(__rt_ctx->unique_string("finally")) };
```

Same for `finally_guard`.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1762,65 +1769,57 @@ namespace jank::codegen
   jtl::option<handle>
   processor::gen(analyze::expr::try_ref const expr, analyze::expr::function_arity const &fn_arity)
   {
-    auto const has_catch{ expr->catch_body.is_some() };
-    auto ret_tmp(runtime::munge(__rt_ctx->unique_string("try")));
+    auto const has_catch{ !expr->catch_bodies.empty() };
+    auto ret_tmp(runtime::munge(__rt_ctx->unique_namespaced_string("try")));
```
jeaye: There is no need for namespacing here. Do not use `unique_namespaced_string` for C++ gen.

jianlingzhong:
Current failing tests in LLVM IR gen:

```
# test/jank/cpp/auto-box/pass-closure.jank
# test/jank/cpp/auto-box/pass-reference-input.jank
# test/jank/cpp/constructor/complex/pass-aggregate-object.jank
# test/jank/cpp/global-value/pass-static-non-copyable.jank
# test/jank/cpp/global-value/pass-static.jank
# test/jank/cpp/literal-value/pass-string-literal.jank
# test/jank/cpp/loop/pass-native-binding-custom-type.jank 
# test/jank/cpp/loop/pass-native-binding-trait-convertible.jank 
# test/jank/cpp/member/pass-static.jank 
# test/jank/cpp/operator/aget/pass-binary-compatible-types.jank 
# test/jank/cpp/operator/aget/pass-binary-return-reference.jank 
# test/jank/cpp/operator/minusminus/pass-unary-pointer.jank 
# test/jank/cpp/operator/plusplus/pass-unary-pointer.jank 
# test/jank/cpp/preprocessor/pass-normal.jank 
# test/jank/cpp/preprocessor/pass-unqualified.jank 
```

all of them, plus one more (`# test/jank/cpp/closure/pass-convertible.jank`) are also failing in `upstream/main`.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
jianlingzhong: updated. only check for non-pointer type now.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
jianlingzhong: moved fix to analyze code.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2901,6 +2900,52 @@ namespace jank::analyze
...
+           && !cpp_util::is_trait_convertible(last_expression_type))
+        {
+          /* TODO: Error. */
+          return error::analyze_invalid_cpp_conversion(
+            util::format("Body is returning a native object of type '{}', which is not "
+                         "convertible to a jank runtime object.",
+                         Cpp::GetTypeAsString(last_expression_type)),
+            object_source(list),
+            latest_expansion(macro_expansions));
+        }
```
jeaye: There's no need for this check. Just try the implicit conversion, with the expected type. It'll fail if it's not convertible.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -2681,25 +2681,24 @@ namespace jank::analyze
...
 
           if constexpr(std::same_as<T, obj::nil>)
           {
             return try_expression_type::other;
           }
           else
           {
-            auto const first(runtime::first(typed_item->seq()));
-            if(runtime::equal(first, catch_))
+            if(auto const first(runtime::first(typed_item->seq())); runtime::equal(first, catch_))
```
jeaye: I'm not a fan of this pattern. `first` is defined within the `if`, but is also used within the `else`. This is not clear and will be surprising to many.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
jeaye: This still needs to be addressed.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
jeaye: This still needs to be addressed.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
jianlingzhong: removed in my latest change.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
jianlingzhong: reverted.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1762,18 +1762,21 @@ namespace jank::codegen
   processor::gen(analyze::expr::try_ref const expr, analyze::expr::function_arity const &fn_arity)
   {
     auto const has_catch{ !expr->catch_bodies.empty() };
-    auto ret_tmp(runtime::munge(__rt_ctx->unique_namespaced_string("try")));
+    auto ret_tmp(runtime::munge(__rt_ctx->unique_string("try")));
     util::format_to(body_buffer, "jank::runtime::object_ref {}{ };", ret_tmp);
 
     util::format_to(body_buffer, "{");
+    auto const finally_name(runtime::munge(__rt_ctx->unique_string("finally")).data());
+    auto const finally_guard_name(runtime::munge(__rt_ctx->unique_string("finally_guard")).data());
```
jeaye: This pattern is asking for UB in C++. `.data()` on a string will give you the internal representation, but it won't keep the string alive. Why did you add it here? There should be no reason for it.

jianlingzhong: hmm.... thought i had a compile error without it but turns out i was wrong. removed it.

jeaye:
Ok, CI's back up. Please make sure you compile locally with `-Djank_analyze=on` in order to catch the clang-tidy issues.

jianlingzhong:
> Ok, CI's back up. Please make sure you compile locally with `-Djank_analyze=on` in order to catch the clang-tidy issues.

should be fixed now.

Issue #647, opened 2026-01-03
Error reporter searches for a source file even with `read-string`


shantanu-sardesai:
There is a small bug in jank's error code snippet reporter wherein it tries to search for a source file even when we are using `read-string` to read in some arbitrary source code from a string.

```jank
user=> (read-string "#cpp")
─ parse/invalid-reader-tag-value ───────────────────────────────────────────────────────────────────
error: The string literal after this '#cpp' is missing.

─────┬──────────────────────────────────────────────────────────────────────────────────────────────
     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-5Wf2yg
─────┼──────────────────────────────────────────────────────────────────────────────────────────────
  1  │ (read-string "#cpp")
     │ ^^^^^^^^^^^^^^^^^^^^ Used here.
─────┴──────────────────────────────────────────────────────────────────────────────────────────────
╭ nil ─────────────────────────────────────────────────────────────────────────────────────────────╮
│Unable to map file: Unable to find module 'user'.                                                 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

PR #673, opened 2026-01-22
Implement `read-string` & `read-file`

shantanu-sardesai:
Implement `read` and it's siblings along with reader options.

### DIFF
--- a/book/src/differences-from-clojure.md
+++ b/book/src/differences-from-clojure.md
@@ -18,7 +18,7 @@ Clojure.
   supported tags
 * jank allows using reader conditionals in `.jank` files as well
 * No automatic method to disable read by setting the `clojure.core/*read-eval*`
-  at start-up
+  at start-up. The value can be overriden via a `binding` operation
 
 ## clojure.core
 * Baked into the `jank` binary, not shipped separately
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -182,6 +182,24 @@ namespace jank::read::parse
   {
   }
 
+  processor::processor(lex::processor::iterator const &b,
+                       lex::processor::iterator const &e,
+                       object_ref const &extended_features,
+                       bool const &allow_reader_conditional,
+                       bool const &in_preservation_mode)
+    : token_current{ b }
+    , token_end{ e }
+    , splicing_allowed_var{ make_box<var>(
+                              __rt_ctx->intern_ns(make_box<obj::symbol>("clojure.core")),
+                              make_box<obj::symbol>("*splicing-allowed?*"),
+                              jank_false)
+                              ->set_dynamic(true) }
+    , extended_features{ extended_features }
+    , allow_reader_conditional{ allow_reader_conditional }
+    , in_preservation_mode{ in_preservation_mode }
+  {
+  }
+
   processor::object_result processor::next()
   {
     if(token_current == token_end)
@@ -1121,7 +1139,7 @@ namespace jank::read::parse
       return error::parse_invalid_reader_conditional(
         { start_token.start, latest_token.end },
         "Conditional read is not allowed by default. Set the :read-cond reader option to either "
-        ":preserve or :allow. Found a nil instead.");
+        ":preserve or :allow.");
     }
 
     if(token_current->is_err())
@@ -1170,8 +1188,7 @@ namespace jank::read::parse
       /* We take the first match, checking for :jank first. If there are duplicates, it doesn't
        * matter. If :default comes first, we'll always take it. In short, order is important. This
        * matches Clojure's behavior. */
-      auto const is_supported_feature{ (equal(feature_kw, jank_keyword)
-                                        || equal(feature_kw, default_keyword)
+      auto const is_supported_feature{ (feature_kw == jank_keyword || feature_kw == default_keyword
                                         || contains(extended_features, feature_kw))
                                        && result.is_none() };
 
--- a/compiler+runtime/src/cpp/jank/runtime/context.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/context.cpp
@@ -307,86 +307,98 @@ namespace jank::runtime
   {
     profile::timer const timer{ "rt read_string" };
     static auto const unknown_kw{ __rt_ctx->intern_keyword("unknown").expect_ok() };
-    static auto const read_eval_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
-    auto const read_eval_enabled{ read_eval_var.is_nil()
-                                  || !equal(read_eval_var->deref(), unknown_kw) };
+    auto const read_eval_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
+    auto const read_eval_enabled{ read_eval_var.is_some()
+                                  && !equal(read_eval_var->deref(), unknown_kw) };
 
     if(!read_eval_enabled)
     {
-      throw std::runtime_error{ util::format(
-        "Reading is disallowed when `clojure.core/*read-eval*` is bound to `:unknown`.") };
+      throw std::runtime_error{
+        read_eval_var.is_nil()
+          ? "Reading is disallowed when `clojure.core/*read-eval*` is unbound."
+          : "Reading is disallowed when `clojure.core/*read-eval*` is bound to `:unknown`."
+      };
     }
 
-
-    object_ref extended_features{};
-    bool in_preservation_mode{};
+    /* The `read` & similar `clojure.core` functions in Clojure, don't read in reader
+     * conditionals by default. This behavior can be configured using the `:read-cond`
+     * reader option using the following values:
+     *   - :allow - Reader conditionals are read in and processed by the parser.
+     *   - :preserve - Reader conditionals are read in but the exact form is preserved,
+     *                 along with the unsupported reader conditional features. */
     bool allow_reader_conditional{ false };
+    bool in_preservation_mode{};
+    /* The supported reader conditional features can be extended using the `:features`
+     * reader option. */
+    object_ref extended_features{};
 
-    visit_map_like(
-      [&](auto const typed_reader_opts) {
-        auto const read_cond_kw{ __rt_ctx->intern_keyword("", "read-cond").expect_ok() };
-        auto const read_cond{ get(typed_reader_opts, read_cond_kw) };
+    if(!is_map(reader_opts))
+    {
+      throw std::runtime_error{ util::format(
+        "The reader options need to be a map. Found a {} instead.",
+        runtime::object_type_str(reader_opts->type)) };
+    }
 
-        if(read_cond.is_some())
-        {
-          auto const reader_cond{ try_object<obj::keyword>(read_cond) };
+    static auto const read_cond_kw{ __rt_ctx->intern_keyword("", "read-cond").expect_ok() };
+    auto const read_cond{ get(reader_opts, read_cond_kw) };
 
-          if(reader_cond == __rt_ctx->intern_keyword("", "preserve").expect_ok())
-          {
-            in_preservation_mode = true;
-            allow_reader_conditional = true;
-          }
-          else if(reader_cond == __rt_ctx->intern_keyword("", "allow").expect_ok())
-          {
-            allow_reader_conditional = true;
-          }
-          else
-          {
-            throw std::runtime_error{ util::format(
-              "Only :preserve or :allow modes are supported for :read-cond reader option. Found {} "
-              "instead.",
-              runtime::to_code_string(read_cond)) };
-          }
-        }
+    if(read_cond.is_some())
+    {
+      auto const reader_cond{ try_object<obj::keyword>(read_cond) };
+      static auto const preserve_kw{ __rt_ctx->intern_keyword("", "preserve").expect_ok() };
+      static auto const allow_kw{ __rt_ctx->intern_keyword("", "allow").expect_ok() };
 
-        auto const features_kw{ __rt_ctx->intern_keyword("", "features").expect_ok() };
-        auto const features{ get(typed_reader_opts, features_kw) };
-
-        if(features.is_some())
-        {
-          visit_set_like([](auto const _) {},
-                         [&]() {
-                           throw std::runtime_error{ util::format(
-                             "The :features reader option needs to be a set. Found a {} instead.",
-                             runtime::object_type_str(features->type)) };
-                         },
-                         features);
-
-          extended_features = features;
-        }
-      },
-      [&]() {
+      if(reader_cond == preserve_kw)
+      {
+        in_preservation_mode = true;
+        allow_reader_conditional = true;
+      }
+      else if(reader_cond == allow_kw)
+      {
+        allow_reader_conditional = true;
+      }
+      else
+      {
         throw std::runtime_error{ util::format(
-          "The reader options need to be a map. Found a {} instead.",
-          runtime::object_type_str(reader_opts->type)) };
-      },
-      reader_opts);
+          "Only :preserve or :allow modes are supported for :read-cond reader option. Found {} "
+          "instead.",
+          runtime::to_code_string(read_cond)) };
+      }
+    }
+
+    static auto const features_kw{ __rt_ctx->intern_keyword("", "features").expect_ok() };
+    auto const features{ get(reader_opts, features_kw) };
+
+    if(features.is_some())
+    {
+      if(!is_set(features))
+      {
+        throw std::runtime_error{ util::format(
+          "The :features reader option needs to be a set. Found a {} instead.",
+          runtime::object_type_str(features->type)) };
+      }
+
+      extended_features = features;
+    }
 
     /* When reading an arbitrary string, we don't want the last *current-file* to
      * be set as source file, so we need to bind it to nil. */
     binding_scope const preserve{ obj::persistent_hash_map::create_unique(
       std::make_pair(current_file_var, jank_nil())) };
     read::lex::processor l_prc{ code };
-    read::parse::processor p_prc{ l_prc.begin(), l_prc.end() };
-    p_prc.extended_features = extended_features;
-    p_prc.in_preservation_mode = in_preservation_mode;
-    p_prc.allow_reader_conditional = allow_reader_conditional;
+    read::parse::processor p_prc{ l_prc.begin(),
+                                  l_prc.end(),
+                                  extended_features,
+                                  allow_reader_conditional,
+                                  in_preservation_mode };
+    auto const read_entire_stream{ nth_form == std::numeric_limits<u64>::max() };
     u64 count{};
     static auto const eof_kw{ __rt_ctx->intern_keyword("", "eof").expect_ok() };
     static auto const eof_throw_kw{ __rt_ctx->intern_keyword("", "eofthrow").expect_ok() };
-    auto const throw_on_eof{ equal(get(reader_opts, eof_kw, eof_throw_kw), eof_throw_kw) };
-    auto throw_eof{ throw_on_eof };
-    object_ref ret{ get(reader_opts, eof_kw) };
+    auto const eof_value{ get(reader_opts, eof_kw) };
+    auto const throw_on_eof{ equal(eof_value, eof_throw_kw) };
+    bool eof_found{ throw_on_eof };
+    object_ref ret{};
 
     for(auto const &form : p_prc)
     {
@@ -395,27 +407,27 @@ namespace jank::runtime
         break;
       }
 
-      ++count;
-
       if(form.expect_ok().is_none())
       {
-        if(throw_on_eof && count <= 0)
-        {
-          throw_eof = true;
-          break;
-        }
-
-        ret = get(reader_opts, eof_kw);
+        eof_found = true;
+        ret = eof_value;
       }
       else
       {
+        /* In Clojure, the reader doesn't count comments, etc. as forms. */
+        ++count;
+        eof_found = false;
         ret = form.expect_ok().unwrap().ptr;
       }
-
-      throw_eof = false;
     }
 
-    if(throw_eof)
+    if(!read_entire_stream && count < nth_form)
+    {
+      eof_found = true;
+      ret = eof_value;
+    }
+
+    if(throw_on_eof && eof_found)
     {
       throw std::runtime_error{ "EOF reached while reading. To override this behavior, provide a "
                                 "return value for the EOF case via the `:eof` reader option." };
@@ -426,18 +438,23 @@ namespace jank::runtime
 
   object_ref context::read_string(jtl::immutable_string const &code, object_ref const reader_opts)
   {
-    return read_string(code, reader_opts, std::numeric_limits<u64>::max());
+    /* The Clojure `read` & similar functions always read the first form in a string
+     * containing more than one form. */
+    return read_string(code, reader_opts, /* nth_form */ 1);
   }
 
   object_ref context::read_string(jtl::immutable_string const &code)
   {
-    return read_string(code, obj::persistent_array_map::empty());
+    static auto const eof_kw{ __rt_ctx->intern_keyword("", "eof").expect_ok() };
+    static auto const eof_throw_kw{ __rt_ctx->intern_keyword("", "eofthrow").expect_ok() };
+    static auto const default_reader_opts{ obj::persistent_array_map::create_unique(eof_kw,
+                                                                                    eof_throw_kw) };
+    return read_string(code, default_reader_opts);
   }
 
   object_ref context::forcefully_read_string(jtl::immutable_string const &code)
   {
-    auto const read_eval_enabled_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
-    /* TODO: Profile C++ codegen. */
+    static auto const read_eval_enabled_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
     binding_scope const bindings{ obj::persistent_hash_map::create_unique(
       std::make_pair(read_eval_enabled_var, jank_true)) };
     return read_string(code);
@@ -452,7 +469,7 @@ namespace jank::runtime
       throw file.expect_err();
     }
 
-    return read_string(file.expect_ok().data(), reader_opts);
+    return read_string(file.expect_ok().data(), reader_opts, std::numeric_limits<u64>::max());
   }
 
   native_vector<analyze::expression_ref>
--- a/compiler+runtime/src/cpp/jank/runtime/core.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core.cpp
@@ -884,23 +884,21 @@ namespace jank::runtime
     if(form_string->type != object_type::persistent_string)
     {
       throw std::runtime_error{ util::format(
-        "Argument to `read` needs to be a string representing a form. Found a {} instead.",
+        "Argument to `read-string` needs to be a string representing a form. Found a {} instead.",
         object_type_str(form_string->type)) };
     }
 
     auto const typed_o{ expect_object<obj::persistent_string>(form_string) };
-    /* The Clojure `read-string` always returns the first form in a string containing more than
-     * one form. */
-    return __rt_ctx->read_string(typed_o->data, opts, /* nth_form */ 1);
+    return __rt_ctx->read_string(typed_o->data, opts);
   }
 
   object_ref read_file(object_ref const file_path, object_ref const opts)
   {
     if(file_path->type != object_type::persistent_string)
     {
-      throw std::runtime_error{ util::format(
-        "Argument to `read` needs to be a string representing a file path. Found a {} instead.",
-        object_type_str(file_path->type)) };
+      throw std::runtime_error{ util::format("Argument to `read-file` needs to be a string "
+                                             "representing a file path. Found a {} instead.",
+                                             object_type_str(file_path->type)) };
     }
 
     auto const typed_o{ expect_object<obj::persistent_string>(file_path) };
--- a/compiler+runtime/src/cpp/jank/runtime/obj/reader_conditional.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/reader_conditional.cpp
@@ -49,22 +49,17 @@ namespace jank::runtime::obj
     return buff.release();
   }
 
-  jtl::immutable_string reader_conditional::to_code_string() const
-  {
-    return to_string();
-  }
-
   object_ref reader_conditional::get(object_ref const key, object_ref const fallback) const
   {
     static auto const form_kw{ __rt_ctx->intern_keyword("form").expect_ok() };
     static auto const splicing_qmark_kw{ __rt_ctx->intern_keyword("splicing?").expect_ok() };
 
-    if(runtime::equal(key, form_kw))
+    if(key == form_kw)
     {
       return form;
     }
 
-    if(runtime::equal(key, splicing_qmark_kw))
+    if(key == splicing_qmark_kw)
     {
       return splicing;
     }
@@ -85,7 +80,6 @@ namespace jank::runtime::obj
     return form_kw == key || splicing_qmark_kw == key;
   }
 
-  /* TODO: Cache this. */
   uhash reader_conditional::to_hash() const
   {
     return hash::combine(splicing->to_hash(), form->to_hash());
--- a/compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
@@ -52,7 +52,6 @@ namespace jank::runtime::obj
     return to_string();
   }
 
-  /* TODO: Cache this. */
   uhash tagged_literal::to_hash() const
   {
     return hash::combine(hash::visit(tag.get()), hash::visit(form.get()));
--- a/compiler+runtime/include/cpp/jank/read/parse.hpp
+++ b/compiler+runtime/include/cpp/jank/read/parse.hpp
@@ -61,6 +61,11 @@ namespace jank::read::parse
     };
 
     processor(lex::processor::iterator const &b, lex::processor::iterator const &e);
+    processor(lex::processor::iterator const &b,
+              lex::processor::iterator const &e,
+              runtime::object_ref const &extended_features,
+              bool const &allow_reader_conditional,
+              bool const &in_preservation_mode);
 
     object_result next();
     object_result parse_list();
@@ -131,7 +136,7 @@ namespace jank::read::parse
      * turns one form into many. */
     native_list<runtime::object_ref> pending_forms;
     /* The Clojure reader allows extending the supported reader conditional features
-     * via reader options.*/
+     * via reader options. */
     runtime::object_ref extended_features{};
     lex::token latest_token;
     jtl::option<shorthand_function_details> shorthand;
@@ -148,11 +153,11 @@ namespace jank::read::parse
      * jank will suppress the reader for the `:default` feature as well. This
      * behaviour is consistent with the Clojure reader. */
     bool is_reader_suppressed{};
+    bool allow_reader_conditional{ true };
     /* When reading in a form via the Clojure read functions the end user might wish
      * to keep even the unsupported reader conditional branches, in such cases they
      * can opt-in to the preserve mode by setting the :read-cond reader option. */
     bool in_preservation_mode{};
-    bool allow_reader_conditional{ true };
     /* Whether or not the next form is considered quoted. */
     bool quoted{};
     /* Whether or not the next form is considered syntax-quoted. */
--- a/compiler+runtime/include/cpp/jank/runtime/obj/reader_conditional.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/reader_conditional.hpp
@@ -24,7 +24,6 @@ namespace jank::runtime::obj
     bool equal(object const &) const override;
     jtl::immutable_string to_string() const override;
     void to_string(jtl::string_builder &buff) const override;
-    jtl::immutable_string to_code_string() const override;
     uhash to_hash() const override;
 
     /* behavior::get */


shantanu-sardesai:
@jeaye 

My current implementation cause jank to crash for the following code:

```clj
(binding [clojure.core/*read-eval* true] (read-string {:eof false :read-cond :preserve} "#?(:jank)"))
```

This is because we throw [parse_invalid_reader_conditional](https://github.com/shantanu-sardesai/jank/blob/4d6a14865c92a933648f9053869889be50784b50/compiler+runtime/src/cpp/jank/read/parse.cpp#L1153) error which I think is not playing well with the error reporter.

---

### Backtrace

```
Process 88216 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00000001838c2388 libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`__pthread_kill:
->  0x1838c2388 <+8>:  b.lo   0x1838c23a8    ; <+40>
    0x1838c238c <+12>: pacibsp
    0x1838c2390 <+16>: stp    x29, x30, [sp, #-0x10]!
    0x1838c2394 <+20>: mov    x29, sp
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00000001838c2388 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x00000001838fb848 libsystem_pthread.dylib`pthread_kill + 296
    frame #2: 0x00000001838049e4 libsystem_c.dylib`abort + 124
    frame #3: 0x00000001004d6d28 jank`jtl::do_assertion_panic(msg=0x000000016fdfcd58) at assert.cpp:12:5
    frame #4: 0x000000010057f7bc jank`jank::error::snippet::add(this=0x000000010db09870, body_source=0x000000010db6df30, n=0x000000010db6df10) at report.cpp:155:5
    frame #5: 0x0000000100582a90 jank`jank::error::plan::add(this=0x000000016fdfd348, body_source=0x000000010db6df30, n=0x000000010db6df10) at report.cpp:369:17
    frame #6: 0x000000010057f2ac jank`jank::error::plan::add(this=0x000000016fdfd348, n=0x000000010db6df10) at report.cpp:354:5
    frame #7: 0x000000010057f1e4 jank`jank::error::plan::plan(this=0x000000016fdfd348, e=jank::error_ref @ 0x000000016fdfcf00) at report.cpp:100:7
    frame #8: 0x000000010057f308 jank`jank::error::plan::plan(this=0x000000016fdfd348, e=jank::error_ref @ 0x000000016fdfcf38) at report.cpp:95:3
    frame #9: 0x0000000100582bf0 jank`jank::error::report(e=jank::error_ref @ 0x000000016fdfd3a0) at report.cpp:537:16
    frame #10: 0x000000010053cf7c jank`jank::util::print_exception(e=jank::error_ref @ 0x000000016fdfd4c8) at try.cpp:112:5
    frame #11: 0x0000000100005598 jank`jank::repl() at main.cpp:277:7
    frame #12: 0x000000010000356c jank`main::$_0::operator()(this=0x000000016fdfe21f, argc=2, argv=0x000000016fdfea30) const at main.cpp:414:9
    frame #13: 0x0000000100003138 jank`main::$_0::__invoke(argc=2, argv=0x000000016fdfea30) at main.cpp:361:60
    frame #14: 0x00000001004fe4d4 jank`jank_init_with_pch(argc=2, argv=0x000000016fdfea30, init_default_ctx='\0', pch_data=0x0000000000000000, pch_size=0, fn=(jank`main::$_0::__invoke(int, char const**) at main.cpp:361)) at c_api.cpp:1077:14
    frame #15: 0x00000001004fe318 jank`jank_init(argc=2, argv=0x000000016fdfea30, init_default_ctx='\0', fn=(jank`main::$_0::__invoke(int, char const**) at main.cpp:361)) at c_api.cpp:1026:12
    frame #16: 0x0000000100002af4 jank`main(argc=2, argv=0x000000016fdfea30) at main.cpp:361:10
    frame #17: 0x000000018355ab98 dyld`start + 6076
```

shantanu-sardesai:
Probably caused due to https://github.com/jank-lang/jank/issues/647.

Shall I just repackage our parser (compile-time) error into a runtime error (`std::runtime_error`) instead? Since anyways there will be no source info available, at least at the moment.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1092,6 +1092,45 @@ namespace jank::read::parse
 
   processor::object_result processor::parse_reader_macro_conditional(bool const splice)
   {
+    auto const reader_opts{ __rt_ctx->reader_opts_var->deref() };
+    auto const has_reader_opts{ reader_opts != jank_nil() };
```
jeaye: Please use `.is_some()`.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1092,6 +1092,45 @@ namespace jank::read::parse
...
+    bool in_preserve_mode{};
```
jeaye: Please add a comment explaining what in the heck preserve mode is.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1092,6 +1092,45 @@ namespace jank::read::parse
...
+    obj::persistent_hash_set_ref features{};
```
jeaye: **You will not know that features is a hash set.** It could be a sorted set, or some other set-like thing. You cannot assume this. Just use an `object_ref`.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1092,6 +1092,45 @@ namespace jank::read::parse
...
+
+    if(has_reader_opts)
+    {
+      auto const reader_opts_map{ try_object<obj::persistent_array_map>(reader_opts) };
```
jeaye: **You will not know that the opts is an array map.** It could be a hash map, or a sorted map, or some other map-like thing. You cannot assume this. Just use an `object_ref`.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1092,6 +1092,45 @@ namespace jank::read::parse
...
+      auto const read_cond_kw{ __rt_ctx->intern_keyword("", "read-cond").expect_ok() };
+      auto const read_cond{ reader_opts_map->get(read_cond_kw) };
+
+      if(read_cond == jank_nil())
```
jeaye: Please use `.is_nil()`.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1092,6 +1092,45 @@ namespace jank::read::parse
...
+      if(reader_cond == __rt_ctx->intern_keyword("", "preserve"))
+      {
+        in_preserve_mode = true;
+      }
+      else if(reader_cond == __rt_ctx->intern_keyword("", "allow"))
+      {
+      }
+      else
+      {
+        throw std::runtime_error{ "Conditional read not allowed" };
```
jeaye: We can provide a better error here, but showing what we found.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1092,6 +1092,45 @@ namespace jank::read::parse
...
+      }
+
+      auto const features_kw{ __rt_ctx->intern_keyword("", "features").expect_ok() };
+      auto const feature_set{ reader_opts_map->get(features_kw) };
+
+      if(feature_set != jank_nil())
```
jeaye: Please use `.is_some()`.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -307,6 +312,57 @@ namespace jank::runtime
...
 
+  object_ref context::read_file(jtl::immutable_string const &file_path)
+  {
+    auto const file(module::loader::read_file(file_path));
+    if(file.is_err())
+    {
+      throw file.expect_err();
+    }
+
+    return read_string(file.expect_ok().view());
```
jeaye: Why use `view()` here? It's not correct, since `read_string` doesn't take a view. Views are very dangerous and should not be used without extreme caution.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -307,6 +312,57 @@ namespace jank::runtime
...
+  }
+
+  object_ref context::read_first_form(jtl::immutable_string const &code,
+                                      obj::persistent_array_map_ref const &reader_opts)
```
jeaye: Taking an `oref` by `const &` is inefficient. You have a reference to a pointer. Just take it by value.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -307,6 +312,57 @@ namespace jank::runtime
...
+  {
+    /* Clojure usually loads the value for `clojure.core/\*read-eval*` from the properties file.
+     * See here: https://github.com/clojure/clojure/blob/6a4ba6aedc8575768b2fff6d9c9c7e6503a0a93a/src/jvm/clojure/lang/RT.java#L198 */
+    auto const reader_eval_enabled{ __rt_ctx->find_var("clojure.core", "*read-eval*")->deref() };
+
+    if(reader_eval_enabled != jank_true)
```
jeaye: This assumes it's exactly equal to `true`, but that's rarely how things work in Clojure. Instead, use `truthy`.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -307,6 +312,57 @@ namespace jank::runtime
...
+    if(first_form.expect_ok().is_none())
+    {
+      auto const eof_kw{ __rt_ctx->intern_keyword("", "eof").expect_ok() };
+
+      if(!reader_opts->contains(eof_kw))
+      {
+        throw std::runtime_error{ "EOF while reading." };
+      }
+
+      return reader_opts->get(eof_kw);
```
jeaye: We need to handle the `:eofthrow` case.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -701,4 +701,44 @@ namespace jank::runtime
 
     return reference;
   }
+
+  object_ref read_string(object_ref const form_string, object_ref const opts)
+  {
+    return visit_object(
```
jeaye: This is a whole visit to check for one type. It's like bringing out an encyclopedia to check if something is an apple. Just see if it's an apple by looking at it. i.e. check the type of `o` or use `dyn_cast`.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -701,4 +701,44 @@ namespace jank::runtime
...
+      [](auto const typed_o, auto const opts) -> object_ref {
+        using T = typename decltype(typed_o)::value_type;
+        auto const typed_opts{ try_object<obj::persistent_array_map>(opts) };
```
jeaye: This is **not correct**, as I explained above.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -701,4 +701,44 @@ namespace jank::runtime
...
+          throw std::runtime_error{ "Argument to read needs to be a string representing a form." };
+        }
+      },
+      form_string,
+      opts);
+  }
+
+  object_ref read(object_ref const file_path)
+  {
+    return visit_object(
```
jeaye: No need for this visit.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -701,4 +701,44 @@ namespace jank::runtime
...
+        else
+        {
+          throw std::runtime_error{ "Argument to read needs to be a string representing a form." };
+        }
+      },
+      form_string,
+      opts);
+  }
+
+  object_ref read(object_ref const file_path)
```
jeaye: This should be called `read_file`. Clojure's default `read` will read from a stream, which is something jank still may do in the future.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -5707,7 +5707,7 @@
   ([s]
    (read-string {} s))
   ([opts s]
-   (cpp/clojure.core_native.read_string opts s)))
+   (cpp/jank.runtime.read_string s opts)))
```
jeaye, line 5714: We were calling `core_native::read_string`. Now we're not. What happens to that fn? Let's not leave dead code lying around.

jeaye:
> Probably caused due to #647.
> 
> Shall I just repackage our parser (compile-time) error into a runtime error (`std::runtime_error`) instead? Since anyways there will be no source info available, at least at the moment.

This is from a failed assertion. Why did the assertion fail? What were the values? You are suggesting hacky workarounds, but we want to first understand the problem and then consider robust solutions. Hacky workarounds are a last resort.

compiler+runtime/src/cpp/jank/read/parse.cpp
shantanu-sardesai: Yeah, for now I'm mimicking what Clojure throws here. Will clean this up on my next pass.

compiler+runtime/src/cpp/jank/runtime/context.cpp
shantanu-sardesai: I originally thought this was intentional, since reading in code at runtime is a potential attack vector and the Clojure implementation might have intentionally chosen to bypass it's own `truthy` behaviour here.

Turns out Clojure checks for `:unkown` for `clojure.core/*read-eval*` and throws an exception in that case else just reads in the code.

```clj
user=> (binding [clojure.core/*read-eval* false] (read-string "{}"))
{}

user=> (binding [clojure.core/*read-eval* true] (read-string "{}"))
{}

user=> (binding [clojure.core/*read-eval* :unknown] (read-string "{}"))
Execution error at user/eval138 (REPL:1).
Reading disallowed - *read-eval* bound to :unknown
```

Anyways, I've fixed this bit as well.

compiler+runtime/src/cpp/jank/read/parse.cpp
shantanu-sardesai: Just confirming, here what you are suggesting is that in the runtime error message I should also print the value for `:read-cond` that was provided by the end user, correct?

compiler+runtime/src/cpp/jank/read/parse.cpp
shantanu-sardesai: I agree that this is not the correct method, it popped in my mind while I added this bit.

Shall I instead use `visit_set_like` & `visit_map_like` to get the values on specific keys? Since I need to `->get` a value from within the collection. Or is there an even better approach?

jeaye: No, don't implement these things yourself. We already have a slew of functions just for this. Look in `runtime/core.hpp` and `runtime/core/`.

compiler+runtime/src/jank/clojure/core.jank
shantanu-sardesai: I'm trying to reuse the existing implementation of `jank::runtime::context::read_string` which was used in the `core_native::read_string` by [modifying](https://github.com/jank-lang/jank/pull/673/commits/3cd1c06f0aa6446831e1a0b33396e73a17c9d056) it with `opts` support although that ended up breaking C++ codegen.

I [fixed](https://github.com/jank-lang/jank/pull/673/commits/0320155d59307993f7884a9a3917f65ea2a12947) the first error that popped up but now figuring out how to fix the following error:

```
─ analyze/macro-expansion-exception ────────────────────────────────────────────
error: Uncaught exception while expanding macro.

─────┬──────────────────────────────────────────────────────────────────────────
     │ …Desktop/code/projects/jank/compiler+runtime/src/jank/clojure/core.jank
─────┼──────────────────────────────────────────────────────────────────────────
651  │
652  │ (defn macroexpand-1
     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Found here.
653  │   "If form represents a macro form, returns its expansion,
654  │    else returns form."
655  │   [form]
656  │   (cpp/jank.runtime.macroexpand1 form))
─────┴──────────────────────────────────────────────────────────────────────────
─ internal/failure ─────────────────────────────────────────────────────────────
error: Can't intern var. Sym isn't qualified: unknown


#0  in jank::error::internal_failure at jank-phase-1
#1  in jank::error::throw_internal_failure at jank-phase-1
#2  in jtl::immutable_string>> at result.hpp:60
#3  in jtl::immutable_string>::assert_ok at result.hpp:155
#4  in jtl::immutable_string>::expect_ok at result.hpp:160
#5  in jank::runtime::context::read_string at jank-phase-1
#6  in jank::runtime::context::read_string at jank-phase-1
#7
#8
#9
#10
#11
#12 in clang::Interpreter::Execute at libclang-cpp.dylib
#13 in clang::Interpreter::ParseAndExecute at libclang-cpp.dylib
#14 in Cpp::Interpreter::ParseAndExecute at CppInterOpInterpreter.h:190
#15 in jank::jit::processor::eval_string at jank-phase-1
#16 in jank::runtime::obj::deferred_cpp_function::call at jank-phase-1
#17 in __ZZN4jank7runtime12dynamic_callENS0_4orefINS0_6objectEEES3_S3_S3_S3_S3_S3_ENK3$_0clINS1_INS0_3obj21deferred_cpp_functionEEEEES3_T_ at jank-phase-1
#18 in jank::runtime::visit_object<jank::runtime::dynamic_call at jank-phase-1
#19 in jank::runtime::dynamic_call at jank-phase-1
#20 in __ZZN4jank7runtime8apply_toENS0_4orefINS0_6objectEEES3_ENK3$_0clINS1_INS0_3obj4consEEEEES3_T_ at jank-phase-1
#21 in jank::runtime::visit_seqable<jank::runtime::apply_to at jank-phase-1
#22 in jank::runtime::visit_seqable<jank::runtime::apply_to at jank-phase-1
#23 in jank::runtime::apply_to at jank-phase-1
#24 in __ZZN4jank7runtime7context12macroexpand1ENS0_4orefINS0_6objectEEEENK3$_1clINS2_INS0_3obj15persistent_listEEEEES4_T_ at jank-phase-1
#25 in jank::runtime::visit_seqable<jank::runtime::context::macroexpand1 at jank-phase-1
#26 in jank::runtime::context::macroexpand1 at jank-phase-1
#27 in jank::runtime::context::macroexpand at jank-phase-1
#28 in jank::analyze::processor::analyze_call at jank-phase-1
#29 in __ZZN4jank7analyze9processor7analyzeENS_7runtime4orefINS2_6objectEEEN3jtl3ptrINS0_11local_frameEEENS0_19expression_positionERKNS6_6optionINS6_3refINS0_4expr16function_contextEEEEEbENK3$_0clINS3_INS2_3obj15persistent_listEEEEENS6_6resultINSC_INS0_10expressionEEENSC_INS_5error4baseEEEEET_ at jank-phase-1
#30 in jank::runtime::visit_object<jank::analyze::processor::analyze at jank-phase-1
#31 in jank::analyze::processor::analyze at jank-phase-1
#32 in jank::analyze::processor::analyze at jank-phase-1
#33 in jank::runtime::context::eval_string at jank-phase-1
#34 in jank::runtime::context::eval_file at jank-phase-1
#35 in jank::runtime::module::loader::load_jank at jank-phase-1
#36 in jank::runtime::module::loader::load at jank-phase-1
#37 in jank::runtime::context::load_module at jank-phase-1
#38 in jank::runtime::context::compile_module at jank-phase-1
#39 in jank::compile_module at main.cpp:177
#40 in main::$_0::operator at main.cpp:411
#41 in main::$_0::__invoke at main.cpp:361
#42 in _jank_init_with_pch at jank-phase-1
#43 in _jank_init at jank-phase-1
#44 in main at main.cpp:361
#45 in start + 6075 at dyld
```

jeaye: Make sure you have 4f0ad027425a60f629f6857eaecbf9cfcc27d8b6 (latest main).

shantanu-sardesai: Yep, I rebased today morning when I saw the `future` support message 🙂.

jeaye: > Yep, I rebased today morning when I saw the `future` support message 🙂.

So you have this error even with the latest main?

shantanu-sardesai: No, `main` compiles fine. Since I changed the `read_string` which seems to be heavily used in the C++ codegen I might have broken something.

shantanu-sardesai: I'll take a look at the generated C++ code to see if I can figure what I've broken. I suspect the issue is due to me introducing the `nth_form`, not sure though.

jeaye: We use `read_string` for reading complex meta forms, instead of encoding them into C++. That should not break unless you broke the API of behavior of `read_string` though. Your changes to `read_string` should only be additive here, unless we have a very good reason to break existing behavior.

shantanu-sardesai: Figured out what the issue was, fixed!

shantanu-sardesai:
@jeaye 

> https://github.com/jank-lang/jank/pull/673#issuecomment-3785194300

The crash was also fixed, now it gives the same output as was reported in this issue (https://github.com/jank-lang/jank/issues/647).

```jank
user=> (read-string {:eof false :read-cond :preserve} "#?(:jank)")
─ parse/invalid-reader-conditional ─────────────────────────────────────────────────────────────────
error: Invalid reader conditional.

     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-WQl59K
  1  │ (read-string {:eof false :read-cond :preserve} "#?(:jank)")
     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Used here.
╭ nil ─────────────────────────────────────────────────────────────────────────────────────────────╮
│Unable to map file: Unable to find module 'user'.                                                 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

shantanu-sardesai:
Although part of the CI is busted.

compiler+runtime/src/cpp/jank/runtime/core.cpp
shantanu-sardesai: I'm also considering implementing the other siblings `read`, `read+string` & `read-line` in this PR. What is the closest equivalent for Java's `java.io.PushbackReader` in C++? Is it [std::basic_istream](https://en.cppreference.com/w/cpp/io/basic_istream.html)? Or will it not be so straight forward in C++?

shantanu-sardesai:
Also, is there a way to know when `read_string` is being invoked by the jank compiler for C++ codegen and when it's being used by the end user? At the moment, `clojure.core/*read-eval*` could be disabled by binding the `var` and that could [hinder](https://github.com/jank-lang/jank/pull/673/changes#diff-241f5ba195a764892c092ad80b4e5b53c0dc994974ddc4f5737c4eafbe3b3529R308) the behaviour of codegen.

Or perhaps it won't matter? Since the `clojure.core/*read-eval*` var can be overwritten by the end user only _after_ our compiler is done with codegen 🤔. Although the issue of initialising the `clojure.core/*read-eval*` with `:unknown` on start-up still remains (currently we don't support that, I've just placed a `TODO` there, perhaps I'll add a note about the potential for this issue there as well).

compiler+runtime/include/cpp/jank/runtime/context.hpp
```diff
@@ -83,7 +83,17 @@ namespace jank::runtime
     jtl::option<object_ref> eval_file(jtl::immutable_string const &path);
     jtl::option<object_ref> eval_string(jtl::immutable_string const &code) const;
     jtl::result<void, error_ref> eval_cpp_string(jtl::immutable_string const &code) const;
-    object_ref read_string(jtl::immutable_string const &code);
+    object_ref read_string(jtl::immutable_string const &code,
+                           object_ref const reader_opts,
+                           jank::u64 const nth_form = std::numeric_limits<jank::u64>::max());
```
jeaye: We don't use default parameters unless we have to. Here, we certainly don't have to. If we need to support this param, add an overload which takes it.

Also, there's no need for `jank::` here. `u64` is already in the `jank` namespace.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1086,6 +1086,55 @@ namespace jank::read::parse
...
+      auto const read_cond_kw{ __rt_ctx->intern_keyword("", "read-cond").expect_ok() };
+      auto const read_cond{ get(reader_opts, read_cond_kw) };
+
+      if(read_cond.is_nil())
+      {
+        return error::parse_invalid_reader_conditional(
+          { start_token.start, latest_token.end },
+          util::format("Conditional read is not allowed by default. Set the :read-cond reader "
+                       "option to either :preserve or :allow. Found {} instead.",
+                       runtime::to_code_string(read_cond)));
```
jeaye: We know it's `nil`. We just checked that. No point formatting it in.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -76,6 +76,11 @@ namespace jank::runtime
     assert_var->bind_root(jank_true);
     assert_var->dynamic.store(true);
 
+    auto const reader_opt_sym(make_box<obj::symbol>("*reader-opts*"));
+    reader_opts_var = core->intern_var(reader_opt_sym);
+    reader_opts_var->bind_root(jank_nil());
+    reader_opts_var->dynamic.store(true);
```
jeaye: You're doing this in multiple places.

1. Here (runtime context), setting to `nil`
2. Health check, setting to `true`
3. AOT processor, setting to `true`
4. In main, setting to `true`
5. In test main, setting to `true`

So, we're always setting this to `true` anyway. Why not just set it here and remove the others?

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -289,27 +294,92 @@ namespace jank::runtime
...
-  object_ref context::read_string(jtl::immutable_string const &code)
+  object_ref context::read_string(jtl::immutable_string const &code,
+                                  object_ref const reader_opts,
+                                  jank::u64 const nth_form)
   {
     profile::timer const timer{ "rt read_string" };
-
+    auto const unknown_kw{ __rt_ctx->intern_keyword("unknown").expect_ok() };
+    auto const read_eval_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
+    auto const read_eval_enabled{ read_eval_var.is_nil() || !equal(read_eval_var, unknown_kw) };
```
jeaye: No way this is valid. You're comparing the var, not what's inside the var.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -289,27 +294,92 @@ namespace jank::runtime
...
 
-  object_ref context::read_string(jtl::immutable_string const &code)
+  object_ref context::read_string(jtl::immutable_string const &code,
+                                  object_ref const reader_opts,
+                                  jank::u64 const nth_form)
   {
     profile::timer const timer{ "rt read_string" };
-
+    auto const unknown_kw{ __rt_ctx->intern_keyword("unknown").expect_ok() };
+    auto const read_eval_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
```
jeaye: These can be static.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -289,27 +294,92 @@ namespace jank::runtime
...
+    {
+      throw std::runtime_error{ util::format(
+        "Reading is disallowed when `clojure.core/*read-eval*` is bound to `:unknown`.") };
+    }
 
     read::lex::processor l_prc{ code };
     read::parse::processor p_prc{ l_prc.begin(), l_prc.end() };
+    auto count{ nth_form };
+    auto const eof_kw{ __rt_ctx->intern_keyword("", "eof").expect_ok() };
+    auto const eof_throw_kw{ __rt_ctx->intern_keyword("", "eofthrow").expect_ok() };
```
jeaye: These can be static.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -289,27 +294,92 @@ namespace jank::runtime
...
+    auto const throw_on_eof{ equal(get(reader_opts, eof_kw, eof_throw_kw), eof_throw_kw) };
+    auto throw_eof{ throw_on_eof };
+    object_ref ret{ get(reader_opts, eof_kw) };
 
-    object_ref ret{};
     for(auto const &form : p_prc)
     {
-      ret = form.expect_ok().unwrap().ptr;
+      if(count <= 0)
```
jeaye: `count` is a `u64` it categorically cannot be `< 0`. If it's `0` and you decrement, it'll wrap around to be the max `u64` value.

Decrementing unsigned values is a great way to end up in an infinite loop, so extra caution is needed. I'm not confident that sufficient caution has been used here. It would be better to count up to the max instead.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -289,27 +294,92 @@ namespace jank::runtime
...
+                                "return value for the EOF case via the `:eof` reader option." };
     }
 
     return ret;
   }
 
+  object_ref context::forcefully_read_string(jtl::immutable_string const &code)
+  {
+    auto const read_eval_enabled_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
+    /* TODO: Profile C++ codegen. */
```
jeaye: Profile codegen for what?

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -866,4 +866,26 @@ namespace jank::runtime
...
   }
+
+  object_ref read_string(object_ref const form_string, object_ref const opts)
+  {
+    if(form_string->type != object_type::persistent_string)
+    {
+      throw std::runtime_error{ "Argument to read needs to be a string representing a form." };
+    }
+
+    auto const typed_o{ dyn_cast<obj::persistent_string>(form_string) };
```
jeaye: There's no need for `dyn_cast`, since you just did a type check. `dyn_cast` will do a type check for you and give you back a `nil` if it's not the correct type. So either just use `dyn_cast` and check for `nil` or use a type check and then `expect_object`. Not both.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -866,4 +866,26 @@ namespace jank::runtime
...
+    return __rt_ctx->read_string(typed_o->data, opts, 1);
```
jeaye: Why is this `1`? (Answer with a comment in the code)

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -866,4 +866,26 @@ namespace jank::runtime
...
+  }
+
+  object_ref read_file(object_ref const file_path, object_ref const opts)
+  {
+    if(file_path->type != object_type::persistent_string)
+    {
+      throw std::runtime_error{ "Argument to read needs to be a string representing a file path." };
```
jeaye: We can do better here.

1. Put read in '' so it's clear that we're talking about a fn
2. Format in the file path that we were given, or at least the type of it

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -866,4 +866,26 @@ namespace jank::runtime
...
+    }
+
+    auto const typed_o{ dyn_cast<obj::persistent_string>(file_path) };
```
jeaye: Again, invalid use of `dyn_cast`.

compiler+runtime/src/cpp/jank/util/cli.cpp
```diff
@@ -283,6 +284,10 @@ OPTIONS
         {
           opts.libs.emplace_back(value);
         }
+        else if(check_flag(it, end, value, "--disable-read", false))
+        {
+          opts.disable_read = true;
+        }
```
jeaye, line 311: Why is this a CLI flag? Who wants to set it? What's the use case? Does Clojure do this?

We're setting it to `true` in every single entrypoint we have, aside from if this flag is passed.

Also, since we already have a dynamic var for this, why not just allow the user to control this by setting the var (which they can already do)? The flag is redundant and doesn't help scenarios like AOT built executables.

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -62,12 +62,12 @@ extern "C"
   jank_object_ref jank_read_string(jank_object_ref const s)
   {
     auto const s_obj(try_object<obj::persistent_string>(reinterpret_cast<object *>(s)));
-    return __rt_ctx->read_string(s_obj->data).erase().data;
+    return __rt_ctx->read_string(s_obj->data, make_box<obj::persistent_array_map>()).erase().data;
```
jeaye: If you use `obj::persistent_array_map::empty()`, you can skip on the allocation. (see other comment about overload so this isn't needed)

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -62,12 +62,12 @@ extern "C"
...
   }
 
   jank_object_ref jank_read_string_c(char const * const s)
   {
-    return __rt_ctx->read_string(s).erase().data;
+    return __rt_ctx->read_string(s, make_box<obj::persistent_array_map>()).erase().data;
```
jeaye: If we provide an overload instead, we don't need to update every call to `read_string` to also take a useless empty map. We've negatively impacted our API here. It's worse to use as a result. We can see that from the changes required throughout this PR. So, we should use an overload instead. Then all of these call sites can actually remain unchanged.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1086,6 +1086,55 @@ namespace jank::read::parse
     auto const start_token(token_current.latest.unwrap().expect_ok());
     ++token_current;
 
+    auto const reader_opts{ __rt_ctx->reader_opts_var->deref() };
+    auto const has_reader_opts{ reader_opts.is_some() };
+    // When reading in a form via the Clojure read functions the end user might wish
+    // to keep even the unsupported reader conditional branches, in such cases they
+    // can opt-in to the preserve mode by setting the :read-cond reader option.
```
jeaye: Please use `/* */` style comments.

compiler+runtime/include/cpp/jank/runtime/context.hpp
```diff
@@ -83,7 +83,17 @@ namespace jank::runtime
...
+    /* jank uses its own `read_string` to parse object metadata.
+     * To prevent users from accidentally disabling `read-string` for jank’s
+     * internal C++ codegen by setting the `*read-eval*` var, we use a
+     * custom wrapper around `read_string` that always enables reading for
+     * jank’s internal codegen paths.
+     */
+    object_ref forcefully_read_string(jtl::immutable_string const &code);
```
jeaye, line 95: I think the comments for fns are unique in that they generally start by saying *what* the function does, followed by *why* it exists. For this comment, I think we're jumping too far into the weeds first without actually explaining when this function should be used. For example:

> Calls `read-string` on the code, regardless of the value of `*read-eval*`.
> This is important, since jank uses `read-string` as part of codegen, for reading
> metadata maps and other non-trivial literals.

It's fewer words, but I think this says more to the point of what's crucial here.

jeaye (changes requested):
Thanks, Shantanu!

compiler+runtime/src/cpp/jank/util/cli.cpp
shantanu-sardesai: Clojure provides an extra configuration option to disable **all** reads via a Java system property ([here](https://github.com/clojure/clojure/blob/6a4ba6aedc8575768b2fff6d9c9c7e6503a0a93a/src/jvm/clojure/lang/RT.java#L198-L198)) which by default sets the value to `true` when not provided.

Since jank doesn't have any equivalent to a system property, we had decided that this should just be a CLI flag. Although I'm not exactly clear on how `read-string` behaves in AOT scenarios and whether if it's even relevant there.

jeaye: `read-string` still behaves normally in our AOT scenarios. The CLI flag will not help there. I think we should remove the flag until we have a clear use case for it.

compiler+runtime/src/cpp/jank/runtime/context.cpp
shantanu-sardesai: Not the same `var`, the one that is being set in multiple places is the `*read-eval*`. Since we've decided to **_not_** support the CLI flag for the moment, I'll remove the other usages and just initialise this `var` in a single place.

compiler+runtime/src/cpp/jank/c_api.cpp
shantanu-sardesai: Heh, I had it but [removed](https://github.com/jank-lang/jank/pull/673/changes/df3ba8dc399bff9e387dbf865e56a775b2162ca3) it because I thought that the overload was polluting the API.

I'll bring it back 👍.

compiler+runtime/include/cpp/jank/runtime/context.hpp
shantanu-sardesai: I'll add an overload then, the reason this type of a function I think is required is to prevent us from parsing the entire form in `clojure.core/read-string` when in reality it only returns the 1st form.

Especially, in Clojure the following will run just fine but won't in jank if we, unlike Clojure, continue reading the entire string.

```clj
user=> (read-string ":valid ::/invalid")
:valid
```

compiler+runtime/src/cpp/jank/runtime/context.cpp
shantanu-sardesai: Since during C++ codegen now there is this `var` constantly being set to `true` to ensure everything keeps working. It could have some costs which might need to be measured, no?

shantanu-sardesai: Unless you are suggesting that `TODO` needs more explanation in the comment.

jeaye: I was suggesting that the TODO was very unclear. But if the concern is just setting a var to true then no, this is not something to profile. :) That will not even be a drop in the bucket.

shantanu-sardesai:
I discovered that I introduced a regression in jank (in [this commit](https://github.com/jank-lang/jank/commit/35093255ed567b36256e4f00e2a5b6047ed96274)) wherein the following will cause the program to hang:

```jank
user=> #?(:jank 0 :clj 0v)           ; Hangs.

user=> #?(:jank 0 :clj)              ; Parses fine.

user=> #?(:jank 0 :default #{:k :k}) ; Doesn't return a duplicate set items error.
```

I've fixed it here in [this commit](https://github.com/jank-lang/jank/pull/673/changes/c55fad4c8767162a008d822cb085a8d9ea64ddfc). One of the tests I added fails with an unknown exception for some reason even though the case works fine in REPL:

```
$ jank repl

user=> #?(:default 0 :jank #{:k :k} :clj #{:k :k})
─ parse/duplicate-items-in-set ─────────────────────────────────────────────────────────────────────
error: Duplicate items in set literals are not allowed.

     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-GNkaZh
  1  │ #?(:default 0 :jank #{:k :k} :clj #{:k :k})
     │                          ^^ Duplicate item.
     │                       ^^ Original item.
```

and via `run`:

```
$ jank run test.jank

─ parse/duplicate-items-in-set ─────────────────────────────────────────────────────────────────────
error: Duplicate items in set literals are not allowed.

     │ test.jank                                                                                                                                                       ─────┼──────────────────────────────────────────────────────────────────────────────────────────────
  1  │ #?(:default 0 :jank #{:k :k} :clj #{:k :k})
     │                          ^^ Duplicate item.
     │                       ^^ Original item.
```

shantanu-sardesai:
This PR is also possibly blocked since currently jank doesn't have a runtime object to represent the reader conditional form itself. So the following which is legal in Clojure just can't be replicated in jank without either introducing the ability to hold a reader conditional form in the runtime or explicitly choosing to not support the `:preserve` mode of `:read-cond` in `read-string` `opts`.

```clj
user=> (read-string {:read-cond :preserve} "#?(:default 0 :jank 0 :clj 0)")
#?(:default 0 :jank 0 :clj 0)

user=> (type (read-string {:read-cond :preserve} "#?(:default 0 :jank 0 :clj 0)"))
clojure.lang.ReaderConditional
```

Currently the best I think we can replicate in jank is the following:

```jank
user=> (read-string {:read-cond :preserve} "#?(:default 0 :jank 0 :clj 0)")
(:default 0 :jank 0 :clj 0)

user=> (type (read-string {:read-cond :preserve} "#?(:default 0 :jank 0 :clj 0)"))
"persistent_list"
```

shantanu-sardesai:
There is a new limitation encountered due to the `:preserve` mode in `read-string` and jank's choice on reader suppression.

[Here is my summary of the problems with the complete context consolidated in a document](https://gist.github.com/shantanu-sardesai/b51510aef1d34cf8136da4a98aafbc93).

shantanu-sardesai:
I've updated the `read-string` implementation to adapt to our decision now. Things look great so far!

book/src/differences-from-clojure.md
```diff
@@ -17,6 +17,8 @@ Clojure.
 * No load operation for `data_readers.(cljc|jank)` at start-up to extend
   supported tags
 * jank allows using reader conditionals in `.jank` files as well
+* No automatic method to disable read by setting the `clojure.core/*read-eval*`
```
jeaye, line 20: We have a mitigation for this, which is `binding`. We should mention it here.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1115,6 +1116,14 @@ namespace jank::read::parse
     auto const start_token(token_current.latest.unwrap().expect_ok());
     ++token_current;
 
+    if(!allow_reader_conditional)
+    {
+      return error::parse_invalid_reader_conditional(
+        { start_token.start, latest_token.end },
+        "Conditional read is not allowed by default. Set the :read-cond reader option to either "
+        ":preserve or :allow. Found a nil instead.");
```
jeaye: I don't think `Found a nil instead.` is helpful here. I'm also not entirely sure if it's accurate, since we're working with a bool and not a runtime object.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1156,13 +1166,14 @@ namespace jank::read::parse
...
-      auto const feature_kw(dyn_cast<obj::keyword>(feature.ptr));
+      auto const feature_kw(expect_object<obj::keyword>(feature.ptr));
       /* We take the first match, checking for :jank first. If there are duplicates, it doesn't
        * matter. If :default comes first, we'll always take it. In short, order is important. This
        * matches Clojure's behavior. */
-      auto const is_supported_feature{
-        (equal(feature_kw, jank_keyword) || equal(feature_kw, default_keyword)) && result.is_none()
-      };
+      auto const is_supported_feature{ (equal(feature_kw, jank_keyword)
+                                        || equal(feature_kw, default_keyword)
```
jeaye: As a side note, we can use `==` on keywords, since they're interned.

compiler+runtime/src/cpp/jank/runtime/obj/reader_conditional.cpp
```diff
@@ -0,0 +1,93 @@
...
+  {
+    jtl::string_builder buff;
+    to_string_impl(form, splicing, buff);
+    return buff.release();
+  }
+
+  jtl::immutable_string reader_conditional::to_code_string() const
+  {
+    return to_string();
+  }
```
jeaye: No need to override this if it does the same as the base.

compiler+runtime/src/cpp/jank/runtime/obj/reader_conditional.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  bool reader_conditional::contains(object_ref const key) const
+  {
+    static auto const form_kw{ __rt_ctx->intern_keyword("form").expect_ok() };
+    static auto const splicing_qmark_kw{ __rt_ctx->intern_keyword("splicing?").expect_ok() };
+
+    return form_kw == key || splicing_qmark_kw == key;
+  }
+
+  /* TODO: Cache this. */
```
jeaye: This is not something we should care to cache.

compiler+runtime/src/cpp/jank/runtime/obj/reader_conditional.cpp
```diff
@@ -0,0 +1,93 @@
...
+  {
+    static auto const form_kw{ __rt_ctx->intern_keyword("form").expect_ok() };
+    static auto const splicing_qmark_kw{ __rt_ctx->intern_keyword("splicing?").expect_ok() };
+
+    if(runtime::equal(key, form_kw))
+    {
+      return form;
+    }
+
+    if(runtime::equal(key, splicing_qmark_kw))
```
jeaye: You use `equal` here and `==` below. Pick one. For keywords, `==` should suffice.

compiler+runtime/src/cpp/jank/runtime/obj/tagged_literal.cpp
```diff
@@ -52,15 +52,16 @@ namespace jank::runtime::obj
     return to_string();
   }
 
+  /* TODO: Cache this. */
```
jeaye: Please don't add TODOs about caching unless there is a strong performance potential gain from it.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -296,27 +301,160 @@ namespace jank::runtime
...
 
-  object_ref context::read_string(jtl::immutable_string const &code)
+  object_ref context::read_string(jtl::immutable_string const &code,
+                                  object_ref const reader_opts,
+                                  u64 const nth_form)
   {
     profile::timer const timer{ "rt read_string" };
+    static auto const unknown_kw{ __rt_ctx->intern_keyword("unknown").expect_ok() };
+    static auto const read_eval_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
+    auto const read_eval_enabled{ read_eval_var.is_nil()
```
jeaye: If the var doesn't exist, we're going to default to enabled. Is this what we want?

Firstly, if the var doesn't exist, something very wonky is going on. Secondly, why would someone remove that var? Perhaps the var was set to disable eval and the attacker wanted to get around that, without setting it to enabled directly? Thinking outside the box here, but we basically have a security flag and when the security flag is missing, we default to full access.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -296,27 +301,160 @@ namespace jank::runtime
...
+                                  || !equal(read_eval_var->deref(), unknown_kw) };
+
+    if(!read_eval_enabled)
+    {
+      throw std::runtime_error{ util::format(
+        "Reading is disallowed when `clojure.core/*read-eval*` is bound to `:unknown`.") };
```
jeaye: No need for `format` here.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -296,27 +301,160 @@ namespace jank::runtime
...
+    }
+
+
+    object_ref extended_features{};
+    bool in_preservation_mode{};
+    bool allow_reader_conditional{ false };
+
+    visit_map_like(
```
jeaye: There is no need for visiting here. `get` is an object behavior. Use that.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -296,27 +301,160 @@ namespace jank::runtime
...
+        if(features.is_some())
+        {
+          visit_set_like([](auto const _) {},
+                         [&]() {
+                           throw std::runtime_error{ util::format(
+                             "The :features reader option needs to be a set. Found a {} instead.",
+                             runtime::object_type_str(features->type)) };
+                         },
+                         features);
+
```
jeaye: Eeeeeek. This is an awful pattern. You are instantiating three different functions (2 set type fns + 1 error fn), doing a switch + dispatch, and the only thing you're doing is just checking object types. Just use `runtime::is_set`.

Object visiting is deprecated, with the new open object model, so we don't want to be adding more of it unless we have to. In this case, even if it weren't deprecated, we wouldn't want this.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -296,27 +301,160 @@ namespace jank::runtime
...
     /* When reading an arbitrary string, we don't want the last *current-file* to
      * be set as source file, so we need to bind it to nil. */
     binding_scope const preserve{ obj::persistent_hash_map::create_unique(
       std::make_pair(current_file_var, jank_nil())) };
-
     read::lex::processor l_prc{ code };
     read::parse::processor p_prc{ l_prc.begin(), l_prc.end() };
+    p_prc.extended_features = extended_features;
+    p_prc.in_preservation_mode = in_preservation_mode;
+    p_prc.allow_reader_conditional = allow_reader_conditional;
```
jeaye: This is not idiomatic C++. Initialization is done in the constructor, not after it.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -296,27 +301,160 @@ namespace jank::runtime
...
+    u64 count{};
+    static auto const eof_kw{ __rt_ctx->intern_keyword("", "eof").expect_ok() };
+    static auto const eof_throw_kw{ __rt_ctx->intern_keyword("", "eofthrow").expect_ok() };
+    auto const throw_on_eof{ equal(get(reader_opts, eof_kw, eof_throw_kw), eof_throw_kw) };
+    auto throw_eof{ throw_on_eof };
```
jeaye: This is somewhat confusing. Why not just have a bool for whether or not EOF was found? Then you can just say `if(found_eof && throw_on_eof)`, which makes more sense than this dance here.

compiler+runtime/src/cpp/jank/runtime/context.cpp
jeaye: This TODO still exists and should not.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -296,27 +301,160 @@ namespace jank::runtime
...
+  }
+
+  object_ref context::read_string(jtl::immutable_string const &code)
+  {
+    return read_string(code, obj::persistent_array_map::empty());
+  }
+
+  object_ref context::forcefully_read_string(jtl::immutable_string const &code)
+  {
+    auto const read_eval_enabled_var{ __rt_ctx->find_var("clojure.core", "*read-eval*") };
```
jeaye: This can be `static`, too.

compiler+runtime/src/cpp/jank/runtime/context.cpp
```diff
@@ -296,27 +301,160 @@ namespace jank::runtime
     return ok();
   }
 
-  object_ref context::read_string(jtl::immutable_string const &code)
+  object_ref context::read_string(jtl::immutable_string const &code,
+                                  object_ref const reader_opts,
+                                  u64 const nth_form)
```
jeaye, line 306: There's a lot of code in here now, with very few comments. Some of the code is quite dense. Please do a comment pass for your new code, to see where we can explain more.

jeaye (changes requested):
Nice work, Shantanu.

compiler+runtime/src/cpp/jank/read/parse.cpp
shantanu-sardesai: The `allow_reader_conditional` boolean is set using the provided reader option (`:read-cond`) so for the end user this message is more actionable. What do you think?

> I don't think Found a nil instead. is helpful here.

Removed.

compiler+runtime/src/cpp/jank/runtime/obj/reader_conditional.cpp
shantanu-sardesai: Neat, didn't check the default implementation.

compiler+runtime/src/cpp/jank/runtime/context.cpp
shantanu-sardesai: > Firstly, if the var doesn't exist, something very wonky is going on. Secondly, why would someone remove that var?

I think the check:

```C++
read_eval_var.is_nil()
```

can be removed, I'll remove it.

> Perhaps the var was set to disable eval and the attacker wanted to get around that, without setting it to enabled directly?

I'm assuming by someone you mean one of us jank compiler devs? Or can `var`s be removed in Clojure? Anyways, the check for it being present or not can/should be removed. I'll check once if everything works as expected.

> Thinking outside the box here, but we basically have a security flag and when the security flag is missing, we default to full access.

The `clojure.core/*read-eval*` `var` is used by the `read-string` function in the `clojure.core`. This `var` _disables_ the `read-string` function. By default in Clojure, `read-string` is enabled and this is because by default Clojure sets this `var` to allow for read. Besides it's unusual for a Clojure app to expose the `read` or it's siblings so I think default of full access should be fine. What do you think?

compiler+runtime/src/cpp/jank/runtime/context.cpp
shantanu-sardesai: The `visit_map_like` that I've added here is for the same reason that I added `visit_set_like` below. As you pointed out, I'll just use the runtime available functions (`is_map`) instead.

compiler+runtime/src/cpp/jank/runtime/context.cpp
jeaye: >I'm assuming by someone you mean one of us jank compiler devs?

I would not call a jank compiler dev an attacker. :) I'm talking about malicious users, library developers, etc.

> Or can vars be removed in Clojure? 

Yes, vars can definitely be removed. Entire namespaces can be. Anything that can be created can be destroyed.

compiler+runtime/src/cpp/jank/runtime/context.cpp
shantanu-sardesai: Not much to comment. Added in the few places I saw something worth adding.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -182,6 +182,24 @@ namespace jank::read::parse
   {
   }
 
+  processor::processor(lex::processor::iterator const &b,
+                       lex::processor::iterator const &e,
+                       object_ref const &extended_features,
+                       bool const &allow_reader_conditional,
```
jeaye, line 188: We should not take `bool` by `const &`, since it's a primitive.

compiler+runtime/src/cpp/jank/runtime/obj/reader_conditional.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <jank/runtime/core/equal.hpp>
+
+namespace jank::runtime::obj
+{
+  reader_conditional::reader_conditional()
+    : object{ obj_type, obj_behaviors }
+  {
+  }
+
+  reader_conditional::reader_conditional(persistent_list_ref f, boolean_ref s)
```
jeaye, line 14: These can be `const`.

PR #675, opened 2026-01-23
WIP: Use LLVM 22 from nixpkgs

kylc:
llvm-22.1.0-rc1 is now in nixpkgs and built into the nixpkgs cache. We can do away with building LLVM from source into the jank cache (this will stop happening automatically).

NOTE: This will fail until CppInterop (or the jank branch of CppInterop) is upgraded to account for some API changes in LLVM 22. I believe this is the breaking PR from LLVM: https://github.com/llvm/llvm-project/pull/175322

### DIFF
--- a/compiler+runtime/bin/jank/compiler+runtime/build+test.clj
+++ b/compiler+runtime/bin/jank/compiler+runtime/build+test.clj
@@ -64,10 +64,7 @@
       (util/with-elapsed-time duration
         (util/quiet-shell {:dir compiler+runtime-dir
                            :extra-env exports}
-                          (str "./bin/compile -v"
-                               ; This uses more memory in CI, so we limit parallelism.
-                               (when (= "Release" build-type)
-                                 " -j1")))
+                          "./bin/compile")
         (util/log-info-with-time duration "Compiled"))
 
       (util/quiet-shell {:dir compiler+runtime-dir


jeaye:
Nice! We're already using LLVM 22, just unreleased, so there should be no breakages. Let me look into getting this into CI to see if everything just works.

jeaye:
I see. There are actually errors in CppInterop, since there were some additional breakages. Unfortunately, the current LLVM API is no longer compatible with CppInterop, since the data we need, as far as I can tell, is now private. I've reached out to the Clang dev who made the changes (same one who leads CppInterop) to verify my understanding and ask for guidance. Aside from that, jank does build cleanly with this! :grin:

jeaye:
Ok, I hacked around this in CppInterop. Everything is working locally. I had Nix disabled in CI. I'll try adding it back into this PR now.

jeaye:
Ok, nice. Looks like the Nix build compiles + runs the functional test suite cleanly. We fail the health check when we try to build our incremental PCH, though. Also, locally, the health check fails to AOT compile due to:

```
/nix/store/cl88v2m1y5q3k6jlkq5jjf73nmfgl1px-binutils-2.44/bin/ld: cannot find Scrt1.o: No such file or directory
/nix/store/cl88v2m1y5q3k6jlkq5jjf73nmfgl1px-binutils-2.44/bin/ld: cannot find crti.o: No such file or directory
/nix/store/cl88v2m1y5q3k6jlkq5jjf73nmfgl1px-binutils-2.44/bin/ld: cannot find crtbeginS.o: No such file or directory
```

Ubuntu + macOS jobs are failing since they're not yet running on the LLVM 22 RC 1, so the CppInterOp changes cause compilation errors. I will fix that by updating our LLVM 22 branch +  build script, to trigger a rebuild of Clang/LLVM for them.

kylc:
I will investigate the AOT failure on Nix, because I see it locally as well.

kylc:
Main fix for CI is to pass include paths identified by `clang -v` during the jank build process via `-isystem` instead of `-I`. Without this, clang complains about GCC header files having invalid constexprs, conflicting noexcepts, etc. Not sure if this affects other platforms, but it seems like a sensible change.

Also included some flake.nix updates to make compiling + JIT + AOT in the dev shell work correctly, and some misc. cleanup.

Tested both the derivation and dev shell builds locally:
```
nix run . check-health

./bin/configure -GNinja -DCMAKE_BUILD_TYPE=Release -Djank_unity_build=ON \
  && ./bin/compile && jank check-health
```

Remaining failures on Ubuntu/Mac just look like some clang-tidy lints. Also looks like the CMake install script needs to be updated to refer to the correct RC version: https://github.com/jank-lang/jank/actions/runs/21324636076/job/61379699970?pr=675#step:9:2223

jeaye:
Thank you, Kyle!

PR #684, opened 2026-02-03
Fix SDKROOT on MacOS

jianlingzhong:


jeaye:
Thanks, Jianling. I have pushed my own version of this which is more graceful for when the user forgets to set the `SDKROOT`. Will you please verify this works for you?

jianlingzhong:
> Thanks, Jianling. I have pushed my own version of this which is more graceful for when the user forgets to set the `SDKROOT`. Will you please verify this works for you?

yep, that works.

PR #685, opened 2026-02-03
feat: C++ Codegen Direct Call

stmonty:
Adds the direct-call optimization for the C++ codegen, somewhat following a similar structure to the LLVM direct-call optimization.

I did some initial benchmarking. Running `ray.jank` with direct-call enabled seem to make some difference as you can see in the results, ~but compiling `clojure.core` with direct-call didn't seem to make much of a difference.~ This is no longer true as `clojure.core` will not be compiled with direct-call due to `eagerness`.

Testing with a more contrived example, with a benchmark that is almost entirely just derefing a var shows that there is significant speedup on repeated var-derefs.
```
Ray Benchmarks

C++ (Direct-Call)
|               ms/op |                op/s |    err% |          ins/op |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|---------------:|--------:|----------:|:----------
|               63.98 |               15.63 |    0.9% |  780,433,458.92 | 139,484,868.79 |    0.2% |     15.51 | `ray`

C++ (No Direct-Call)
|               ms/op |                op/s |    err% |          ins/op |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|---------------:|--------:|----------:|:----------
|               66.93 |               14.94 |    0.5% |  834,574,624.55 | 147,811,569.75 |    0.2% |     16.14 | `ray`


Micro Benchmark

C++ (Direct-Call)
|               ms/op |                op/s |    err% |          ins/op |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|---------------:|--------:|----------:|:----------
|                0.00 |      154,654,261.13 |    0.2% |           76.00 |          13.00 |    0.0% |      0.01 | `basic`

C++ (No Direct-Call)
|               ms/op |                op/s |    err% |          ins/op |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|---------------:|--------:|----------:|:----------
|                0.00 |        8,191,773.65 |    0.4% |        2,038.00 |         333.00 |    0.0% |      0.01 | `basic`

LLVM-IR (Direct-Call)
|               ns/op |                op/s |    err% |          ins/op |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|---------------:|--------:|----------:|:----------
|               15.29 |       65,398,408.30 |    0.0% |          238.00 |          35.00 |    0.0% |      0.19 | `basic`

LLVM-IR (No Direct-Call)
|               ns/op |                op/s |    err% |          ins/op |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|---------------:|--------:|----------:|:----------
|              147.86 |        6,763,029.82 |    0.4% |        2,608.00 |         442.00 |    0.0% |      1.79 | `basic`


```

### DIFF
--- a/compiler+runtime/test/bash/direct-call/pass-test
+++ b/compiler+runtime/test/bash/direct-call/pass-test
@@ -22,7 +22,11 @@
     out))
 
 (defn object-paths [module]
-  (fs/glob "." (str "target/**/jank_test/direct_call/" module ".o")))
+  (let [suffix (str "jank_test/direct_call/" module ".o")]
+    (concat (fs/glob "." (str "target/" suffix))
+            (fs/glob "." (str "target/**/" suffix))
+            (fs/glob "." (str "build/target/" suffix))
+            (fs/glob "." (str "build/target/**/" suffix)))))
 
 (use-fixtures
   :each


jeaye:
Thanks, Monty! Please look into the CI failures and make sure everything is passing.

Also, please gather benchmarks for both IR gen and C++ gen. We can then reason about them.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -92,6 +93,23 @@ namespace jank::codegen
...
+        return false;
+      }
+      auto const sep{ qualified_name.find('/') };
+      if(sep == jtl::immutable_string::npos)
+      {
+        return false;
+      }
+      auto const var_ns{ qualified_name.substr(0, sep) };
+      return !(var_ns == module || var_ns == "clojure.core");
+    }
```
jeaye, line 107: This whole parsing of the name is awkward. So is the storing of whether or not the var is dynamic, in `lifted_var`. If we just store the `var_ref` itself in there, we can access the namespace and whether it's dynamic without needing to duplicate state and manually parse strings.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -92,6 +93,23 @@ namespace jank::codegen
       return native_name;
     }
 
+    static bool should_defer_var_init(processor::lifted_var const &v,
```
jeaye: This fn warrants a comment explaining why some var inits are deferred.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -584,37 +608,48 @@ namespace jank::codegen
...
         }
-      case analyze::expression_position::tail:
-        util::format_to(body_buffer, "return ");
 
-        [[fallthrough]];
+        /* Only update the cached root when we opted into direct-call caching. */
+        if(var_native.is_some())
+        {
+          util::format_to(body_buffer,
+                          "if(var_root_{}.is_nil()) {{ var_root_{} = {}->get_root(); }}",
```
jeaye: Why do we have this check here? This is codegen for `def`. So, we're preventing future `def` usages from changing the root? Is that the reason?

I wouldn't expect that we're checking for `nil` at all, but we're doing it a lot in this PR.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1065,6 +1145,86 @@ namespace jank::codegen
...
+          arg_tmps.emplace_back(gen(arg_expr, fn_arity).unwrap());
+        }
+
+        util::format_to(body_buffer, "jank::runtime::object_ref {}{ };", ret_tmp.str(false));
+
+        /* Use direct-call for non-variadic roots; fall back when variadic or over max arity. */
+        if(expr->arg_exprs.size() <= runtime::max_params)
+        {
+          auto const arity_flags_tmp{ runtime::munge(__rt_ctx->unique_string("arity_flags")) };
+          util::format_to(body_buffer,
```
jeaye: By doing all of this work here, we've basically copied the internals of `dynamic_call` and put them into our codegen. That means we still need to do all of this work for every call. Can't we do this in the *generating* code and not the *generated* code, since the value of the var root is going to be the same as it is now?

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -2781,7 +2969,7 @@ namespace jank::codegen
         param_shadows_fn |= param->name == root_fn->name;
       }
 
-      util::format_to(body_buffer, ") const final {");
+      util::format_to(body_buffer, ") const override final {");
```
jeaye: `final` implies `override`.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -2899,8 +3087,25 @@ namespace jank::codegen
                         last->second.native_name);
       }
 
+      /* For direct-call eager init for local vars and clojure.core; other namespaces may not
+       * exist until jank_load has executed (require runs there), so defer those. */
```
jeaye: This comment doesn't make much sense to me. Can you explain in a different way?

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -2952,6 +3178,45 @@ namespace jank::codegen
...
+        /* Now that jank_load has executed (including require), init deferred vars. */
+        for(auto const &v : lifted_vars)
+        {
+          if(!detail::should_defer_var_init(v.second, v.first, module))
+          {
+            continue;
+          }
+          if(v.second.owned)
+          {
+            util::format_to(
```
jeaye, line 3258: These vars have not yet been initialized, so we cannot use `=` on them, since their current state is undefined. We need to use in-place `new`, just like the rest of the code is doing.

compiler+runtime/test/bash/direct-call/pass-test
```diff
@@ -0,0 +1,78 @@
...
+(def run-file (str module-path "/jank_test/direct_call/run.jank"))
+(def do-case-file (str module-path "/jank_test/direct_call/do_case.jank"))
+(def expected-output "42\n")
+(def do-case-output "Foo\n")
+(def base-cmd (str "jank --module-path " module-path " --direct-call --codegen cpp"))
+(def base-cmd-no-direct-call (str "jank --module-path " module-path " --codegen cpp"))
+
+(def redefs-run-file (str module-path "/jank_test/direct_call/redefs_run.jank"))
+(def redefs-direct-call-output "1\n1\n")
+(def redefs-control-output "1\n999\n")
```
jeaye: If you have data that's being used *once*, for one test, please put it in a `let` for that test.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1078,12 +1238,15 @@ namespace jank::codegen
       }
       if(!variadic)
       {
-        auto const &source_tmp(gen(expr->source_expr, fn_arity));
-        format_direct_call(source_tmp.unwrap().str(false),
-                           ret_tmp.str(true),
-                           expr->arg_exprs,
-                           fn_arity);
-        elided = true;
+        if(expr->arg_exprs.size() <= runtime::max_params)
```
jeaye: This `if` should be combined with the one above it.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1065,6 +1145,86 @@ namespace jank::codegen
...
+      {
+        auto const qualified_name{ ref->var->to_qualified_symbol()->to_string() };
+        auto const &var_native_name(
+          lift_var(lifted_vars, qualified_name, false, ref->var->dynamic.load()));
+        auto const source_tmp_root{ util::format("var_root_{}", var_native_name) };
+
+        if(target == compilation_target::module
+           && detail::should_defer_var_init(lifted_vars.at(qualified_name), qualified_name, module))
+        {
+          util::format_to(
```
jeaye, line 1173: This is at a call site. Why are we changing var roots here? This codegen is adding a LOT of additional checks and assignments, if we're doing this at call sites.

stmonty: Hopefully the comments make this more clear, but this check is only when we have to defer the init of the var-root for non-local vars, like those in other namespaces. It is a small penalty but should still be faster than  a `deref`.

stmonty:
@jeaye Made the changes from the comments. CI is failing because it can't find `boost/asio.hpp` from the most recent nrepl additions. I am not sure how to resolve this. I didn't find any details if this is a known issue.

jeaye:
> @jeaye Made the changes from the comments. CI is failing because it can't find `boost/asio.hpp` from the most recent nrepl additions. I am not sure how to resolve this. I didn't find any details if this is a known issue.

Merging in main will fix this issue.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -1298,6 +1298,9 @@ namespace jank::analyze
                                              meta_source(sym),
                                              latest_expansion(macro_expansions));
     }
+    auto const dynamic{ truthy(
+      get(qualified_sym->meta, __rt_ctx->intern_keyword("dynamic").expect_ok())) };
+    var_res.expect_ok()->dynamic.store(dynamic);
```
jeaye, line 1303: Why are we changing a var during analysis? This is very unconventional.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -517,18 +531,24 @@ namespace jank::codegen
   static jtl::immutable_string
   lift_var(native_unordered_map<jtl::immutable_string, processor::lifted_var> &lifted_vars,
            jtl::immutable_string const &qualified_name,
+           runtime::var_ref const var,
            bool const owned)
   {
     auto const existing{ lifted_vars.find(qualified_name) };
     if(existing != lifted_vars.end())
     {
+      existing->second.owned = existing->second.owned || owned;
```
jeaye, line 540: You can use `|=` for this.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -584,37 +607,52 @@ namespace jank::codegen
...
                           meta.unwrap(),
                           dynamic);
-          return var_tmp;
         }
         else
         {
-          util::format_to(body_buffer, "{}->bind_root({})->with_meta({});", var_tmp, val.str(true));
-          return var_tmp;
+          util::format_to(body_buffer,
+                          "{}->bind_root({})->with_meta(jank::runtime::jank_nil());",
```
jeaye, line 633: `->with_meta(jank_nil())` doesn't do anything, since meta defaults to nil. We can remove that entirely.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -626,6 +664,25 @@ namespace jank::codegen
...
+          auto const var_root{ util::format("var_root_{}", var_native.unwrap()) };
+          auto const var_root_flags{ util::format("var_root_arity_flags_{}", var_native.unwrap()) };
+          util::format_to(body_buffer,
+                          "if({}.is_nil()) {{ {} = {}->get_root(); {} = {}->get_arity_flags(); }}",
+                          var_root,
+                          var_root,
+                          var_tmp,
+                          var_root_flags,
+                          var_root);
+        }
```
jeaye, line 680: Why copy/paste these 12 lines? Can we not lift them out of the `switch` so that we just do them for both positions?

compiler+runtime/src/cpp/jank/codegen/processor.cpp
jeaye: Is this what Clojure JVM does? Have you taken a look at its generated byte code (or a decompiled Java representation)? Before we commit to anything, I'd want to make sure we have a good understanding of how Clojure does it.

This approach _will work_. But is it the best approach? I'm not certain.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -1065,6 +1155,123 @@ namespace jank::codegen
...
+        {
+          auto const root{ ref->var->get_root() };
+          auto const arity_flags{ root->get_arity_flags() };
+          bool const is_variadic{ (arity_flags & jank::runtime::mask_variadic_arity(0)) != 0 };
+
+          if(can_direct_call && !is_variadic)
+          {
+            util::format_to(body_buffer, "{} = {}->call(", ret_tmp.str(false), source_tmp_root);
+            bool need_comma{};
+            for(size_t i{}; i < arg_tmps.size(); ++i)
```
jeaye, line 1207: All of these loops here should be range-based.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -2633,13 +2840,26 @@ namespace jank::codegen
...
                         lifted_const,
                         v.second.native_name);
+        if(util::cli::opts.direct_call && !v.second.var->dynamic.load())
+        {
+          util::format_to(lifted_buffer,
+                          "jank::runtime::object_ref {} var_root_{};",
+                          lifted_root_const,
+                          v.second.native_name);
+          util::format_to(lifted_buffer,
+                          "jank::runtime::callable_arity_flags var_root_arity_flags_{};",
```
jeaye, line 2860: As discussed, these need to be `thread_local`. Otherwise, we have undefined behavior when multiple threads are used.

compiler+runtime/src/cpp/jank/codegen/processor.cpp
```diff
@@ -2899,8 +3139,29 @@ namespace jank::codegen
...
+        if(util::cli::opts.direct_call && detail::should_defer_var_init(v.second, module))
+        {
+          util::format_to(footer_buffer,
+                          "new (&{}::{}) jank::runtime::var_ref{};",
+                          ns,
+                          v.second.native_name);
+          if(util::cli::opts.direct_call && !v.second.var->dynamic.load())
+          {
+            util::format_to(footer_buffer,
+                            "new (&{}::var_root_{}) jank::runtime::object_ref{};",
```
jeaye, line 3155: This can lead to premature GC of the roots. If direct calls are enabled and a var is replaced, this var_root may be the only reference to the old root, but it's not registered with the GC. So we'll need to make sure these are uncollectable.

jeaye:
Closing, since you're taking a different approach.

PR #688, opened 2026-02-07
Rework tagged literals

### LINKED ISSUE #639: Rework tagged literals

shantanu-sardesai:
An alternative implementation for #650.

- [x] Graceful parsing.
  - [x] Meta-hint.
  - [x] Reader function macro.

---

Resolves #639.

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -431,7 +431,7 @@ namespace jank::read::parse
     }
 
     native_unordered_map<runtime::object_ref, object_source_info> parsed_keys{};
-    auto const &items = ret.expect_ok();
+    auto const &items{ ret.expect_ok() };
     auto const build_map([&](auto &map) -> jtl::result<void, error_ref> {
       using T = std::remove_reference_t<decltype(map)>;
 


shantanu-sardesai:
Ready for the first round. Although there is a potential for code reuse for verifying if the error is of an insuppressible kind or not, but I'm not sure how to proceed on that.

compiler+runtime/include/cpp/jank/error.hpp
```diff
@@ -17,7 +17,10 @@ namespace jank::error
   /* TODO: Rename internal failures to have correct prefix. i.e. lex_internal_failure. */
   enum class kind : u8
   {
-    lex_unexpected_eof,
+    /* Represents errors that, when encountered during parsing, prevent the parser
+     * from gracefully consuming the rest of the syntactic form. */
+    suppressed_reader_error_min,
```
jeaye, line 22: Let's use a complete sentence here. We're missing the noun, which will help us be clear that we're talking about supressed reader errors here. This comment sits on top of a whole enum definition.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -14,6 +14,7 @@
 #include <jank/runtime/sequence_range.hpp>
 #include <jank/util/scope_exit.hpp>
 #include <jank/util/fmt.hpp>
+#include <jank/util/fmt/print.hpp>
```
jeaye: This isn't needed.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -315,124 +316,137 @@ namespace jank::read::parse
     }
   }
 
-  processor::object_result processor::parse_list()
+  jtl::result<native_vector<object_source_info>, error_ref>
+  processor::gracefully_parse(lex::token_kind const &upto,
+                              error_ref (*unterminated_form_error)(read::source const &))
```
jeaye, line 320: It'd be better to use a `jtl::ref<error_ref (*)(read::source const&)>` so that we don't need to worry about stray null derefs.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -315,124 +316,137 @@ namespace jank::read::parse
     }
   }
 
-  processor::object_result processor::parse_list()
+  jtl::result<native_vector<object_source_info>, error_ref>
+  processor::gracefully_parse(lex::token_kind const &upto,
```
jeaye, line 319: This fn warrants a comment explaining why we need to be explicitly graceful here.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -315,124 +316,137 @@ namespace jank::read::parse
...
-        return err(it.latest.unwrap().expect_err());
-      }
-      ret.push_back(it.latest.unwrap().expect_ok().unwrap().ptr);
+      return ret.expect_err();
     }
-    if(expected_closer.is_some())
+
+    runtime::detail::native_transient_vector items;
+
+    for(auto &it : ret.expect_ok())
```
jeaye: There is no need for a mutable reference here. `auto const &it` is what we want.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -315,124 +316,137 @@ namespace jank::read::parse
...
+                                     error::parse_unterminated_list) };
+
+    if(ret.is_err())
+    {
+      return ret.expect_err();
+    }
+
+    runtime::detail::native_transient_vector items;
+
+    for(auto &it : ret.expect_ok())
```
jeaye: No need for a mutable `it` here.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -315,124 +316,137 @@ namespace jank::read::parse
...
     {
-      return error::parse_unterminated_map(start_token.start);
+      return ret.expect_err();
     }
 
-    expected_closer = prev_expected_closer;
-
     native_unordered_map<runtime::object_ref, object_source_info> parsed_keys{};
-
+    auto items{ ret.expect_ok() };
```
jeaye: No need to copy the whole vector here. Also, `items` is not mutable. Use `auto const &` instead.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -689,27 +726,20 @@ namespace jank::read::parse
...
-    for(auto it(begin()); it != end(); ++it)
-    {
-      if(it.latest.unwrap().is_err())
-      {
-        return err(it.latest.unwrap().expect_err());
-      }
+    runtime::detail::native_transient_hash_set set;
 
-      auto const item(it.latest.unwrap().expect_ok().unwrap());
+    for(auto &it : ret.expect_ok())
```
jeaye: No need for mutable `it` here. **Use const.**

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1040,25 +1041,109 @@ namespace jank::read::parse
...
+
+    auto const form_end(form_result.expect_ok().unwrap().end);
+    auto const form(form_result.expect_ok().unwrap().ptr);
+    auto const data_readers_result{ __rt_ctx->find_var("clojure.core", "*data-readers*") };
+
+    if(data_readers_result.is_some())
+    {
+      auto const data_readers{ data_readers_result->deref() };
+      auto const data_reader_result{ visit_map_like(
+        [](auto const typed_o, obj::symbol_ref const sym)
```
jeaye: Visiting must always add `typed_` to the original variable name, which here was `data_readers`.

compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -1040,25 +1041,109 @@ namespace jank::read::parse
...
+        {
+          return object_source_info{ default_data_reader_fn->call(sym, form),
+                                     form_token,
+                                     form_end };
+        }
+        else
+        {
+          return error::parse_invalid_reader_symbolic_value(
+            util::format(
+              "The 'clojure.core/*default-data-reader-fn*' var must be a function toking 2 "
```
jeaye: `toking`

compiler+runtime/test/cpp/jank/read/parse.cpp
```diff
@@ -1376,6 +1472,31 @@ namespace jank::read::parse
...
+          SUBCASE("Ignore unmatched conditionals")
+          {
+            lex::processor lp{ "[#?@(:cljs #js {} :jank [42])]" };
+            processor p{ lp.begin(), lp.end() };
+            auto const r(p.next());
+            CHECK(equal(r.expect_ok().unwrap().ptr,
+                        make_box<obj::persistent_vector>(std::in_place, make_box(42))));
+          }
+
+          SUBCASE("Throw lex errors in unsupported reader conditionals")
```
jeaye: Please don't use the term "throw" when not talking about exceptions.

compiler+runtime/test/jank/reader-macro/tag/fail-non-fn-data-reader.jank
```diff
@@ -0,0 +1,2 @@
+(binding [*data-readers* {'foo 42}]
```
jeaye: 1. All directories here must be singular (i.e. `tags` => `tag`)
2. These are not compiler feature functional tests; these are library tests for `read-string`; they belong in clojure-test-suite
3. The `#foo 42` test is the only one which belongs here, but it doesn't need its own directory; just put it in `test/jank/reader-macro/fail-unsupported-tag.jank`

jeaye (changes requested):
Nice work, Shantanu!

compiler+runtime/src/cpp/jank/read/parse.cpp
shantanu-sardesai: Another one of the C++ things that I keep forgetting 🙂 . Would this be better:

```C++
auto const &items(ret.expect_ok());

// OR

auto const &items = ret.expect_ok();
```

compiler+runtime/test/jank/reader-macro/tag/fail-non-fn-data-reader.jank
shantanu-sardesai: Hmm... I also grappled with this one.

I've chosen to use `read-string` here because the following tests would end up modifying the global testing context:

```jank
(alter-var-root #'*data-readers* (fn [_] {'foo identity}))

#foo 42

(alter-var-root #'*data-readers* (fn [_] {}))
```

And since this modified the parser's capability as well, it felt like a grey area. I agree that this also needs to be added to the Clojure test suite since jank also deviates from Clojure JVM in what we support... Gah, I suppose I can just move this to the test suite after all...

shantanu-sardesai: I'll give it some more thought.

shantanu-sardesai:
Weird that the tests in `test/jank/syntax-quote/unquote-splice/pass-values.jank` fail. I can't replicate the error in my local, it runs just fine.

compiler+runtime/test/jank/reader-macro/tag/fail-non-fn-data-reader.jank
shantanu-sardesai: Also here, I'm testing the parsers ability to read in tags, and not the entire `read-string` function, if you still think it doesn't belong here then I'll remove them. Besides that I've renamed the test directory from `tags` -> `tag`.

compiler+runtime/src/cpp/jank/read/parse.cpp
jeaye: Never the latter. The former is acceptable. This form is preferred to both.

```cpp
auto const &items{ ret.expect_ok() };
```

compiler+runtime/test/jank/reader-macro/tag/fail-non-fn-data-reader.jank
jeaye: I don't think this is a gray area. These tests don't belong in this suite. This suite shouldn't use any `clojure.core` vars other than `assert`, without very good reason. The whole point of these tests you've written is to use `clojure.core/read-string`, which doesn't fit into this suite.

If you're only using `read-string` to test the parser, perhaps these should be parse tests where you set the var instead. But since these tests would also work with other dialects, we should just do this in clojure-test-suite and have everyone benefit from it.

jeaye:
> Weird that the tests in `test/jank/syntax-quote/unquote-splice/pass-values.jank` fail. I can't replicate the error in my local, it runs just fine.

This would imply we have some undefined behavior (UB) in your code, a GC issue, or CI is running out of memory, in order of likeliness.

compiler+runtime/src/cpp/jank/read/parse.cpp
jeaye: `=` is for mutation, not assignment.

compiler+runtime/test/jank/reader-macro/tag/fail-non-fn-data-reader.jank
shantanu-sardesai: Alright, mentioning [the ticket](https://github.com/jank-lang/clojure-test-suite/issues/433) for `read-string` here so that they can draw some inspiration from these tests.

shantanu-sardesai:
I suspect the UB was due to this [commit](https://github.com/jank-lang/jank/pull/688/changes/472a00e1f9aabb04516153105b1dded7933736b5), let's see 🤞.

compiler+runtime/src/cpp/jank/read/parse.cpp
shantanu-sardesai: Since `jtl::ref`s don't currently support function types, skipping this.

PR #690, opened 2026-02-09
Fix `to_string` implementations of runtime function objects

shantanu-sardesai:
```jank
user=> (defn foo [])
#'user/foo

user=> [foo]
[#object [user/foo deferred_cpp_function 0x10fa3f210]]

user=> {foo 0}
{#object [user/foo deferred_cpp_function 0x10fa3f210] 0}

user=> {0 foo}
{0 #object [user/foo deferred_cpp_function 0x10fa3f210]}

user=> #{foo}
#{#object [user/foo deferred_cpp_function 0x10fa3f210]}
```

---

Resolves https://github.com/jank-lang/jank/discussions/689.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/obj/deferred_cpp_function.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/deferred_cpp_function.cpp
@@ -27,7 +27,7 @@ namespace jank::runtime::obj
 
   void deferred_cpp_function::to_string(jtl::string_builder &buff) const
   {
-    auto const name(get(meta, __rt_ctx->intern_keyword("name").expect_ok()));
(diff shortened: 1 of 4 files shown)


shantanu-sardesai:
@jeaye

If my LSP is running correctly then all the unqualified `get` have been removed with this PR. I just tried to go to all references of these [2](https://github.com/shantanu-sardesai/jank/blob/24ab221315d38aef2ec28770710fc92fa5aec7ec/compiler+runtime/src/cpp/jank/runtime/object.cpp#L180) [`get`s](https://github.com/shantanu-sardesai/jank/blob/24ab221315d38aef2ec28770710fc92fa5aec7ec/compiler+runtime/src/cpp/jank/runtime/object.cpp#L185) and besides it's own default implementation and overrides these 4 were seemingly the only places of referrences.

jeaye:
Thanks!

PR #697, opened 2026-02-14
Consider cpp as compiled module instead of source

Samy-33:


compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -726,94 +726,118 @@ namespace jank::runtime::module
 
     if(found.is_none())
     {
-      /* TODO: If it contains -, suggest using _. Very common issue. */
```
jeaye, line 729: Why was this removed? This is still a valid TODO.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -726,94 +726,118 @@ namespace jank::runtime::module
...
     if(ori == origin::source)
     {
-      if(entry.jank.is_some())
+      return find_result{ entry, source_type };
+    }
+
+    jtl::option<file_entry> compiled_entry;
+    module_type compiled_type{};
+
+    auto const get_mod_time = [](file_entry const &e) {
```
jeaye: This lambda doesn't capture any locals, so let's just lift it to a standalone function instead.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -726,94 +726,118 @@ namespace jank::runtime::module
...
+    /* Ignoring object files from the archives here for security and portability reasons.
+     *
+     * Security:
+     * A dependency can include a binary version of a module that doesn't belong
+     * to it.
+     *
+     * Portability:
+     * Unlike class files, object files are tied to the OS, architecture, C++ stdlib etc,
+     * making it hard to share them. */
+    bool const has_o
```
jeaye: Please update this whole PR to use direct initialization.

`=` is for mutation, not initialization.

```cpp
auto const has_o{ entry.o.is_some() && entry.o.unwrap().exists() && entry.o.unwrap().archive_path.is_none() };
```

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -726,94 +726,118 @@ namespace jank::runtime::module
...
+      = entry.o.is_some() && entry.o.unwrap().exists() && entry.o.unwrap().archive_path.is_none();
+    bool const has_cpp = entry.cpp.is_some() && entry.cpp.unwrap().exists();
+
+    using target_t = util::cli::compilation_target;
```
jeaye: The `_t` suffix is reserved by POSIX for type names. If you want a shorthand here, just do:

```cpp
using util::cli::compilation_target;

compilation_target::cpp
```

I personally wouldn't bother with it, since you use `target_t` only twice.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -726,94 +726,118 @@ namespace jank::runtime::module
...
+    auto const target = util::cli::opts.output_target;
+
+    /* During compilation jank will only consider the cache of chosen type.
```
jeaye: This first sentence is not grammatically correct. More importantly, it's not easily understood.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -726,94 +726,118 @@ namespace jank::runtime::module
...
+      else if(has_cpp)
       {
-        return find_result{ entry, module_type::jank };
+        compiled_entry = entry.cpp.unwrap();
+        compiled_type = module_type::cpp;
       }
-      else if(entry.cljc.is_some())
+    }
+
+    /* Freshness check vs source! */
```
jeaye: All of the logic leading up to this reads pretty cleanly. This comment, however, doesn't add anything. I'd write something like:

> We now have the latest compiled entry. If it's up to date with the source, we can use it. Otherwise, we'll use the source.

compiler+runtime/test/bash/module/compiled-modules/pass-test
```diff
@@ -44,28 +67,118 @@
...
+(deftest compiled-candidate-selection-by-freshness
+  (testing "Runtime ignores flags and selects the freshest binary"
+    ;; Define times relative to `now` to ensure validity
+    (let [now         (System/currentTimeMillis)
+          base-time   (- now 30000)
+          source-time base-time
+          stale-time  (- now 20000)
+          fresh-time  (- now 10000)]
+
+      ;; Compile Object with unique output
```
jeaye: These are great tests. I find it odd that we're capitalizing `Object` in a couple of places (but not all of them). I'd just use `object` here.

jeaye (changes requested):
Nice work, Saket!

jeaye:
@Samy-33 Looks like some e2e tests broke and there are some clang-tidy issues. Once CI is green, this should be good to go!

jeaye:
Nice work!

PR #700, opened 2026-02-22
Fix static member access mismatch between analysis and codegen

jianlingzhong:
# Before fix: 18 LLVM IR failures:
```
Detailed Failure Reports (18 Tests)
● test/jank/cpp/closure/pass-convertible.jank (llvm-ir, 0ms)
  ↳ Assertion Failed
● test/jank/cpp/global-value/pass-static.jank (llvm-ir, 11ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/global-value/pass-static-non-copyable.jank (llvm-ir, 10ms)
  ↳ Crashed (Abort - Sig 6)
● test/jank/cpp/member/pass-static.jank (llvm-ir, 14ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/auto-box/pass-closure.jank (llvm-ir, 0ms)
  ↳ Assertion Failed
● test/jank/cpp/auto-box/pass-reference-input.jank (llvm-ir, 44ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/literal-value/pass-string-literal.jank (llvm-ir, 55ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/preprocessor/pass-normal.jank (llvm-ir, 63ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/preprocessor/pass-unqualified.jank (llvm-ir, 63ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/operator/aget/pass-binary-return-reference.jank (llvm-ir, 52ms)
  ↳ Assertion Failed
● test/jank/cpp/operator/aget/pass-binary-compatible-types.jank (llvm-ir, 54ms)
  ↳ Assertion Failed
● test/jank/cpp/operator/plusplus/pass-unary-pointer.jank (llvm-ir, 33ms)
  ↳ Crashed (Abort - Sig 6)
● test/jank/cpp/operator/minusminus/pass-unary-pointer.jank (llvm-ir, 36ms)
  ↳ Crashed (Abort - Sig 6)
● test/jank/cpp/loop/pass-nested-let.jank (llvm-ir, 55ms)
  ↳ Assertion Failed
● test/jank/cpp/loop/pass-native-binding-custom-type.jank (llvm-ir, 5048ms)
  ↳ Timeout (>5.0s limit)
● test/jank/cpp/loop/pass-native-binding-trait-convertible.jank (llvm-ir, 5019ms)
  ↳ Timeout (>5.0s limit)
● test/jank/cpp/constructor/complex/pass-aggregate-multiple-conversions.jank (llvm-ir, 162ms)
  ↳ Result is not :success: nil
● test/jank/cpp/constructor/complex/pass-aggregate-object.jank (llvm-ir, 0ms)
  ↳ Exception thrown: Unable to find conversion function match for policy 'from_object' with conversion type 
'jank::cpp::constructor::complex::pass_aggregate_object::foo' and input type 
'jank::runtime::oref<jank::runtime::object>'.
```

# After fix: 13 LLVM IR failures:
```
Detailed Failure Reports (13 Tests)
● test/jank/cpp/closure/pass-convertible.jank (llvm-ir, 0ms)
  ↳ Assertion Failed
● test/jank/cpp/call/global/pass-pointer-to-array.jank (llvm-ir, 25ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/auto-box/pass-closure.jank (llvm-ir, 0ms)
  ↳ Assertion Failed
● test/jank/cpp/literal-value/pass-string-literal.jank (llvm-ir, 50ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/preprocessor/pass-normal.jank (llvm-ir, 61ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/preprocessor/pass-unqualified.jank (llvm-ir, 61ms)
  ↳ Crashed (Segfault - Sig 11)
● test/jank/cpp/operator/plusplus/pass-unary-pointer.jank (llvm-ir, 32ms)
  ↳ Crashed (Abort - Sig 6)
● test/jank/cpp/operator/minusminus/pass-unary-pointer.jank (llvm-ir, 38ms)
  ↳ Crashed (Abort - Sig 6)
● test/jank/cpp/loop/pass-nested-let.jank (llvm-ir, 53ms)
  ↳ Assertion Failed
● test/jank/cpp/loop/pass-native-binding-custom-type.jank (llvm-ir, 5029ms)
  ↳ Timeout (>5.0s limit)
● test/jank/cpp/loop/pass-native-binding-trait-convertible.jank (llvm-ir, 5042ms)
  ↳ Timeout (>5.0s limit)
● test/jank/cpp/constructor/complex/pass-aggregate-multiple-conversions.jank (llvm-ir, 110ms)
  ↳ Result is not :success: nil
● test/jank/cpp/constructor/complex/pass-aggregate-object.jank (llvm-ir, 0ms)
  ↳ Exception thrown: Unable to find conversion function match for policy 'from_object' with conversion type 
'jank::cpp::constructor::complex::pass_aggregate_object::foo' and input type 
'jank::runtime::oref<jank::runtime::object>'.
```

# Fixed:

* test/jank/cpp/global-value/pass-static-non-copyable.jank
* test/jank/cpp/member/pass-static.jank
* test/jank/cpp/auto-box/pass-reference-input.jank
* test/jank/cpp/operator/aget/pass-binary-return-reference.jank
* test/jank/cpp/operator/aget/pass-binary-compatible-types.jank

## AI Disclosure
- [x] This PR was assisted by Google Gemini.
- [x] This PR was manually reviewed for accuracy.

### DIFF
--- a/compiler+runtime/src/cpp/jank/analyze/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/analyze/processor.cpp
@@ -3465,7 +3465,10 @@ namespace jank::analyze
     if(Cpp::IsVariable(scope))
     {
       vk = expr::cpp_value::value_kind::variable;
-      if(!Cpp::IsPointerType(type))
+      bool const is_copyable{ Cpp::IsConstructible(type, type) };
+      bool const is_static_datamember{ Cpp::IsStaticDatamember(scope)
+                                       && Cpp::IsClass(Cpp::GetParentScope(scope)) };
+      if(!Cpp::IsPointerType(type) && (!is_static_datamember || !is_copyable))
       {
         type = Cpp::GetLValueReferenceType(type);
       }
@@ -4645,7 +4648,15 @@ namespace jank::analyze
       }
 
       val->val_kind = expr::cpp_value::value_kind::variable;
-      val->type = Cpp::GetLValueReferenceType(Cpp::GetTypeFromScope(member_scope));
+      auto const type{ Cpp::GetTypeFromScope(member_scope) };
+      if(!Cpp::IsPointerType(type) && !Cpp::IsConstructible(type, type))
+      {
+        val->type = Cpp::GetLValueReferenceType(type);
+      }
+      else
+      {
+        val->type = type;
+      }
       val->scope = member_scope;
       return val;
     }
--- a/compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
+++ b/compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
@@ -2680,7 +2680,8 @@ namespace jank::codegen
     /* For static variables that are non-copyable, we can't use MakeAotCallable (which copies).
      * Instead, create a wrapper function that returns a pointer to the variable.
      * Check if the type is copyable by seeing if it can be constructed from itself.  */
-    bool const is_copyable{ Cpp::IsConstructible(expr->type, expr->type) };
+    auto const underlying_type{ Cpp::GetNonReferenceType(expr->type) };
+    bool const is_copyable{ Cpp::IsConstructible(underlying_type, underlying_type) };
     if(expr->val_kind == expr::cpp_value::value_kind::variable && Cpp::IsVariable(expr->scope)
        && !is_copyable)
     {


compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -2641,7 +2680,8 @@ namespace jank::codegen
     /* For static variables that are non-copyable, we can't use MakeAotCallable (which copies).
      * Instead, create a wrapper function that returns a pointer to the variable.
      * Check if the type is copyable by seeing if it can be constructed from itself.  */
-    bool const is_copyable{ Cpp::IsConstructible(expr->type, expr->type) };
+    auto const underlying_type{ Cpp::GetNonReferenceType(expr->type) };
```
jeaye, line 2683: This isn't the underlying type, it's a non-reference type. Underlying type is a separate thing.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -4632,7 +4648,15 @@ namespace jank::analyze
       }
 
       val->val_kind = expr::cpp_value::value_kind::variable;
-      val->type = Cpp::GetLValueReferenceType(Cpp::GetTypeFromScope(member_scope));
+      auto const type{ Cpp::GetTypeFromScope(member_scope) };
+      if(!Cpp::IsPointerType(type) && !Cpp::IsConstructible(type, type))
```
jeaye, line 4652: This says that if we have a static member which is copyable, we'll copy it. That is not correct behavior. It means we could never mutate a static member, since we'll only ever get a copy of it. This will also impact performance.

jeaye (changes requested):
Thanks, Jianling. I see three changes here.

1. Copyable check for variables, which defaults us to copy them
2. Copying check for static members, which defaults us to copy them
3. Reference stripping before copyable check during codegen

The first two are semantic changes which we don't want. They will also negatively impact perf. The last change is viable, but the variable name used is incorrect.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
jianlingzhong: updated name. the reference itself is always copyable. i believe the Cpp::MakeAotCallable is looking at the actual type, so we need to check the non referenced type instead.

compiler+runtime/src/cpp/jank/analyze/processor.cpp
jianlingzhong: in codegen:

```cpp
auto const callable{
  Cpp::IsFunctionPointerType(expr->type)
    /* We pass the type and the scope in here so that unresolved template
                        * scopes can be turned into the correct specialization which matches
                        * the type we have. */
    ? Cpp::MakeFunctionValueAotCallable(expr->scope, expr->type, unique_munged_string())
    : Cpp::MakeAotCallable(expr->scope, unique_munged_string())
};
```

iiuc, if something is copyable, we *are* copying it, isn't it?

jianlingzhong: Without this change, `test/jank/cpp/member/pass-static.jank` will fail.

jianlingzhong: we should change the `if` logic to handle static members differently if we want to avoid a copy:

```cpp
    if(expr->val_kind == expr::cpp_value::value_kind::variable && Cpp::IsVariable(expr->scope)
       && !is_copyable)
    {
      auto const fn_type(llvm::FunctionType::get(ctx->builder->getPtrTy(), false));
      auto const fn{ ensure_static_wrapper(expr->scope) };
...
    }

    auto const callable{
      Cpp::IsFunctionPointerType(expr->type)
        /* We pass the type and the scope in here so that unresolved template
         * scopes can be turned into the correct specialization which matches
         * the type we have. */
        ? Cpp::MakeFunctionValueAotCallable(expr->scope, expr->type, unique_munged_string())
        : Cpp::MakeAotCallable(expr->scope, unique_munged_string())
    };
```

jeaye: There is one special case for static members, which is a static integer that's initialized inline. We cannot reference those, since they have no address. Every other static must be referenced. I think we want a new test which grabs a static variable and mutates it. That way we can be sure we're not copying.

Something like:

```clojure
(let* [v cpp/jank.test.something.static_val]
  (cpp/= v #cpp 5)
  (if (cpp/== #cpp 5 cpp/jank.test.something.static_val)
    :success))
```

compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -3463,7 +3465,10 @@ namespace jank::analyze
     if(Cpp::IsVariable(scope))
     {
       vk = expr::cpp_value::value_kind::variable;
-      if(!Cpp::IsPointerType(type))
+      bool const is_copyable{ Cpp::IsConstructible(type, type) };
+      bool const is_static_datamember{ Cpp::IsStaticDatamember(scope)
+                                       && Cpp::IsClass(Cpp::GetParentScope(scope)) };
```
jeaye, line 3470: Following the discussion elsewhere, I don't think this is right. Perhaps it is, but I'll need some explanation for as to why. Static members, outside of the one special case of inline-defined integrals, should be referenced by default.

If this is failing with IR gen, it's more likely that we have an IR gen bug that doesn't handle reference properly here than it is that we need to change analysis.

jianlingzhong: when we pass the static member to `MakeAotCallable` below, i believe `MakeAotCallable` is trying to return a value and therefore making copy:

```cpp
    auto const callable{
      Cpp::IsFunctionPointerType(expr->type)
        /* We pass the type and the scope in here so that unresolved template
         * scopes can be turned into the correct specialization which matches
         * the type we have. */
        ? Cpp::MakeFunctionValueAotCallable(expr->scope, expr->type, unique_munged_string())
        : Cpp::MakeAotCallable(expr->scope, unique_munged_string())
    };
```

jeaye: Ok, the bug is there then. I don't think this is an analysis bug; it's a codegen bug.

jianlingzhong: shall we patch the CppInterop code then?

jeaye: > shall we patch the CppInterop code then?

If the CppInterop code is forcing values instead of refs, yes.

jianlingzhong: Added a print in the MakeAotCallable [here](https://github.com/jank-lang/CppInterOp/blob/1b632319e4a3b2553af755c9b929f9f67a6e1df9/lib/Interpreter/CppInterOp.cpp#L4325) to print the actual code it generated. 

This is before the changes I made (so it still crashes):

```
❯ build/jank run --codegen llvm-ir test/jank/cpp/global-value/pass-static.jank  
generated wrapper code: __attribute__((used)) __attribute__((annotate("__cling__ptrcheck(off)")))
extern "C" [[gnu::always_inline]]
inline void clojure_core_G__79051(void*, int, void**, [[gnu::nonnull]] void* ret)
{
   new (ret) (const int)(jank::cpp::global_value::pass_static::foo::boop);
}

[1]    6893 segmentation fault  build/jank run --codegen llvm-ir test/jank/cpp/global-value/pass-static.jank



❯ build/jank run --codegen llvm-ir test/jank/cpp/auto-box/pass-reference-input.jank        
generated wrapper code: __attribute__((used)) __attribute__((annotate("__cling__ptrcheck(off)")))
extern "C" [[gnu::always_inline]]
inline void clojure_core_G__79052(void*, int, void**, [[gnu::nonnull]] void* ret)
{
   new (ret) (const unsigned long)(std::basic_string<char>::npos);
}

[1]    7431 segmentation fault  build/jank run --codegen llvm-ir 
```

jeaye: No need to hack in a print. Just change this line in `main.cpp`:

```cpp
    Cpp::EnableDebugOutput(false);
```

Yes, it's copying there, but is it copying because the input type is not an lvalue reference?

jianlingzhong: Current code makes everything a reference, including static members. And references are always copyable so they do not pass this if:

```cpp
if(expr->val_kind == expr::cpp_value::value_kind::variable && Cpp::IsVariable(expr->scope) && !is_copyable)
```

and therefore they get pass to `MakeAotCallable` and end up being copied.

jeaye:
Closing this, since it's out of date and (I think) no longer needed. Let me know if that's not the case.

PR #702, opened 2026-02-24
Switch from struct to fn cppgen; drastically improve LLVM load

jeaye:
## Overview
Previously, given some Clojure code like this:

```clojure
(defn unreduced
  "If x is reduced?, returns (deref x), else returns x"
  [x]
  (if (reduced? x)
    (deref x)
    x))
```

We would generate some C++ like this:

```cpp
namespace clojure::core
{
  struct clojure_core_unreduced_42519 : jank::runtime::obj::jit_function
  {
    jank::runtime::var_ref const clojure_core_deref_42521;
    jank::runtime::var_ref const clojure_core_reduced_QMARK__42520;

    clojure_core_unreduced_42519()
      : jank::runtime::obj::jit_function{ jank::runtime::__rt_ctx->read_string(
          "{:name \"clojure.core/unreduced\"}") }
      , clojure_core_deref_42521{ jank::runtime::__rt_ctx->intern_var("clojure.core", "deref")
                                    .expect_ok() }
      , clojure_core_reduced_QMARK__42520{
        jank::runtime::__rt_ctx->intern_var("clojure.core", "reduced?").expect_ok()
      }
    {
    }

    jank::runtime::object_ref call(jank::runtime::object_ref const x) final
    {
      using namespace jank;
      jank::runtime::object_ref const unreduced{ this };
      jank::runtime::object_ref clojure_core_if_53727{};
      auto const clojure_core_call_53728(
        jank::runtime::dynamic_call(clojure_core_reduced_QMARK__42520->deref(), x));
      if(jank::runtime::truthy(clojure_core_call_53728))
      {
        auto const clojure_core_call_53729(
          jank::runtime::dynamic_call(clojure_core_deref_42521->deref(), x));
        return clojure_core_call_53729;
      }
      else
      {
        return x;
      }
    }
  };
}
```

We would also generate something like this to instantiate it:

```cpp
jank::runtime::make_box<clojure_core_unreduced_42519>().erase().data
```

Now, we just generate functions per arity, the same as we do with LLVM IR gen. Since each function has the same type `jit_function`, we can skip on the second bit of generation and build all of that without more JIT compilation, which drastically reduces compile times and memory usage.

```cpp
extern "C" jank::runtime::object_ref
clojure_core_unreduced_74270_1(jank::runtime::object_ref const unreduced,
                               jank::runtime::object_ref const x)
{
  jank::runtime::oref<jank::runtime::object> if_74271{};
  auto const call_74272(
    jank::runtime::dynamic_call(reinterpret_cast<jank::runtime::var *>(0x7fffe7da7fa0)->deref(),
                                x));
  if(jank::runtime::truthy(call_74272))
  {
    auto const call_74273(
      jank::runtime::dynamic_call(reinterpret_cast<jank::runtime::var *>(0x7fffe79aa550)->deref(),
                                  x));
    return call_74273;
  }
  else
  {
    return x;
  }
}
```

## Results
|test|branch: `main`| branch: `cppgen` | reduction |
|--|--|--|--|
| `./bin/compile` | 5m 3s | 3m 25s | 32% |
| `jank repl` | 1m 16s | 1s | 99% |

### DIFF
--- a/.github/workflows/build-compiler+runtime.yml
+++ b/.github/workflows/build-compiler+runtime.yml
@@ -186,6 +186,12 @@ jobs:
           packages: default-jdk software-properties-common lsb-release npm lcov leiningen ccache curl git git-lfs build-essential entr libssl-dev libdouble-conversion-dev pkg-config ninja-build cmake zlib1g-dev libffi-dev doctest-dev gcc g++ libgc-dev libboost-all-dev
           # Increment this when the package list changes.
           version: 12
+      # Boost doesn't get cached properly with the above action.
+      # https://github.com/awalsh128/cache-apt-pkgs-action/issues/151
+      - name: Install boost
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt install -y libboost-all-dev
       - name: Install Homebrew packages
         if: runner.os == 'macOS'
         run: |


jeaye:
@copilot I'm seeing some premature collections from the GC with these changes. Generally, this happens when there are JIT compiled globals which haven't been added as roots. However, I believe I have addressed all of those. In fact, we no longer JIT compile globals, since we reference existing objects in memory by address (and add them as roots for safety). Please investigate this.

PR #704, opened 2026-02-25
Comply with Clojure's reader suppression

shantanu-sardesai:
- jank will limit itself to the reader suppression that Clojure has.
- Still keeping the helper utility that I made.
- jank will treat the reader conditional as having syntax of it's own.
  -  Will not consume it as a list.
- Remove the entry for jank's reader suppression from the book.

Added context: https://gist.github.com/shantanu-sardesai/b51510aef1d34cf8136da4a98aafbc93.

### DIFF
--- a/compiler+runtime/include/cpp/jank/read/parse.hpp
+++ b/compiler+runtime/include/cpp/jank/read/parse.hpp
@@ -106,10 +106,9 @@ namespace jank::read::parse
     iterator end();
 
   private:
-    /* TODO: Switch to `jtl::ref` for the function pointer. */
     jtl::result<native_vector<object_source_info>, error_ref>
     parse_upto(lex::token_kind const &upto,
-               error_ref (*unterminated_form_error)(read::source const &));
+               jtl::ref<error_ref(read::source const &)> const unterminated_form_error);
     jtl::result<runtime::object_ref, error_ref> syntax_quote(runtime::object_ref const form);
     jtl::result<runtime::object_ref, error_ref>
     syntax_quote_expand_seq(runtime::object_ref const seq);
--- a/compiler+runtime/src/cpp/jank/read/parse.cpp
+++ b/compiler+runtime/src/cpp/jank/read/parse.cpp
@@ -317,7 +317,7 @@ namespace jank::read::parse
 
   jtl::result<native_vector<object_source_info>, error_ref>
   processor::parse_upto(lex::token_kind const &upto,
-                        error_ref (*unterminated_form_error)(read::source const &))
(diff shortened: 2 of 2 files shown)


shantanu-sardesai:
The lexer error fix was merged in the first PR. We have a couple of tests in parser for those cases ([here](https://github.com/jank-lang/jank/blob/main/compiler+runtime/test/cpp/jank/read/parse.cpp#L699-L705)).

I'll try to check the CI failure, but it's on Linux and seems to be in the `nrepl` code. Anyways, I'll check the error more thoroughly and see how my changes could have affected that part.

jeaye:
> The leader error fix was merged in the first PR. We have a couple of tests in parser for those cases ([here](https://github.com/jank-lang/jank/blob/main/compiler+runtime/test/cpp/jank/read/parse.cpp#L699-L705)).

Sounds good.

> I'll try to check the CI failure, but it's on Linux and seems to be in the `nrepl` code. Anyways, I'll check the error more thoroughly and see how my changes could have affected that part.

Just merge main for the fix.

shantanu-sardesai:
> > I'll try to check the CI failure, but it's on Linux and seems to be in the nrepl code. Anyways, I'll check the error more thoroughly and see how my changes could have affected that part.

> Just merge main for the fix.

First ever meme attempt and it failed 🙃

I'll rebase instead!

Jokes aside, the build is green now. I guess this is related to [this](https://github.com/jank-lang/jank/commit/7084399dbbc4c693912a0e953fdedc84536d04b6#diff-1dcb9b7ac26f355aafe3d5a8d1427ea9af878d26baf20ca512aefcc1eacf4ad9R189-R194)?

compiler+runtime/include/cpp/jank/read/parse.hpp
```diff
@@ -106,12 +106,10 @@ namespace jank::read::parse
     iterator end();
 
   private:
-    /* Ideally the parser shuold parse the entire form irrespective of whether an error is found
-     * mid-parse or not. Doing so let's us continue parsing in cases where the error is suppressed
-     * (such as in cases of unsupported features in reader conditionals). */
+    /* TODO: Switch to `jtl::ref` for the function pointer. */
```
jeaye: `jtl::ref` supports this now.

shantanu-sardesai: [Dunzo](https://github.com/jank-lang/jank/pull/704/commits/dd33d0746d6a50c0ec87b90d6dba5d7db608c095).

PR #711, opened 2026-03-11
feat(lein-jank): adds testing support

Samy-33:
Usage:

`lein jank test`

# What is it?

- leiningen's jank plugin should support running all the tests in the project
- `lein jank test` should follow `lein test` behaviour

# How does `lein test` behave?

- Merges the `project` map with `:test` profile
- Traverses the paths and finds all the namespaces based on the file-names
  - `my/app/test_app.clj` becomes `my.app.test-app` namespace
- Triggers evaluation of these namespaces using `leiningen.core.eval/eval-in-project`
- In this process, it dynamically requires all the namespaces and calls `clojure.test/run-tests`
- Calls `System.exit(error-code)` on errors and prints whatever `clojure.test/run-tests` prints

# How should `lein jank test` should behave

- Merge project with `[:test :jank]` profile
- Traverse the paths and find all jank and .cljc namespaces in test paths
- Create a temporary file while requiring all the found namespaces
- Call clojure.test/run-tests
- Check if jank binary's exit code and propagate it to the user

# Todos

- [x] Find how `lein test` behaves
- [x] Find how `lein jank test` should behave
- [x] Divide the task into smaller tasks to achieve final `lein jank test` behavior
- [x] Merge the project with `:test` profile and get the test paths
- [x] Find all the namespaces in the test paths
- [x] Generate a temporary file that requires all the namespaces and calls `clojure.test/run-tests`
- [x] Shell out to `jank` binary with module path that has paths, test paths and temporary file
- [x] Fail if jank binary exits with a non-zero code

### DIFF
--- a/book/src/getting-started/03-hello-leiningen.md
+++ b/book/src/getting-started/03-hello-leiningen.md
@@ -111,9 +111,16 @@ test which will fail.
 $ lein test
 ```
 
-> [!NOTE]
-> This is not yet supported in jank. Accomplishing this requires finding
-> all test namespaces and running them using `clojure.test`.
+If tests exist in directories other than `test/`, use `:test` profile to point leiningen
+to them.
+
+```clojure
+(defproject hello_lein "0.1-SNAPSHOT"
+  ;; ...
+  :profiles {;; ...
+             :test {:test-paths ["tests"]}})
+
+```
 
 ## Compiling a Leiningen project
 It's possible to AOT (ahead of time) compile our whole Leiningen project to an


lein-jank/README.md
```diff
@@ -1,37 +1,15 @@
...
-- Run the project: `lein jank run <args>`
-
-Make sure you have `jank` executable on your `PATH`.
-
-Run the following to know more
-
-```
-lein jank help
-```
+For more information run `lein jank help`
```
jeaye: Let's not duplicate docs from the jank book. Instead, we should just link there.

Issue #713, opened 2026-03-15
Support jank on MS-Windows (MSYS2/CLANG64 toolchain)


ikappaki:
Add support for building, testing, documenting and running jank on Windows using the MSYS2/CLANG64 toolchain.

A separate port has been developed in [jank-win](https://github.com/ikappaki/jank-win), tracking the main branch. Current status is documented here: 
https://github.com/jank-lang/jank/discussions/371#discussioncomment-15732984

Initial plan:

1. Open a PR to merge the [cppinterop](https://github.com/ikappaki/jank-CppInterOp/compare/main-jank.26-01-24..main-win)/[folly](https://github.com/ikappaki/jank-folly/compare/main-jank.first-25-04-07..main-win)/[zip](https://github.com/ikappaki/jank-zip/compare/master-jank.25-10-11..master-win) third party changes.
2. Open a PR to merge the baselines for the new third party libs (libchstk, MINGW-packages).
3. Open a PR to merge the [ jank changes](https://github.com/ikappaki/jank-win/compare/main-jank.26-02-26..main-win).

Thanks

PR #716, opened 2026-03-17
Add support for MS-Windows using the MSYS2/CLANG64 toolchain

ikappaki:
Hi,

could you please consider adding support for Windows using the MSYS2/clang64 toolchain in support of #713.

The port requires updates to the following third-party libraries:

1. CppInterop: https://github.com/jank-lang/CppInterOp/pull/3
2. Folly: https://github.com/jank-lang/folly/pull/2
3. Zip: https://github.com/jank-lang/zip/pull/2

It also introduces the ([libchkstk](https://nullprogram.com/blog/2024/02/05/) (is a simple workaround for the `Symbols not found: [ ___chkstk_ms ] build error`) and [MINGW-packages/mingw-w64-llvm](https://github.com/msys2/MINGW-packages/tree/286268ebe2a59500cbcba3d2f1e86533a5c42cc1/mingw-w64-llvm)  (a package definition extracted, with the necessary patches to compile on the MSYS2 clang64 toolchain), under the third-party libraries, with their original snapshots from source starting at: 
https://github.com/jank-lang/jank/pull/714

The port is fully functional except for exception stack unwinding from the JIT layer, which is not yet supported by LLVM (tracked in https://github.com/ikappaki/jank-win/issues/2). As a result, several tests that throw exceptions from JIT code have been explicitly disabled:

- jank-test: as listed in compiler+runtime/test/jit_windows_skips.txt
- compiler+runtime/test/bash/clojure.test/src/clojure_test/self_test.jank
- compiler+runtime/test/bash/error-reporting/bin/jank/test/error_reporting.clj
- compiler+runtime/test/bash/cli/pass-test

I will update the documentation once the review process is complete. We should also consider how to approach binary packaging, since there is currently no registry available to host `mingw-w64-jank` alpha binaries.

I also plan to update this description with a bullet point summary of the changes.

Thanks


ikappaki:
Hi @jeaye,

I think this is ready for review.

The third-party patches have been committed upstream. The remaining step is to switch to upstream (see [https://github.com/ikappaki/jank-win/pull/40](https://github.com/ikappaki/jank-win/pull/40)). I plan to do this as a final commit after review, since keeping the current setup makes it easier to iterate and test fixes on my third-party branches if needed.

There is one commit authored under `chaos`, which is not a GH account and therefore cannot sign the CLA required for merging. Rewriting the history to change the author to `ikappaki` would involve resolving a large number of conflicts, so instead I plan to squash the commits after review so that no commits remain under `chaos`.

The remaining items are:

* Move the build instructions from `README.md` to the dedicated build documentation (planned next).
* Decide on a plan for publishing binary releases (now or in a follow-up)

Let me know if there’s anything I can do to make the review easier.

Thanks

compiler+runtime/bin/jank/compiler+runtime/bash_test.clj
```diff
@@ -3,22 +3,55 @@
 (ns jank.compiler+runtime.bash-test
   (:require [clojure.string]
             [jank.util :as util]
+            [babashka.cli :as cli]
             [babashka.process :as b.p]
             [babashka.fs :as b.f]))
 
 (def compiler+runtime-dir (str (b.f/parent *file*) "/../../.."))
 
+(def cli-spec
```
jeaye: This is a new feature which is not related to Windows. Please do not include anything in this PR unless it's specific to the Windows port. Anything else can be a separate discussion.

bin/jank/util.clj
```diff
@@ -32,11 +53,11 @@
           (recur rem-ms (rest units) (if val (conj result val) result)))))))
 
 (defn log-boundary [title]
-  (println "\n────────────────" title "────────────────")
+  (println "\n----------------" title "----------------")
   (summary/boundary title))
 
 (defn log-step [title]
-  (println "\n────" title "────")
+  (println "\n----" title "----")
```
jeaye: Is Windows not able to render Unicode in terminal output?

.github/workflows/build-win.yml
```diff
@@ -0,0 +1,138 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include: [
+          # { msystem: UCRT64, runner: windows-2022},
+          { msystem: CLANG64, runner: windows-2022},
+          # { msystem: MINGW64, runner: windows-2022},
+          # { msystem: MINGW32, runner: windows-2022},
+          # { msystem: CLANGARM64, runner: windows-11-arm}
+        ]
```
jeaye: Do we want these other runners? If not, let's remove them or leave a comment explaining why we have them here but disabled.

.github/workflows/build-win.yml
```diff
@@ -0,0 +1,138 @@
+name: build MS-Windows
```
jeaye: Why do we have two workflows for building on Windows? They both do very similar things and seem to me like they should be one workflow.

compiler+runtime/bin/jank/compiler+runtime/build+test.clj
```diff
@@ -41,13 +41,13 @@
...
                             (conj "-DCMAKE_CXX_CLANG_TIDY=clang-tidy-cache-wrapper"))
-          configure-cmd (str "./bin/configure " (clojure.string/join " " configure-flags))
+          configure-cmd (apply util/bash-run "./bin/configure" configure-flags)
           stats-cmd (if (= "on" analyze)
                       "clang-tidy-cache"
                       "ccache")]
       (util/quiet-shell {:dir compiler+runtime-dir
                          :extra-env exports}
-                        (str stats-cmd " --zero-stats"))
+                        (util/bash-run stats-cmd " --zero-stats"))
```
jeaye: Instead of adding `util/bash-run` to every call of `util/quiet-shell`, why not do it within `quiet-shell`?

compiler+runtime/bin/compile
```diff
@@ -5,4 +5,11 @@ set -euo pipefail
 here="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 build_dir="${here}/../build"
 
+# echo :build-dir ${build_dir}
+# ninja -C "${build_dir}" CMakeFiles/jank_lib.dir/src/cpp/jank/codegen/llvm_processor.cpp.obj
+# cmake --build "${build_dir}" "$@" --target CMakeFiles/jank_lib.dir/src/cpp/jank/codegen/llvm_processor.cpp.obj
+# cmake --build "${build_dir}" "$@" --target  third-party/ftxui/libftxui-screen.a
+# cmake --graphviz=deps.dot "${build_dir}"
+# cmake --build "${build_dir}" --verbose "$@" -- -j1
+# cmake --build "${build_dir}" --verbose "$@"
```
jeaye: Looks like we can remove all of this.

compiler+runtime/include/cpp/jank/prelude.hpp
```diff
@@ -13,3 +13,13 @@
...
+
+namespace jank
+{
+  constexpr bool win32 =
+#ifdef _WIN32
+    true;
+#else
+    false;
+#endif
+}
```
jeaye: We already have a platform check in `primitive.hpp`. Please remove this.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -115,7 +115,7 @@ int main(int argc, const char** argv)
   )");
 
     auto const tmp_dir{ std::filesystem::temp_directory_path() };
-    std::string main_file_path{ tmp_dir / "jank-main-XXXXXX" };
+    std::string main_file_path = (tmp_dir / "jank-main-XXXXXX").string();
```
jeaye: Please always use direct initialization, as we were before.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -138,19 +138,24 @@ int main(int argc, const char** argv)
     }
     auto const clang_dir{ std::filesystem::path{ clang_path_str.unwrap().c_str() }.parent_path() };
     compiler_args.emplace_back(strdup("-I"));
-    compiler_args.emplace_back(strdup((clang_dir / "../include").c_str()));
+    compiler_args.emplace_back(strdup((clang_dir / "../include").string().c_str()));
+#ifdef __MINGW64__
+#else
```
jeaye: I think you just want `#ifndef` here.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -216,7 +221,10 @@ int main(int argc, const char** argv)
     {
       compiler_args.push_back(strdup("-Wl,--export-dynamic"));
     }
+#ifdef _WIN32
+#else
```
jeaye: Same, this should just be `#ifndef`.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -342,12 +353,13 @@ int main(int argc, const char** argv)
...
         ? util::format("{}/{}.o", util::cli::opts.output_dir, module::module_to_path(module_name))
-        : jtl::immutable_string{ util::cli::opts.output_module_filename }
+            .c_str()
+        : jtl::immutable_string{ util::cli::opts.output_module_filename }.c_str()
     };
     std::filesystem::create_directories(module_path.parent_path());
 
     auto const tmp{ std::filesystem::temp_directory_path() };
-    std::string path_tmp{ tmp / "jank-compile-XXXXXX" };
+    std::string path_tmp = (tmp / "jank-compile-XXXXXX").string();
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/aot/processor.cpp
```diff
@@ -359,11 +371,13 @@ int main(int argc, const char** argv)
     std::filesystem::path const jank_resource_dir{ util::resource_dir().c_str() };
     compiler_args.push_back("-include");
     auto const prelude_path{ jank_resource_dir / "include/cpp/jank/prelude.hpp" };
-    compiler_args.push_back(prelude_path.c_str());
+    std::string const prelude_path_str = prelude_path.string();
+    std::string const module_path_str = module_path.string();
+    compiler_args.push_back(prelude_path_str.c_str());
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
```diff
@@ -2189,8 +2189,16 @@ namespace jank::codegen
        * exception information and transfers control flow to the landing pad block. */
       auto personality_fn_type{ llvm::FunctionType::get(ctx->builder->getInt32Ty(),
                                                         /*isVarArg=*/true) };
+#if defined(__MINGW64__)
+      // Windows SEH exceptions are not supported yet.
+      // This is only to allow compilation; otherwise __gxx_personality_v0
+      // would be undefined.
+      auto personality_fn{ llvm_module->getOrInsertFunction("__gxx_personality_seh0",
```
jeaye: Please use `/* */` style comments, which is what we use in this codebase.

compiler+runtime/src/cpp/jank/environment/check_health.cpp
```diff
@@ -373,8 +373,9 @@ namespace jank::environment
     JANK_TRY
     {
       auto const tmp{ std::filesystem::temp_directory_path() };
-      std::string path_tmp{ tmp / "jank-aot-XXXXXX" };
-      mkstemp(path_tmp.data());
+      std::string path_tmp = (tmp / "jank-aot-XXXXXX").string();
+      int const fd = mkstemp(path_tmp.data());
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/environment/check_health.cpp
```diff
@@ -406,11 +407,12 @@ namespace jank::environment
       aot_prc.build_executable(util::cli::opts.target_module).expect_ok();
 
       auto const stdout_file{ std::filesystem::path{ path_tmp } / "stdout" };
+      std::string const stdout_file_str = stdout_file.string();
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/read/lex.cpp
```diff
@@ -1230,7 +1230,8 @@ namespace jank::read::lex
               while(true)
               {
                 auto const result(convert_to_codepoint(file.substr(pos), pos));
-                if(result.is_err() || !is_symbol_char(result.expect_ok().character))
+                if(result.is_err() || !is_symbol_char(result.expect_ok().character)
+                   || result.expect_ok().len == 0)
```
jeaye, line 1243: Why is this change needed?

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -31,15 +35,22 @@
 
 namespace jank::runtime::module
 {
+#ifdef _WIN32
+  constexpr char PATH_SEP = ';';
+#else
+  constexpr char PATH_SEP = ':';
+#endif
+
```
jeaye: We already have path separators in `loader.hpp`.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -1,4 +1,8 @@
-#include <sys/mman.h>
+#ifdef __MINGW64__
+  #include <windows.h>
```
jeaye, line 4: I think we can define `WIN32_LEAN_AND_MEAN` before including this.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -672,6 +689,48 @@ namespace jank::runtime::module
...
+
+#ifdef _WIN32
+    HANDLE hFile = CreateFileA(path.c_str(),
+                               GENERIC_READ,
+                               FILE_SHARE_READ,
+                               nullptr,
+                               OPEN_EXISTING,
+                               FILE_ATTRIBUTE_NORMAL,
+                               nullptr);
+
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
```diff
@@ -672,6 +689,48 @@ namespace jank::runtime::module
...
+    }
+
+    LARGE_INTEGER fileSize;
+    if(!GetFileSizeEx(hFile, &fileSize))
+    {
+      CloseHandle(hFile);
+      return error::internal_runtime_failure("Failed to get file size");
+    }
+
+    HANDLE hMapping = CreateFileMappingA(hFile, nullptr, PAGE_READONLY, 0, 0, nullptr);
```
jeaye: Please use direct initialization AND default-construct every primitive. i.e. `fileSize{};`.

compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -829,7 +829,7 @@ namespace jank::runtime
         break;
     }
 
-#ifdef JANK_MACOS_LIKE
+#if defined(JANK_MACOS_LIKE) || defined(_WIN32)
```
jeaye: We have a `JANK_WINDOWS_LIKE` for this.

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -33,10 +33,11 @@ namespace jank::util
   static bool is_clang_correct_version(std::filesystem::path const &path)
   {
     auto const tmp{ std::filesystem::temp_directory_path() };
-    std::string path_tmp{ tmp / "jank-clang-XXXXXX" };
-    mkstemp(path_tmp.data());
-    auto const proc_code{ llvm::sys::ExecuteAndWait(path.c_str(),
-                                                    { path.c_str(), "--version" },
+    std::string path_tmp = (tmp / "jank-clang-XXXXXX").string();
+    int const fd = mkstemp(path_tmp.data());
+    close(fd);
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/util/clang.cpp
```diff
@@ -230,15 +231,21 @@ namespace jank::util
     auto const resource{ aot::find_resource("incremental.pch") };
     if(resource.is_some())
     {
-      runtime::__rt_ctx->jit_prc.vfs["/virtual/incremental.pch"]
+      static constexpr char virtual_path[] =
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/util/environment.cpp
```diff
@@ -18,6 +18,10 @@
 #include <jank/util/fmt.hpp>
 #include <jank/error/system.hpp>
 
+#ifdef _WIN32
+  #include <windows.h>
```
jeaye, line 24: Whenever we include `windows.h`, we want to define `WIN32_LEAN_AND_MEAN`.

compiler+runtime/src/cpp/jank/util/environment.cpp
```diff
@@ -136,22 +153,32 @@ namespace jank::util
...
+    char path[MAX_PATH];
+    const DWORD size = GetModuleFileNameA(nullptr, path, MAX_PATH);
+
+    if(size == 0)
+    {
+      return std::string();
+    }
+    std::string const spath(path, size);
+    return spath;
 #else
```
jeaye: 1. Please default initialize all primitives; i.e. `char path[MAX_PATH]{};`
2. Please use direct initialization
3. You can just `return {};` on line 162
4. By making `spath` a local which is `const`, we prevent NRVO; instead, return an rvalue or remove the `const`

compiler+runtime/src/cpp/jank/util/environment.cpp
```diff
@@ -207,7 +234,7 @@ namespace jank::util
       if(sdk_path.empty())
       {
         auto const tmp{ std::filesystem::temp_directory_path() };
-        std::string path_tmp{ tmp / "jank-xcrun-XXXXXX" };
+        std::string path_tmp = (tmp / "jank-xcrun-XXXXXX").string();
```
jeaye: Please use direct initialization.

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -1,6 +1,10 @@
 #include <cstdarg>
 #include <utility>
 
+#ifdef _WIN32
+  #include <windows.h>
```
jeaye, line 7: Please use `WIN32_LEAN_AND_MEAN`.

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -1032,10 +1036,17 @@ extern "C"
   {
     JANK_TRY
     {
+#ifdef _WIN32
+      std::setlocale(LC_CTYPE, ".UTF8");
+      std::locale::global(std::locale(".UTF-8"));
+      SetConsoleOutputCP(CP_UTF8);
+      SetConsoleCP(CP_UTF8);
```
jeaye, line 1044: If we can set the locale to UTF-8, why can't we print Unicode characters in the bash tests?

compiler+runtime/src/cpp/main.cpp
```diff
@@ -246,8 +246,9 @@ namespace jank
      * to later review that for error reporting. We automatically clean it up
      * and we reuse the same file over and over. */
     auto const tmp{ std::filesystem::temp_directory_path() };
-    std::string path_tmp{ tmp / "jank-repl-XXXXXX" };
-    mkstemp(path_tmp.data());
+    std::string path_tmp{ (tmp / "jank-repl-XXXXXX").string() };
+    int const fd = mkstemp(path_tmp.data());
```
jeaye: Please use direct initialization.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
```diff
@@ -20,17 +19,17 @@
   "Redirect the given file stream to a capture sink, available by `readback`.
   Returns a capture handle."
   [fp*]
-  (let [fp                  (cpp/unbox (:* FILE) fp*)
+  (let [fp                  (cpp/unbox (:* cpp/FILE) fp*)
```
jeaye: This change is not required.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
```diff
@@ -1,15 +1,14 @@
...
+#include <filesystem>")
 
-(def temp-file-template "/tmp/jank-nrepl-XXXXXX")
+(def temp-file-template (str (cpp/getenv #cpp "TEMP") "/" "jank-nrepl-XXXXXX"))
 
 (defn- make-temp-file []
-  (let [name (cpp/std.string (cpp/cast std.string temp-file-template))
-        fd   (cpp/mkstemp (.data name))]
+  (let [name (cpp/std.string (cpp/cast cpp/std.string temp-file-template))
+        fd   (cpp/mkstemp (cpp/.data name))]
```
jeaye: This is not required.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
```diff
@@ -1,15 +1,14 @@
...
+(cpp/raw "#include <cstdio>
+#include <filesystem>")
 
-(def temp-file-template "/tmp/jank-nrepl-XXXXXX")
+(def temp-file-template (str (cpp/getenv #cpp "TEMP") "/" "jank-nrepl-XXXXXX"))
 
 (defn- make-temp-file []
-  (let [name (cpp/std.string (cpp/cast std.string temp-file-template))
-        fd   (cpp/mkstemp (.data name))]
+  (let [name (cpp/std.string (cpp/cast cpp/std.string temp-file-template))
```
jeaye: This is not required.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
```diff
@@ -1,15 +1,14 @@
-(ns jank.nrepl.server.capture)
+(ns jank.nrepl.server.capture
+  (:require [jank.nrepl.server.util :refer [close dup dup2 unlink]]))
 
-(cpp/raw "#include <unistd.h>")
-(cpp/raw "#include <stdio.h>")
-(cpp/raw "#include <fcntl.h>")
-(cpp/raw "#include <filesystem>")
+(cpp/raw "#include <cstdio>
+#include <filesystem>")
```
jeaye: If we're redoing the includes, we should just use the `:include` feature of `ns` instead.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
```diff
@@ -42,22 +41,22 @@
   "End the capture. Make sure to pair this with every `start` call or else the
   stream will remain permanently redirected."
   [{:keys [fp* bak-fd temp-fd]}]
-  (let [fp (cpp/unbox (:* FILE) fp*)
+  (let [fp (cpp/unbox (:* cpp/FILE) fp*)
```
jeaye: Please don't use `cpp/` when it's not needed.

compiler+runtime/src/jank/jank/nrepl/server/inspect.jank
```diff
@@ -1,14 +1,29 @@
...
 namespace jank::nrepl::server::inspect
 {
   inline std::string hash_to_hex_string(unsigned int i)
   {
+
+#ifdef _WIN32
+    // MSYS2/clang64 JIT workaround: std::format emits 128-bit division helpers
+    // which the JIT cannot resolve:
+    //
+    // JIT session error: Symbols not found: [ __umodti3, __udivti3 ]
```
jeaye: Has this been reported as a Clang bug?

compiler+runtime/test/bash/ahead-of-time/pass-test
```diff
@@ -28,6 +31,11 @@
     (println "Cleaned up 'target'")
     (f)))
 
+(defmacro string=
+  "Compare L and R strings ignoring line ending differences."
+  [l r]
+  `(= (str/split-lines ~l) (str/split-lines ~r)))
```
jeaye, line 37: We already defined this in the `util` ns.

compiler+runtime/test/bash/clojure.test/src/clojure_test/self_test.jank
```diff
@@ -1,13 +1,20 @@
 (ns clojure-test.self-test
   (:require [clojure.test :refer [deftest is] :as t]))
 
+;; Skip exception tests on Windows, not yet supported
+(cpp/raw "#ifdef _WIN32\n# define SELF_TEST_WIN 1\n#else\n# define SELF_TEST_WIN 0\n#endif")
```
jeaye: We already have `JANK_WINDOWS_LIKE`.

compiler+runtime/include/cpp/jank/runtime/module/loader.hpp
```diff
@@ -63,6 +64,27 @@ namespace jank::runtime::module
     jtl::immutable_string path;
   };
 
+#ifdef _WIN32
+  #include <windows.h>
```
jeaye: Including `windows.h` in a jank header file is HEAVY. If the only thing we want from it is `HANDLE`, let's just use `void*` instead.

Also, everywhere in this PR which you're using `_WIN32`, please instead use `JANK_WINDOWS_LIKE`.

jeaye (changes requested):

compiler+runtime/bin/jank/compiler+runtime/bash_test.clj
ikappaki: Before I open a separate discussion, is there a way to run or filter specific tests within the bash test suite that I might have missed?

This change has been extremely useful while developing the Windows port and investigating test failures.

Thanks

bin/jank/util.clj
ikappaki: This was an early change I made while getting everything working during the port.

On Windows, the terminal starts in an ASCII code page and only switches to Unicode if a program explicitly changes the code page. I wasn’t able to trigger that change from within babashka/Java, so I worked around it by forcing output to ASCII.

However, once runs one windows once, the terminal is forced to switch to Unicode and everything works correctly afterward. So it’s not that Unicode output is unavailable, just that jank needs to run first before invoking these scripts, which is likely already the case during development on a Windows bash terminal.

Given that, I think it’s better to revert this change.

.github/workflows/build-win.yml
ikappaki: We don’t need this anymore, this was the initial workflow I created during the port and can now be retired in favor of `build-win-compiler+runtime.yml`, which aligns with the standard unix/macos workflow.

.github/workflows/build-win.yml
ikappaki: We don’t need this anymore, this was the initial workflow I created during the port and can I will reiterate in in favor of `build-win-compiler+runtime.yml`, which aligns with the standard unix/macos workflow.

compiler+runtime/bin/jank/compiler+runtime/bash_test.clj
jeaye: There is no filtering capability right now, from within the `bash_test.clj` script. However, the bash tests are all just directories with a `pass-test` script inside. Each test is standalone, so I just `cd` into whichever test I want to focus on and run` ./pass-test` when I want to run it.

ikappaki: I've opened a feature request at https://github.com/jank-lang/jank/discussions/725.

ikappaki: The suggestion to cd into the test directory and run pass-test didn’t work in my case, since the path to the jank executable is set up by bash_test.clj. I’ve added more details to the discussion.

I’ve reverted the change for now, but running individual tests is still quite cumbersome, so please do consider the feature request.

bin/jank/util.clj
ikappaki: reverted.

.github/workflows/build-win.yml
ikappaki: workflow removed.

.github/workflows/build-win.yml
ikappaki: workflow removed.

compiler+runtime/bin/jank/compiler+runtime/build+test.clj
ikappaki: Incorporated running with bash on windows under `util/quiet-shell` as suggested, under the name of `command-make-portable`. There are still some cases that do not go through `quiet-shell` and have been updated to use the new function directly.

compiler+runtime/bin/compile
ikappaki: reverted to original.

compiler+runtime/include/cpp/jank/prelude.hpp
ikappaki: I wasn't aware about the definitions in `primitive.hpp`. Change reverted, thanks

compiler+runtime/src/cpp/jank/aot/processor.cpp
ikappaki: updated.

compiler+runtime/src/cpp/jank/aot/processor.cpp
ikappaki: This was a stylistic choice to emphasize that the general case is the else branch, with `__MINGW64_`_ as the exception. Since that’s a bit confusing, I’ve switched it to `#ifndef` instead.

compiler+runtime/src/cpp/jank/aot/processor.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/aot/processor.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/aot/processor.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/codegen/llvm_processor.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/environment/check_health.cpp
ikappaki: Updated

compiler+runtime/src/cpp/jank/environment/check_health.cpp
ikappaki: Updated

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
ikappaki: Changed to use `module_separator`.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
ikappaki: Good idea! I’ve updated all occurrences to use `WIN32_LEAN_AND_MEAN` to improve build times (though I don’t expect a noticeable difference).

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/runtime/module/loader.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/runtime/core.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/util/clang.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/util/clang.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/util/environment.cpp
ikappaki: Updated.

ikappaki: Updated

compiler+runtime/src/cpp/jank/util/environment.cpp
ikappaki: Updated on all counts.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
ikappaki: Updated, I wasn't aware we can avoid using `cpp/` in types.

compiler+runtime/src/jank/jank/nrepl/server/inspect.jank
ikappaki: I took a closer look and found that clang already ships a static library providing these symbols. Linking against `libclang_rt.builtins-x86_64.a` resolves the issue. This also fixes the missing `__chkstk_ms` symbol, allowing us to remove the `libchkstk` workaround entirely, which I've done with the latest update. Thanks for prompting the investigation.

compiler+runtime/test/bash/ahead-of-time/pass-test
ikappaki: The `pass-test` and `fail-test` scripts can’t access `jank.util` because it’s not on the classpath.

compiler+runtime/test/bash/clojure.test/src/clojure_test/self_test.jank
ikappaki: I’ve switched to using `jtl.current_platform` instead, which avoids relying on raw C++ to check if a preprocessor symbols is defined.

compiler+runtime/include/cpp/jank/runtime/module/loader.hpp
ikappaki: Great point! I’ve updated it to forward declare as `using HANDLE = void*;` instead.

Also changed all `_WIN32` occurrences to `JANK_WINDOWS_LIKE`.

compiler+runtime/src/cpp/jank/util/environment.cpp
ikappaki: updated.

compiler+runtime/src/cpp/jank/c_api.cpp
ikappaki: Updated.

compiler+runtime/src/cpp/jank/c_api.cpp
ikappaki: It’s not just the locale, but also the calls to set console output. As far as I know, and based on my research, the latter API isn’t accessible in java or babashka, so we’re a bit stuck.

compiler+runtime/src/cpp/main.cpp
ikappaki: Updated.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
ikappaki: reverted.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
ikappaki: reverted

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
ikappaki: reverted.

compiler+runtime/src/jank/jank/nrepl/server/capture.jank
ikappaki: Updated to use `:include` instead.

compiler+runtime/src/cpp/jank/read/lex.cpp
ikappaki: From memory, the lexer on Windows would enter an endless loop, trying to parse additional characters for a symbol at the end of the stream. I can't recall whether this was due to CRLF line endings or a Windows specific parsing quirk, but the fix was to check that the remaining data length was 0.

I can revert the change locally and investigate further if a precise explanation is required.

ikappaki:
Hi @jeaye,

Could you please take another look at the PR? I believe I’ve addressed and responded to all comments.

Additionally:

1. I squashed the original commits to remove the `chaos` commit author, resolving the CLA issue. All tests are visible and passing.
2. I reset all thirdparty libs to upstream.
3. I updated `bin/configure` to set the correct install prefix path for `bin/install` to work on MSYS2.
4. I updated `compiler+runtime/doc/build.md` with instructions for building jank on Windows using MSYS2/CLANG64, and reverted `README.md` to its original state.
5. I merged the latest upstream changes (as of 2026-04-05). This introduced several tests around exception handling that were crashing the test runner, so I’ve added logic in `compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc` to skip those tests on Windows.

Thanks

compiler+runtime/CMakeLists.txt
```diff
@@ -731,6 +752,14 @@ foreach(dir ${jank_lib_includes_prop})
...
 
 set(jank_lib_linker_flags_list "")
+
+if(WIN32)
+  list(APPEND jank_lib_linker_flags_list
+    -lws2_32
+    -ldbghelp
+  )
+endif()
+
```
jeaye: Only one of these should be needed, right?

compiler+runtime/CMakeLists.txt
```diff
@@ -850,6 +889,20 @@ target_link_libraries(
...
+  target_link_libraries(
+    jank_exe_phase_1 PUBLIC
+    ${jank_link_whole_start} libclang_rt.builtins-x86_64.a ${jank_link_whole_end}
+    # Required at least for mkstemp
+    ${jank_link_whole_start} libmingwex.a ${jank_link_whole_end}
+
+    pthread
+    zstd
+    ws2_32
+    dbghelp
```
jeaye: This shouldn't be needed, if libjank already links to them.

compiler+runtime/CMakeLists.txt
```diff
@@ -964,6 +1017,20 @@ if(jank_test)
...
+      jank_test_exe PUBLIC
+      ${jank_link_whole_start} libclang_rt.builtins-x86_64.a ${jank_link_whole_end}
+      # Required at least for mkstemp
+      ${jank_link_whole_start} libmingwex.a ${jank_link_whole_end}
+
+      pthread
+      zstd
+      ws2_32
+      dbghelp
+    )
```
jeaye, line 1019: Same here.

jeaye (commented):
Great work! I have two questions.

1. I left some comments about seemingly redundant CMake linkage stuff
2. Why was BDWGC updated? Where is the PR for that?

compiler+runtime/CMakeLists.txt
ikappaki: I haven’t investigated how linking is handled across the different phases, but removing `ws2_32` results in the following errors:

```bash
 bin/compile
[374/422] Linking CXX executable jank-phase-1.exe
FAILED: [code=1] jank-phase-1.exe libjank-phase-1.dll.a
C:\WINDOWS\system32\cmd.exe /C "cd . && D:\msys64\jank-20\clang64\bin\clang++.exe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wno-pass-failed -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -O3 -DNDEBUG -Wl,--stack,16777216 -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++ -L/opt/homebrew/opt/llvm/lib/unwind -Wl,--export-all-symbols -L D:/src/jank-win/compiler+runtime/build CMakeFiles/jank_exe_phase_1.dir/src/cpp/main.cpp.obj CMakeFiles/jank_exe_phase_1.dir/src/cpp/jank/environment/check_health.cpp.obj -o jank-phase-1.exe -Wl,--out-implib,libjank-phase-1.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -L/lib -Wl,--whole-archive  libjank-standalone-phase-1.a  -Wl,--no-whole-archive  -lz  D:/removeme/msys64/jank-20/clang64/lib/libclang-cpp.dll.a  D:/removeme/msys64/jank-20/clang64/lib/libcrypto.dll.a  -Wl,--whole-archive  -Wl,-Bstatic  -lclang_rt.builtins-x86_64  -Wl,-Bdynamic  -Wl,--no-whole-archive  -Wl,--whole-archive  -Wl,-Bstatic  -lmingwex  -Wl,-Bdynamic  -Wl,--no-whole-archive  -lpthread  -lzstd  D:/removeme/msys64/jank-20/clang64/lib/libLLVM-22-rc1.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
ld.lld: warning: CMakeFiles/jank_exe_phase_1.dir/src/cpp/main.cpp.obj: locally defined symbol imported: cpptrace::detail::do_prepare_unwind_interceptor(char (*)(unsigned long long)) (defined in libjank-standalone-phase-1.a(from_current.cpp.obj)) [LNK4217]
ld.lld: warning: CMakeFiles/jank_exe_phase_1.dir/src/cpp/main.cpp.obj: locally defined symbol imported: cpptrace::detail::collect_current_trace(unsigned long long) (defined in libjank-standalone-phase-1.a(from_current.cpp.obj)) [LNK4217]
ld.lld: warning: CMakeFiles/jank_exe_phase_1.dir/src/cpp/jank/environment/check_health.cpp.obj: locally defined symbol imported: cpptrace::detail::do_prepare_unwind_interceptor(char (*)(unsigned long long)) (defined in libjank-standalone-phase-1.a(from_current.cpp.obj)) [LNK4217]
ld.lld: warning: CMakeFiles/jank_exe_phase_1.dir/src/cpp/jank/environment/check_health.cpp.obj: locally defined symbol imported: cpptrace::detail::collect_current_trace(unsigned long long) (defined in libjank-standalone-phase-1.a(from_current.cpp.obj)) [LNK4217]
ld.lld: error: undefined symbol: __declspec(dllimport) WSACleanup
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(__dtor__ZN5folly6netops12_GLOBAL__N_111winsockInitE)
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(boost::asio::detail::winsock_init<2, 2>::~winsock_init())
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(__dtor__ZN5boost4asio6detailL21winsock_init_instanceE)
>>> referenced 2 more times

ld.lld: error: undefined symbol: __declspec(dllimport) accept
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(folly::netops::accept(folly::NetworkSocket, sockaddr*, int*))
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(folly::netops::socketpair(int, int, int, folly::NetworkSocket*))
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(boost::asio::detail::socket_ops::sync_accept(unsigned long long, unsigned char, void*, unsigned long long*, boost::system::error_code&))
>>> referenced 1 more times
...
```

compiler+runtime/CMakeLists.txt
ikappaki: similarly:
```bash
[49/50] Linking CXX executable jank-test.exe
FAILED: [code=1] jank-test.exe libjank-test.dll.a
C:\WINDOWS\system32\cmd.exe /C "cd . && D:\msys64\jank-20\clang64\bin\clang++.exe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wno-pass-failed -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -O3 -DNDEBUG -Wl,--stack,16777216 -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++ -L/opt/homebrew/opt/llvm/lib/unwind -Wl,--export-all-symbols -L D:/src/jank-win/compiler+runtime/build CMakeFiles/jank_test_exe.dir/test/cpp/main.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jtl/immutable_string.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jtl/string_builder.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/util/fmt.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/util/path.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/read/lex.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/read/parse.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/analyz[49/50] Linking CXX executable jank-test.exe
FAILED: [code=1] jank-test.exe libjank-test.dll.a
C:\WINDOWS\system32\cmd.exe /C "cd . && D:\msys64\jank-20\clang64\bin\clang++.exe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wno-pass-failed -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -O3 -DNDEBUG -Wl,--stack,16777216 -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++ -L/opt/homebrew/opt/llvm/lib/unwind -Wl,--export-all-symbols -L D:/src/jank-win/compiler+runtime/build CMakeFiles/jank_test_exe.dir/test/cpp/main.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jtl/immutable_string.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jtl/string_builder.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/util/fmt.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/util/path.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/read/lex.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/read/parse.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/analyze/box.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/behavior/call.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/core/seq.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/detail/native_persistent_list.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/big_integer.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/big_decimal.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/persistent_string.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/ratio.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/persistent_list.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/persistent_vector.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/persistent_array_map.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/transient_array_map.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/range.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/integer_range.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/repeat.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/jit/processor.cpp.obj CMakeFiles/jank_test_exe.dir/core-libs/clojure/core.cpp.obj -o jank-test.exe -Wl,--out-implib,libjank-test.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -L/lib -Wl,--whole-archive  libjank-standalone-phase-1.a  -Wl,--no-whole-archive  -lz  D:/removeme/msys64/jank-20/clang64/lib/libclang-cpp.dll.a  D:/removeme/msys64/jank-20/clang64/lib/libcrypto.dll.a  -Wl,--whole-archive  -Wl,-Bstatic  -lclang_rt.builtins-x86_64  -Wl,-Bdynamic  -Wl,--no-whole-archive  -Wl,--whole-archive  -Wl,-Bstatic  -lmingwex  -Wl,-Bdynamic  -Wl,--no-whole-archive  -lpthread  -lzstd  D:/removeme/msys64/jank-20/clang64/lib/libLLVM-22-rc1.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
ld.lld: error: undefined symbol: __declspec(dllimport) WSACleanup
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(__dtor__ZN5folly6netops12_GLOBAL__N_111winsockInitE)
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(boost::asio::detail::winsock_init<2, 2>::~winsock_init())
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(__dtor__ZN5boost4asio6detailL21winsock_init_instanceE)
>>> referenced 2 more times

ld.lld: error: undefined symbol: __declspec(dllimport) accept
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(folly::netops::accept(folly::NetworkSocket, sockaddr*, int*))
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(folly::netops::socketpair(int, int, int, folly::NetworkSocket*))
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(boost::asio::detail::socket_ops::sync_accept(unsigned long long, unsigned char, void*, unsigned long long*, boost::system::error_code&))
>>> referenced 1 more times
e/box.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/behavior/call.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/core/seq.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/detail/native_persistent_list.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/big_integer.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/big_decimal.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/persistent_string.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/ratio.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/persistent_list.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/persistent_vector.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/persistent_array_map.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/transient_array_map.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/range.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/integer_range.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/runtime/obj/repeat.cpp.obj CMakeFiles/jank_test_exe.dir/test/cpp/jank/jit/processor.cpp.obj CMakeFiles/jank_test_exe.dir/core-libs/clojure/core.cpp.obj -o jank-test.exe -Wl,--out-implib,libjank-test.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -L/lib -Wl,--whole-archive  libjank-standalone-phase-1.a  -Wl,--no-whole-archive  -lz  D:/removeme/msys64/jank-20/clang64/lib/libclang-cpp.dll.a  D:/removeme/msys64/jank-20/clang64/lib/libcrypto.dll.a  -Wl,--whole-archive  -Wl,-Bstatic  -lclang_rt.builtins-x86_64  -Wl,-Bdynamic  -Wl,--no-whole-archive  -Wl,--whole-archive  -Wl,-Bstatic  -lmingwex  -Wl,-Bdynamic  -Wl,--no-whole-archive  -lpthread  -lzstd  D:/removeme/msys64/jank-20/clang64/lib/libLLVM-22-rc1.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
ld.lld: error: undefined symbol: __declspec(dllimport) WSACleanup
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(__dtor__ZN5folly6netops12_GLOBAL__N_111winsockInitE)
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(boost::asio::detail::winsock_init<2, 2>::~winsock_init())
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(__dtor__ZN5boost4asio6detailL21winsock_init_instanceE)
>>> referenced 2 more times

ld.lld: error: undefined symbol: __declspec(dllimport) accept
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(folly::netops::accept(folly::NetworkSocket, sockaddr*, int*))
>>> referenced by libjank-standalone-phase-1.a(NetOps.cpp.obj):(folly::netops::socketpair(int, int, int, folly::NetworkSocket*))
>>> referenced by libjank-standalone-phase-1.a(server.cpp.obj):(boost::asio::detail::socket_ops::sync_accept(unsigned long long, unsigned char, void*, unsigned long long*, boost::system::error_code&))
>>> referenced 1 more times
...
```

ikappaki:
>     2. Why was BDWGC updated? Where is the PR for that?

Sorry, this happened while I was experimenting with upstream bdwgc to investigate a GC issue that turned out to be the same as one on macOS, which has since been addressed. I wasn’t aware that upstream had moved on. I’ve now reverted back to the upstream commit, so there’s no difference anymore.

compiler+runtime/CMakeLists.txt
ikappaki: Sorry, could you elaborate a bit on this? If you’re referring to `ws2_32` and `dbghelp`, they are separate libraries: one provides socket functionality, while the other is used for stack traces and debugging. Including one doesn’t replace the other. For example, if `dbghelp` is omitted, we get the following linker errors:

```bash
$ bin/compile
[374/422] Linking CXX executable jank-phase-1.exe
FAILED: [code=1] jank-phase-1.exe libjank-phase-1.dll.a
C:\WINDOWS\system32\cmd.exe /C "cd . && D:msys64\jank-20\clang64\bin\clang++.exe -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wno-pass-failed -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -O3 -DNDEBUG -Wl,--stack,16777216 -L/opt/homebrew/opt/llvm/lib -L/opt/homebrew/opt/llvm/lib/c++ -L/opt/homebrew/opt/llvm/lib/unwind -Wl,--export-all-symbols -L D:/src/jank-win/compiler+runtime/build CMakeFiles/jank_exe_phase_1.dir/src/cpp/main.cpp.obj CMakeFiles/jank_exe_phase_1.dir/src/cpp/jank/environment/check_health.cpp.obj -o jank-phase-1.exe -Wl,--out-implib,libjank-phase-1.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -L/lib -Wl,--whole-archive  libjank-standalone-phase-1.a  -Wl,--no-whole-archive  -lz  D:/removeme/msys64/jank-20/clang64/lib/libclang-cpp.dll.a  D:/removeme/msys64/jank-20/clang64/lib/libcrypto.dll.a  -Wl,--whole-archive  -Wl,-Bstatic  -lclang_rt.builtins-x86_64  -Wl,-Bdynamic  -Wl,--no-whole-archive  -Wl,--whole-archive  -Wl,-Bstatic  -lmingwex  -Wl,-Bdynamic  -Wl,--no-whole-archive  -lpthread  -lzstd  -lws2_32  D:/removeme/msys64/jank-20/clang64/lib/libLLVM-22-rc1.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
ld.lld: warning: CMakeFiles/jank_exe_phase_1.dir/src/cpp/main.cpp.obj: locally defined symbol imported: cpptrace::detail::do_prepare_unwind_interceptor(char (*)(unsigned long long)) (defined in libjank-standalone-phase-1.a(from_current.cpp.obj)) [LNK4217]
ld.lld: warning: CMakeFiles/jank_exe_phase_1.dir/src/cpp/main.cpp.obj: locally defined symbol imported: cpptrace::detail::collect_current_trace(unsigned long long) (defined in libjank-standalone-phase-1.a(from_current.cpp.obj)) [LNK4217]
ld.lld: warning: CMakeFiles/jank_exe_phase_1.dir/src/cpp/jank/environment/check_health.cpp.obj: locally defined symbol imported: cpptrace::detail::do_prepare_unwind_interceptor(char (*)(unsigned long long)) (defined in libjank-standalone-phase-1.a(from_current.cpp.obj)) [LNK4217]
ld.lld: warning: CMakeFiles/jank_exe_phase_1.dir/src/cpp/jank/environment/check_health.cpp.obj: locally defined symbol imported: cpptrace::detail::collect_current_trace(unsigned long long) (defined in libjank-standalone-phase-1.a(from_current.cpp.obj)) [LNK4217]
ld.lld: error: undefined symbol: __declspec(dllimport) SymGetLineFromAddr64
>>> referenced by libjank-standalone-phase-1.a(symbols_with_dbghelp.cpp.obj):(cpptrace::detail::dbghelp::resolve_frame(void*, unsigned long long))

ld.lld: error: undefined symbol: __declspec(dllimport) SymFromAddr
>>> referenced by libjank-standalone-phase-1.a(symbols_with_dbghelp.cpp.obj):(cpptrace::detail::dbghelp::resolve_frame(void*, unsigned long long))

ld.lld: error: undefined symbol: __declspec(dllimport) SymSetContext
>>> referenced by libjank-standalone-phase-1.a(symbols_with_dbghelp.cpp.obj):(cpptrace::detail::dbghelp::resolve_frame(void*, unsigned long long))
...
```

jeaye: I am trying to point out that this PR links to `ws2_32` and `dbghelp` in three different places and only one of them should be needed.

1. libjank (the only one we should need)
2. jank executable (which links to libjank)
3. jank test executable (which links to libjank)

Generally, just linking deps into libjank is sufficient. If it's not, on Windows, then linking to in libjank shouldn't matter at all and we should just link in the other two. I don't think there's a world where all three require linking the same libs.

ikappaki: Thanks for the explanation. I was using the wrong variable, `jank_lib_linker_flags_list` which only covered AOT compiled binaries. The correct one, which also covers the executables, is `jank_linker_flags`.

I’ve updated `CMakeLists.txt` to use only the latter for referencing the two libraries.

compiler+runtime/CMakeLists.txt
ikappaki: I've removed it in favor of using `jank_linker_flags`, which covers both executables and AOT compiled binaries.

compiler+runtime/CMakeLists.txt
ikappaki: I've removed it in favor of using `jank_linker_flags`, which covers both executables and AOT compiled binaries.

ikappaki:
Hi @jeaye,

I’ve updated `CMakeLists.txt` to reference only `jank_linker_flags` for linking against `ws2_32` and `dbghelp`, which covers both executables and AOT compiled binaries.

I’ve also updated the PR to the latest main branch. Could you please have another look at the patch? I believe I’ve addressed all the comments raised, but it would be great if you could double check in case anything was missed. Thanks.

jeaye (approved):
Great work!

jeaye:
As a follow up to this, it would be good to hook our daily E2E test suite into the Windows build, too. We have a `.github/workflows/package-check.yml` workflow which installs jank from the binary packages on macOS, Ubuntu, and Arch and then runs a health check and the full E2E test suite. This workflow runs every morning (currently Arch is failing). Getting Windows into here should be our next goal, but it will require the ability to install packaged jank binaries for Windows.

ikappaki:
> As a follow up to this, it would be good to hook our daily E2E test suite into the Windows build, too. We have a `.github/workflows/package-check.yml` workflow which installs jank from the binary packages on macOS, Ubuntu, and Arch and then runs a health check and the full E2E test suite. This workflow runs every morning (currently Arch is failing). Getting Windows into here should be our next goal, but it will require the ability to install packaged jank binaries for Windows.

Hi @jeaye,

We can eventually add a package to the [MSYS2 packages](https://github.com/msys2/MSYS2-packages) repo, which is published via `pacman`. From my understanding, this requires stable, versioned releases rather than unversioned alpha builds. The workflow would be to submit a package definition PR first, and then update it via follow up PRs for new releases.

I’ve already created a pacman package definition in the [ikappaki/jank-win-release](https://github.com/ikappaki/jank-win-release) repo that builds and serves prebuilt Windows binaries for both jank and LLVM. This could be a good starting point when we’re ready to upstream into MSYS2.

For distributing prebuilt binaries in the meantime, do you happen to have a place where we could upload and serve them as an alternative `pacman` repo alongside the official one mentioned above? It looks like pacman only requires a simple HTTP server hosting the repo database and packages, but I haven’t fully explored this yet.

Alternatively, would you be open to repurposing the current  [ikappaki/jank-win-release](https://github.com/ikappaki/jank-win-release) so it triggers builds on successful build of `jank-lang/jank`? It could then publish releases to GH, and users could install or update via the small `jank-win-updater` script provided by the repo.  The idea would be to move ownership under `jank-lang` (e.g. `jank-lang/jank-win-release`) so binaries are officially published by the project and not perceived as third party builds.

Once we have a reliable distribution path for prebuilt binaries, we can then enable `.github/workflows/package-check.yml` to validate packaging.

Also, would you be OK with moving the GH issue describing the Windows LLVM exception problem from https://github.com/ikappaki/jank-win/issues/2 to `jank-lang/jank` for consinstency? And do you have any thoughts on how best to escalate or engage with LLVM on that issue?

Thanks!

jeaye:
> We can eventually add a package to the [MSYS2 packages](https://github.com/msys2/MSYS2-packages) repo, which is published via `pacman`. From my understanding, this requires stable, versioned releases rather than unversioned alpha builds. The workflow would be to submit a package definition PR first, and then update it via follow up PRs for new releases.

Ok, gotcha.


> For distributing prebuilt binaries in the meantime, do you happen to have a place where we could upload and serve them as an alternative `pacman` repo alongside the official one mentioned above? It looks like pacman only requires a simple HTTP server hosting the repo database and packages, but I haven’t fully explored this yet.

Yes! I have a dedicated VPS for storing jank binaries. We're using that for our Ubunutu PPA, which is also HTTP-based. However, since MSYS2 is using pacman, can we just use AUR and then have the AUR script pull the binary from the VPS? That way users don't need to add a custom repo to their pacman configs.

> Alternatively, would you be open to repurposing the current [ikappaki/jank-win-release](https://github.com/ikappaki/jank-win-release) so it triggers builds on successful build of `jank-lang/jank`? It could then publish releases to GH, and users could install or update via the small `jank-win-updater` script provided by the repo. The idea would be to move ownership under `jank-lang` (e.g. `jank-lang/jank-win-release`) so binaries are officially published by the project and not perceived as third party builds.

Let's not publish binary releases to Github for now. I would prefer to keep packages within our users' package managers.

> Also, would you be OK with moving the GH issue describing the Windows LLVM exception problem from [ikappaki#2](https://github.com/ikappaki/jank-win/issues/2) to `jank-lang/jank` for consinstency? And do you have any thoughts on how best to escalate or engage with LLVM on that issue?

I have escalated this and the LLVM relevant dev has sketched up a patch for you to try. :) Neither he nor I can try it, since we don't have Windows machines, but hopefully that's something you're up for.

Thank you!

ikappaki:
Hi @jeaye, sorry for the delay in replying, I was focused on putting together a patch to support SEH exceptions in LLVM, based on the sketch Lang provided at https://github.com/llvm/llvm-project/issues/163503. I finally got it working, and all exception tests in jank are now passing. I’ve submitted the patch for review to Lang so we can discuss how best to move it forward.

The patch currently targets the LLVM 23 branch, but LLVM 23 is unable to compile jank for reasons unrelated to the JIT or my changes. It gets stuck compiling `number.cpp` and eventually consumes all available memory. Because of that, I’ve thus also backported it to LLVM 22.

I haven't yet looked into how distribution works with the AUR or whether it’s compatible with the MSYS2/pacman infrastructure. That’s next on my list once the LLVM patch is sorted out.

Admin question:

Would it be okay if I moved the https://github.com/ikappaki/jank-win/issues/2 issue describing the Windows exception failures into the main jank project, so I can track the remaining changes there?

Thanks!

jeaye:
> Hi @jeaye, sorry for the delay in replying, I was focused on putting together a patch to support SEH exceptions in LLVM, based on the sketch Lang provided at [llvm/llvm-project#163503](https://github.com/llvm/llvm-project/issues/163503). I finally got it working, and all exception tests in jank are now passing. I’ve submitted the patch for review to Lang so we can discuss how best to move it forward.

Great job getting this going! I reached out to Lang to discuss this and when he offered to sketch something, I wasn't sure if it was going to be approachable. I'm glad you were able to make something of it.

> The patch currently targets the LLVM 23 branch, but LLVM 23 is unable to compile jank for reasons unrelated to the JIT or my changes. It gets stuck compiling `number.cpp` and eventually consumes all available memory. Because of that, I’ve thus also backported it to LLVM 22.

Oh, interesting. Sounds like a LLVM 23 bug. Smart move holding off for now.
 

> Would it be okay if I moved the [ikappaki#2](https://github.com/ikappaki/jank-win/issues/2) issue describing the Windows exception failures into the main jank project, so I can track the remaining changes there?

Sure, no problem. Feel free to make discussion and track progress there. We'll use issues for any concrete tasks which come out of the discussion.

ikappaki:
> Great job getting this going! I reached out to Lang to discuss this and when he offered to sketch something, I wasn't sure if it was going to be approachable. I'm glad you were able to make something of it.

Yes, thanks a lot for this, it definitely helped move things forward!

I’ve opened a couple of follow up discussions to track the next steps
1. Tracking SEH support in LLVM/jank:  https://github.com/jank-lang/jank/discussions/796
2. Discussing prebuild windows binaries for jank: https://github.com/jank-lang/jank/discussions/797

PR #730, opened 2026-04-07
Fix ambiguous overloads & bitwise operators

shantanu-sardesai:
Creating the PR to get feedback on error messages & possibly a better way of dealing with unnecessary argument usage in the error messages.

Brings us to **109 of 227 tests**.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
@@ -1010,161 +1010,182 @@ namespace jank::runtime
 
   i64 bit_not(object_ref const l)
   {
-    return visit_type<obj::integer>([](auto const typed_l) -> i64 { return ~typed_l->data; }, l);
+    auto const typed_l{ dyn_cast<obj::integer>(l) };
+
+    if(typed_l == jank_nil())
+    {
+      throw std::runtime_error{
+        "Bitwise operators are only supported for values of type integer."
+      };
+    }
+
+    return ~typed_l->data;
   }
 
   i64 bit_and(object_ref const l, object_ref const r)
   {
-    return visit_type<obj::integer>(
-      [](auto const typed_l, auto const r) -> i64 {
-        return visit_type<obj::integer>(
-          [](auto const typed_r, auto const typed_l) -> i64 { return typed_l & typed_r->data; },
-          r,
-          typed_l->data);
-      },
-      l,
-      r);
+    auto const typed_l{ dyn_cast<obj::integer>(l) };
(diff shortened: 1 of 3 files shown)


compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -938,161 +1010,182 @@ namespace jank::runtime
 
   i64 bit_not(object_ref const l)
   {
-    return visit_type<obj::integer>([](auto const typed_l) -> i64 { return ~typed_l->data; }, l);
+    auto const typed_l{ dyn_cast<obj::integer>(l) };
+
+    if(typed_l == jank_nil())
```
jeaye: `typed_l.is_nil()`

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -938,161 +1010,182 @@ namespace jank::runtime
...
-          [](auto const typed_r, auto const typed_l) -> i64 { return typed_l & typed_r->data; },
-          r,
-          typed_l->data);
-      },
-      l,
-      r);
+    auto const typed_l{ dyn_cast<obj::integer>(l) };
+    auto const typed_r{ dyn_cast<obj::integer>(r) };
+
+    if(typed_l == jank_nil() || typed_r == jank_nil())
```
jeaye: Please use `.is_nil()` for the rest of this PR. I won't comment on it again, but there are many other places you're doing this comparison with `jank_nil()`.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1116,6 +1209,27 @@ namespace jank::runtime
       r);
   }
 
+  bool lt(obj::nil_ref const l, obj::nil_ref const r)
+  {
+    throw std::runtime_error{
+      util::format("Can't compare {} with {}.", to_string(l), to_string(r))
```
jeaye: Please us `to_code_string` instead of `to_string` for all of these new functions.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
shantanu-sardesai: Done.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
shantanu-sardesai: Done.

PR #733, opened 2026-04-11
Enable `assoc!` tests

shantanu-sardesai:
Brings us to **111 of 227 tests**.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/obj/transient_vector.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/transient_vector.cpp
@@ -49,6 +49,43 @@ namespace jank::runtime::obj
     return this;
   }
 
+  transient_vector_ref transient_vector::assoc_in_place(object_ref const key, object_ref const val)
+  {
+    assert_active();
+    if(key->type != object_type::integer)
+    {
+      throw std::runtime_error{ "Key must be an integer." };
+    }
+
+    auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
+    auto const size(data.size());
+
+    if(i > size)
+    {
+      throw std::runtime_error{ "Index out of bounds." };
+    }
+    else if(i == size)
+    {
+      data.push_back(val);
+    }
+    else
+    {
+      data.set(i, val);
+    }
+
+    return this;
+  }
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+
+  transient_vector_ref transient_vector::dissoc_in_place(object_ref const key)
+  {
+    throw std::runtime_error{ "Unsupported dissoc operation!" };
+  }
+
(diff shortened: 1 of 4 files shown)


compiler+runtime/src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -49,6 +49,43 @@ namespace jank::runtime::obj
...
 
+  transient_vector_ref transient_vector::assoc_in_place(object_ref const key, object_ref const val)
+  {
+    assert_active();
+    if(key->type != object_type::integer)
+    {
+      throw std::runtime_error{ "Key must be an integer." };
+    }
+
+    auto const i(static_cast<size_t>(expect_object<integer>(key)->data));
```
jeaye: By converting to an unsigned type here, you will turn negative values into very large positive values. Do not do this. Instead, keep it as an `i64` and specifically check for negative values along with `i > size`. To avoid warnings, cast the actual `size` to an `i64` instead.

compiler+runtime/src/cpp/jank/runtime/obj/transient_vector.cpp
```diff
@@ -49,6 +49,43 @@ namespace jank::runtime::obj
...
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+
+  transient_vector_ref transient_vector::dissoc_in_place(object_ref const key)
+  {
+    throw std::runtime_error{ "Unsupported dissoc operation!" };
+  }
+
+#pragma clang diagnostic pop
```
jeaye: Instead of adding all of this ceremony for a warning, just remove the parameter name when it's unused. :)

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/obj/transient_vector.cpp
shantanu-sardesai: Hmm... will I ever learn the entirety of the C++ language I never know 😂.

compiler+runtime/src/cpp/jank/runtime/obj/transient_vector.cpp
shantanu-sardesai: I was lazy here and just copied the [implementation](https://github.com/shantanu-sardesai/jank/blob/b8bb921d4f5d32da05a7ea64bd5db6d0889167b7/compiler+runtime/src/cpp/jank/runtime/obj/persistent_vector.cpp#L265) of the `assoc` in the `persistent_vector`. Shall I update it as well?

shantanu-sardesai: Done.

jeaye (approved):

jeaye:
Nice work!

PR #735, opened 2026-04-15
Add custom IR

jeaye:
This PR removes our previous AST-based codegen (C++ and LLVM IR) and instead adds a new SSA-based custom jank IR, which has its own C++ codegen. This sets us up for writing optimization passes prior to codegen. jank's IR is semantically at the Clojure level, which is significantly higher level than something like LLVM IR. This allows us to perform optimizations, similarly, at the level of Clojure's semantics, which LLVM will not be able to do.


jeaye:
@kylc Sorry for the ping. Looks like the Nix build is failing with a new error, failing to find `memcpy.inline` at runtime. I've done some light research into this and it looks like several others have seen the same thing on Nix, but I haven't found any clear indicators for what we can actually do. Is this something you've seen before? If not, is it something you have the capacity to investigate in the short term? I've not been able to reproduce this on my local NixOS setup.

If you're not available, that's ok. I might need to disable the Nix build, though, since I have too many other things right now that are higher priority than Nix builds.

jeaye:
@ikappaki Any idea why `test/jank/cpp/if/pass-condition-convertible-to-bool.jank` would lead to a parsing failure on Windows?

ikappaki:
> @ikappaki Any idea why `test/jank/cpp/if/pass-condition-convertible-to-bool.jank` would lead to a parsing failure on Windows?

I’ll take a look. Nothing obvious stands out, and it doesn’t fail when run in isolation.

kylc:
@jeaye See https://github.com/jank-lang/jank/pull/753 for a Nix build workaround.

I also see the `pass-condition-convertible-to-bool.jank` test failing on NixOS.

ikappaki:
> > @ikappaki Any idea why `test/jank/cpp/if/pass-condition-convertible-to-bool.jank` would lead to a parsing failure on Windows?
> 
> I’ll take a look. Nothing obvious stands out, and it doesn’t fail when run in isolation.

Hi @jeaye, The root issue seems to occur earlier: `fail-condition-not-convertible-to-bool.jank` throws an exception, and due to missing exception support on Windows, the stack isn’t properly unwound. This corrupts state and causes later tests to fail. I’ve always seen this pattern while porting the tests, and is quite misleading.

Could you please add `pass-condition-convertible-to-bool.jank`  to the `jit_windows_skips.txt`, the suite passes locally for me.

Thanks

PR #737, opened 2026-04-17
Add tap support

djblue:
This PR is my first approximation at how to implement tap support in jank.

Not sure if we want `array_blocking_queue` or if it's in the right place in the code base. Also, I couldn't find a better way to initialize the tapq outside of `cpp/raw` and using a future is a hack because I couldn't figure out how to start a thread via interop 😂


djblue:
@jeaye I've been using this for a few days and while it mostly works, it does appear to deadlock which I assume is due to the usage of future. I noticed this when trying to `tap>` from the nrepl server which I think also uses futures. Is there an easy way to spawn a cpp thread which can run a jank function? Also, does the POC look reasonable to you?

jeaye:
> @jeaye I've been using this for a few days and while it mostly works, it does appear to deadlock which I assume is due to the usage of future. I noticed this when trying to `tap>` from the nrepl server which I think also uses futures.

Hm, there are some issues here which make this unsafe for GC use. That may be related to the deadlocks or may be unrelated.

>  Is there an easy way to spawn a cpp thread which can run a jank function?

`clojure.core/future` uses `jank::runtime::future`, which is the way. Threads need GC registration, so one cannot just make a C++ thread and use it normally. Furthermore, `jank::runtime::future` copies dynamic bindings over, which is an important aspect of thread creation in Clojure.

> Also, does the POC look reasonable to you?

As a PoC, sure. The queue will need to use `native_vector` to ensure GC compatibility. I don't think it belongs in the `jtl`, since it relies on jank's runtime objects and the GC. The rest of my feedback would mainly be stylistic.

The biggest concern here is the deadlocks. I'm not aware of any deadlocks with `future`, but if we have some then that's a big deal.

djblue:
I updated the code to use `native_vector` and moved the code into `jank::runtime::obj`. After doing so, `tap>` works just fine from the nrepl server code. So my assumption about it being related to future was incorrect, and it was more likely related to GC as you suggested. Does that mean using future in the implementation is fine to keep?

jeaye:
> I updated the code to use `native_vector` and moved the code into `jank::runtime::obj`. After doing so, `tap>` works just fine from the nrepl server code. So my assumption about it being related to future was incorrect, and it was more likely related to GC as you suggested. Does that mean using future in the implementation is fine to keep?

Yeah, `future` is preferable here. Glad to hear it's working!

compiler+runtime/include/cpp/jank/runtime/obj/array_blocking_queue.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <mutex>
+#include <condition_variable>
+
+#include <jank/type.hpp>
+#include <jank/runtime/object.hpp>
+
+namespace jank::runtime::obj
+{
+  using namespace jank::runtime;
```
jeaye: Putting this in a header means that everyone who includes it is going to be impacted. This is not good practice in C++. It's only desired here because you've defined all fns within the struct, but we should move them to a cpp file instead. We need to have the smallest header footprint we can, since it not only affects jank's compile-times, it also affects jank's startup time.

compiler+runtime/include/cpp/jank/runtime/obj/array_blocking_queue.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+  struct array_blocking_queue
```
jeaye: This should go into `jank::runtime::detail`, alongside `native_array_map`. We should call it `native_array_blocking_queue`.

compiler+runtime/include/cpp/jank/runtime/obj/array_blocking_queue.hpp
```diff
@@ -0,0 +1,53 @@
...
+  {
+    jank::native_vector<object_ref> buffer;
```
jeaye: No need for `jank::` here, since we're within the jank namespace.

compiler+runtime/include/cpp/jank/runtime/obj/array_blocking_queue.hpp
```diff
@@ -0,0 +1,53 @@
...
+    size_t head{};
+    size_t tail{};
+    size_t size{};
+    size_t capacity{};
```
jeaye: Please use `usize`.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -7429,23 +7430,20 @@ fails, attempts to require sym's namespace and retries."
...
-  ;;                (tap x)
-  ;;                (catch Throwable ex)))
-  ;;            (recur))
-  ;;         "clojure.core/tap-loop")
-  ;;    (.setDaemon true)
-  ;;    (.start)))
-  nil)
+  (delay
+    (future
+      (let [t (cpp/.take (cpp/unbox (:* cpp/jank.runtime.obj.array_blocking_queue) tapq))
```
jeaye: 1. There's no need for `cpp/` for `.take`.
2. Thereps no need for `cpp/` within the type DSL.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -7429,23 +7430,20 @@ fails, attempts to require sym's namespace and retries."
 
 (defonce ^:private tapset (atom #{}))
 
-;; (defonce ^:private ^java.util.concurrent.ArrayBlockingQueue tapq (java.util.concurrent.ArrayBlockingQueue. 1024))
+(defonce ^:private tapq
+  (cpp/box (cpp/& (cpp/jank.runtime.obj.array_blocking_queue. 1024))))
```
jeaye: This is not safe, since you're stack-allocating a value, getting its address, and hanging onto it. Instead, use `(cpp/box (cpp/new jank.runtime.obj.array_blocking_queue 1024))`.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -7468,8 +7466,9 @@ fails, attempts to require sym's namespace and retries."
   false if not (dropped)."
   [x]
   (force tap-loop)
-  ;; (.offer tapq (if (nil? x) ::tap-nil x))
-  (throw "TODO: port tap>"))
+  (cpp/.offer
```
jeaye: 1. No need for `cpp/` with `.offer`
2. No need for `cpp/` in the type DSL

jeaye (changes requested):

compiler+runtime/src/jank/clojure/core.jank
djblue: Ohh, my bad. I assumed the `.` was shorthand for `new`, not stack allocation. When switching to heap allocation, I get the following error:

```
input_line_132:2:150: error: call to implicitly-deleted copy constructor of 'jank::runtime::detail::native_array_blocking_queue'
    2 | static auto const finalizer_34642{ [](void * const obj, void *){reinterpret_cast<T*>(obj)->~T();} };auto cpp_new_34641{ new (UseGC, finalizer_34642) jank::runtime::detail::native_array_blocking_queue{ cpp_ctor_34643 } };auto cpp_box_34640{ jank::runtime::make_box<jank::runtime::obj::opaque_box>(cpp_new_34641, "jank::runtime::detail::native_array_blocking_queue*") };
      |                                                                                                                                                      ^                                                 ~~~~~~~~~~~~~~~~~~
/home/chris/repos/jank/compiler+runtime/include/cpp/jank/runtime/detail/native_array_blocking_queue.hpp:24:16: note: copy constructor of 'native_array_blocking_queue' is implicitly deleted because field 'mutex' has a deleted copy constructor
   24 |     std::mutex mutex;
      |                ^
/usr/lib64/gcc/x86_64-pc-linux-gnu/15.2.1/../../../../include/c++/15.2.1/bits/std_mutex.h:109:5: note: 'mutex' has been explicitly marked deleted here
  109 |     mutex(const mutex&) = delete;
      |     ^
error: Parsing failed.
─ internal/codegen-failure ─────────────────────────────────────────────────────
error: Unable to compile C++ source. 
```

jeaye: The `.` doesn't actually do anything and is entirely optional. What you've found here is a codegen bug, since valid jank code (i.e. fully analyzed) should never result in bad codegen.

We can get around this by either defining the value in C++ or using `cpp/raw` again. I've made a note of it locally, so it can be fixed. Please leave a TODO next to it, to remove the `cpp/raw` when we can.

djblue: @jeaye what does it look like to create the value via `cpp/raw` and store it in the var?

jeaye: Don't mix `cpp/raw` and vars. Use one or the other. The way you had it originally, just defining a C++ global, will work fine. I'd put it in `clojure::core`, though, instead of `jank::runtime`.

djblue: When I tried create a C++ global via `cpp/raw`, I got linking errors. What did work for me was creating a constructor style function `create_native_array_blocking_queue`.

jeaye (approved):
Works for me. Ready to merge?

djblue:
@jeaye, yeah I think this is good to go 👍

jeaye:
Thanks, Chris!

PR #739, opened 2026-04-18
Update nREPL output to be compatible with Calva jackin

djblue:
It appears that [Calva matches against the exact format of the JVM nREPL implementation output](https://github.com/BetterThanTomorrow/calva/blob/6a18844915e4e77f9a89ac6cc97ef44aacb595b9/src/extension-test/unit/nrepl/jack-in-port-detection-test.ts#L7-L15) to assist with setting up a REPL via jackin. There is another problem which I only have a workout for which involves stdout buffering that I get around via: `script -qc 'jank repl' /dev/null`, still looking into it though.


jeaye:
Oh, that's surprising! Will you please leave a comment above the println saying the same thing? That way this won't be changed again.

jeaye:
Thanks, Chris!

PR #740, opened 2026-04-18
Allow nREPL clients to set source info during eval

djblue:
Tools such as [`portal.console`](https://github.com/djblue/portal/blob/master/src/portal/console.cljc#L45) and [`taoensso.timbre`](https://github.com/taoensso/encore/blob/master/src/taoensso/encore.cljc#L2412) leverage source info during macro expansion via metadata. When working interactively via nREPL, and eval'ing single forms, preserving this source info keeps these tools working.

### DIFF
--- a/compiler+runtime/include/cpp/jank/read/lex.hpp
+++ b/compiler+runtime/include/cpp/jank/read/lex.hpp
@@ -256,7 +256,7 @@ namespace jank::read::lex
 
     processor(jtl::immutable_string_view const &f);
     processor(jtl::immutable_string_view const &f, usize offset);
-    processor(jtl::immutable_string_view const &f, usize line, usize col);
+    processor(jtl::immutable_string_view const &f, source_position const &p);
 
     jtl::result<token, error_ref> next();
     jtl::result<codepoint, error_ref> peek(usize const ahead = 1) const;
--- a/compiler+runtime/include/cpp/jank/read/lex.hpp
+++ b/compiler+runtime/include/cpp/jank/read/lex.hpp
@@ -256,6 +256,7 @@ namespace jank::read::lex
 
     processor(jtl::immutable_string_view const &f);
     processor(jtl::immutable_string_view const &f, usize offset);
+    processor(jtl::immutable_string_view const &f, usize line, usize col);
 
     jtl::result<token, error_ref> next();
     jtl::result<codepoint, error_ref> peek(usize const ahead = 1) const;
--- a/compiler+runtime/src/jank/jank/nrepl/server/eval.jank
+++ b/compiler+runtime/src/jank/jank/nrepl/server/eval.jank
@@ -57,9 +57,9 @@ namespace jank::nrepl::server::eval
   ;; C++ and catch the exceptions there.
 
   (let [p (cpp/jank.read.source_position.
-           (cpp/cast cpp/jtl.usize 0)
-           (cpp/cast cpp/jtl.usize line)
-           (cpp/cast cpp/jtl.usize col))]
+           (cpp/jtl.usize 0)
+           (cpp/jtl.usize line)
+           (cpp/jtl.usize col))]
     (cpp/jank.nrepl.server.eval.safe_eval code p)))
 
 (comment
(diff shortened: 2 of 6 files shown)


compiler+runtime/src/jank/jank/nrepl/server/handler/eval.jank
```diff
@@ -14,16 +14,17 @@
 (binding [*ns* default-ns]
   (clojure.core/refer-clojure))
 
-(defn- do-eval [ns code inspect?]
+(defn- do-eval [ns code file-path line col inspect?]
```
djblue, line 21: I'm not a fan of how many params this fn has to take now, but I didn't want to refactor too much 😂

compiler+runtime/include/cpp/jank/read/lex.hpp
```diff
@@ -256,6 +256,7 @@ namespace jank::read::lex
 
     processor(jtl::immutable_string_view const &f);
     processor(jtl::immutable_string_view const &f, usize offset);
+    processor(jtl::immutable_string_view const &f, usize line, usize col);
```
jeaye: We should not be working with line/col separately. We have `read::source_position` for fixed positions and `read::source` for ranges.

djblue: @jeaye I updated the code to use `read::source_position`. Please let me know if I did it incorrectly.

compiler+runtime/src/jank/jank/nrepl/server/eval.jank
```diff
@@ -56,7 +56,11 @@ namespace jank::nrepl::server::eval
   ;; error will crash the server. Instead we create a safe evaluation wrapper in
   ;; C++ and catch the exceptions there.
 
-  (cpp/jank.nrepl.server.eval.safe_eval code))
+  (let [p (cpp/jank.read.source_position.
+           (cpp/cast cpp/jtl.usize 0)
```
jeaye: `(cpp/jtl.usize 0)` should be all that's needed. Same with the other two.

compiler+runtime/src/jank/jank/nrepl/server/eval.jank
```diff
@@ -56,7 +56,11 @@ namespace jank::nrepl::server::eval
...
+           (cpp/cast cpp/jtl.usize line)
+           (cpp/cast cpp/jtl.usize col))]
```
jeaye: So, we're setting offset to zero, but we're giving non-zero line/col potentially. This means we have an only partially constructed source position. The lexer works based on the offset, not the other two, so it's still going to start at the beginning. What is this actually intending to do?

djblue: So I think the following nREPL message will help clarify my intention:

```clojure
{"op" "eval"
 "ns" "user"
 "line" 23
 "column" 3
 "file" "/home/chris/repos/portal/dev/user.clj"
 "code" "(tap> ::foo)"}
```

When you eval a single expression in your editor, the nREPL client only sends the snippet of code you are evaluating to the server along with where in the larger file the code belongs. So, to preserve the line/column info we need to jump to the correct file position, but for analyzing tokens, we need to start from the beginning.

For reference, the jvm nREPL implementation does something similar [here](https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/middleware/interruptible_eval.clj#L39-L44).

jeaye: Ok, so nREPL messages include source info for evals, but it doesn't include the offset. So our offset will be `0`, but the line will be `23` and column will be `3`. Then, when we lex the code, we'll increase our offset and the line/col info will build from where it started.

Then this information becomes important when we try to show an error message, or is it some other scenario where you need the adjusted source info?

djblue: Yeah, I think it's important anytime you have tooling that reports source location info. For me, this shows up when I use logging type macros. This will also help keep the source metadata on vars be accurate.

jeaye: Ok, I understand now. Thanks for explaining. Would you do me a solid and put a comment block above `do-eval` which says everything you just explained to me? With that, I'm good to merge.

djblue: @jeaye Let me know if the comment I added is sufficient.

jeaye:
Thanks, Chris!

PR #741, opened 2026-04-19
Add support for `*print-meta*`

djblue:


compiler+runtime/src/cpp/jank/runtime/core/to_string.cpp
```diff
@@ -18,20 +20,47 @@ namespace jank::runtime
...
+            }
+          }
+
+          buff(typed_o->to_code_string());
+        },
+        o);
+    }
+    else
+    {
+      visit_object([&](auto const typed_o) { buff(typed_o->to_code_string()); }, o);
```
djblue, line 56: I wonder if `to_code_string` should have an overload that also accepts a buffer?

compiler+runtime/src/cpp/jank/runtime/core/to_string.cpp
```diff
@@ -18,20 +20,47 @@ namespace jank::runtime
...
   {
     buff(obj::character{ ch }.to_code_string());
   }
 
   void to_code_string(object_ref const o, jtl::string_builder &buff)
   {
-    auto const value{ visit_object([](auto const typed_o) { return typed_o->to_code_string(); },
-                                   o) };
-    buff(value);
+    auto const print_meta(make_box<obj::symbol>("clojure.core", "*print-meta*"));
```
jeaye, line 30: These two `to_code_string` functions are no longer actually needed. We now have a `to_code_string` on the object. In fact, we don't even need to visit the object anymore to be able to get that, since it's now a virtual function.

So these global `to_code_string` functions should be considered deprecated. The vast majority of the time, when we do `to_code_string`, we will NOT be using these functions. So we need to handle this differently. Perhaps we will need to handle this in each object's `to_code_string` function.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/core/to_string.cpp
djblue: In looking through some of the method implementations for `to_code_string`, I found a few that reference these functions, such as [`range.cpp`](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/runtime/obj/range.cpp#L239-L242). Should those functions be updated to call the method instead?

jeaye: Yeah, we ultimately just want to remove those old uses entirely. jank is in a transitional state between the closed object model and the open object model right now. `visit`, as well, is the old way of doing things, but we need to build out the virtual fns on `object` more in order to replace it.

djblue:
Closing since this isn't the approach we want to take for printing metadata.

PR #744, opened 2026-04-20
lein-jank: add :frameworks and :runtime-flags support to build-declarative-flag

rgkirch:


lein-jank/src/leiningen/jank/core.clj
```diff
@@ -57,6 +63,8 @@
                      (build-declarative-flags project)
                      compiler-args
                      ["--"]
+                     (flatten (build-declarative-flag :runtime-flags (get-in project [:jank :runtime-flags] [])))
+                     (flatten (build-declarative-flag :frameworks    (get-in project [:jank :frameworks] [])))
```
jeaye, line 67: I don't see why these two lines are needed. We're already calling `build-declarative-flags` above, which will handle all flags. We don't want these flags to come after the `--`, since the flags which come after that just end up being input to the running program (i.e. `*command-line-args*`).

Does this actually fix your issue for you? I don't see how it would.

lein-jank/src/leiningen/jank/core.clj
```diff
@@ -43,12 +43,18 @@
     :linked-libraries
     (map (fn [v] (str "-l" v)) value)
 
+    :frameworks
+    (mapcat (fn [v] ["-framework" v]) value)
+
+    :runtime-flags
+    value
```
jeaye, line 50: I'm ok adding this, if we call it `:extra-flags`. These flags are not just for the runtime. They can also be for the compiler.

jeaye (changes requested):
Thanks for the PR!

rgkirch:
Superseded by https://github.com/jank-lang/jank/pull/748
Thank you for your consideration!

PR #745, opened 2026-04-21
Fix use-fixtures silently dropped due to (:ns (meta var)) nil

rgkirch:
group-by (comp :ns meta) collapses all vars under nil because var
  metadata is lossy in jank (#197), so join-fixtures returns
  default-fixture and :once/:each fixtures never run. Swap in the
  existing -workaround-var->ns helper, matching the shim style already
  used at lines 615/635/659 for the same #197 gap.

  Repro: https://github.com/.../jank-bugs/repros/use-fixtures-noop


jeaye:
Thanks for the PR!

PR #746, opened 2026-04-21
Fix silent SIGSEGV when -l flag names a missing library

rgkirch:
load_dynamic_libs was called from processor::processor(). On failure it
  threw error::system_failure, which unwound through the constructor and
  triggered CppInternal::Interpreter::~Interpreter in a partially-torn-down
  state, corrupting LLVM state and causing SIGSEGV in llvm_shutdown().

  Move load_dynamic_libs to context::context() body after jit_prc is fully
  constructed. On failure, print a diagnostic to stderr and call std::exit(1)
  to avoid running the interpreter destructor during stack unwinding.

### DIFF
--- a/compiler+runtime/src/cpp/jank/jit/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/jit/processor.cpp
@@ -239,7 +239,8 @@ namespace jank::jit
     auto const &load_result{ load_dynamic_libs(util::cli::opts.libs) };
     if(load_result.is_err())
     {
-      throw error::system_failure(load_result.expect_err().c_str());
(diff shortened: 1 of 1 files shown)


jeaye:
Thanks for the PR! It pointed me in the right direction. Ultimately, the issue was that we were owning our `clang::Interpreter` in a `std::unique_ptr`, but it wasn't ours to own. So it was being deleted twice during unwinding. I addressed this by removing ownership by just using a `jtl::ptr` instead.

The commit is here: https://github.com/jank-lang/jank/commit/3bebc3b1b664001b5cb32c46e585a8e6b65baac2

PR #750, opened 2026-04-24
Fix nREPL capture when $TEMP environment variable not set

djblue:
I was trying to use the nREPL server on my native Linux install, which I guess doesn't set the `$TEMP` environment variable. It appears [`std::filesystem::temp_directory_path`](https://en.cppreference.com/cpp/filesystem/temp_directory_path) will do the following:

> On POSIX systems, the path may be the one specified in the environment variables TMPDIR, TMP, TEMP, TEMPDIR, and, if > none of them are specified, the path "/tmp" is returned.

### DIFF
--- a/compiler+runtime/src/jank/jank/nrepl/server/capture.jank
+++ b/compiler+runtime/src/jank/jank/nrepl/server/capture.jank
@@ -3,7 +3,8 @@
             "filesystem")
   (:require [jank.nrepl.server.util :refer [close dup dup2 unlink]]))
 
-(def temp-file-template (str (cpp/getenv #cpp "TEMP") "/" "jank-nrepl-XXXXXX"))
+(def temp-file-template
+  (str (.string (cpp/std.filesystem.temp_directory_path)) "/" "jank-nrepl-XXXXXX"))
 
 (defn- make-temp-file []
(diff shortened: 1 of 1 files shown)


djblue:
@ikappaki I assume you are a windows user of jank, if so, do you mind checking if this PR breaks your setup? :pray:

ikappaki:
Hi @djblue, I can confirm the change works on windows:

_compiler+runtime/src/jank/jank/nrepl/server/capture.jank_:
```clojure
(ns jank.nrepl.server.capture
  (:include "cstdio"
            "filesystem")
  (:require [jank.nrepl.server.util :refer [close dup dup2 unlink]]))

(def temp-file-template
  (str (.string (cpp/std.filesystem.temp_directory_path)) "/" "jank-nrepl-XXXXXX"))
...
```

```clojure
$ build/jank repl
nREPL server started on port 51022 on host 127.0.0.1 - nrepl://127.0.0.1:51022
user=> (require '[jank.nrepl.server.capture :as c])
nil

user=> c/temp-file-template
"D:\\msys64\\jank-20\\tmp/jank-nrepl-XXXXXX"

```

Thanks

jeaye:
Thanks, Chris! Thanks for checking on Windows, ikappaki!

PR #751, opened 2026-04-25
Allow `(reset! (atom nil) nil)`

djblue:
I was hitting this check when trying to clear my atoms. I don't think we need this check as any value can be reset into an atom as far as I'm aware.

Jank
```clojure
% jank repl
user=> (reset! (atom nil) nil)
Assertion failed! o.is_some()
```

Clojure
```clojure
% clj
Clojure 1.12.4
user=> (reset! (atom nil) nil)
nil
```

This closes #724.


ikappaki:
(Also raised as an issue here: #724 )

jeaye:
Thanks, Chris!

PR #752, opened 2026-04-25
Include stdio info to nREPL client on errors

djblue:
I was struggling to debug `cpp/raw` errors which are only available via [stderr](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/jit/processor.cpp#L346-L347). This update allows the nREPL client to provide why you got the error: `"Unable to compile C++ source."`.


jeaye:
Nice! That should be a good improvement.

PR #753, opened 2026-04-25
Disable fortification in Nix build to avoid linker errors.

kylc:
Tested locally, fixes `nix build`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -111,6 +111,15 @@
               )
             '';
 
+            # Disable _FORTIFY_SOURCE to prevent linker errors for substituted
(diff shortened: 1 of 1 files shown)


jeaye:
Oh, interesting! I recall we have something similar in our flake:

```nix
          # Nix assumes fortification by default, but that fails with debug builds.
          # Since this shell is used for development, we disabled fortification. It's
          # still enabled for our release builds in build.nix.
          # https://github.com/NixOS/nixpkgs/issues/18995
          hardeningDisable = ["fortify"];
```

jeaye:
Thank you, Kyle!

PR #756, opened 2026-05-01
Enable `disj` & `disj!` tests

shantanu-sardesai:
Brings us to **117 of 231 tests**. Will try to get more out of the way during this weekend 💪.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -219,11 +219,15 @@ namespace jank::runtime
     {
       return expect_object<obj::transient_hash_set>(coll)->disjoin_in_place(o);
     }
+<<<<<<< HEAD
     else if(coll.get_type() == object_type::transient_array_map)
     {
       return expect_object<obj::transient_array_map>(coll)->dissoc_in_place(o);
     }
     else if(coll.get_type() == object_type::transient_sorted_set)
+=======
+    else if(coll->type == object_type::transient_sorted_set)
+>>>>>>> 7ce8cde9 (Update `disj!` tests)
     {
       return expect_object<obj::transient_sorted_set>(coll)->disjoin_in_place(o);
     }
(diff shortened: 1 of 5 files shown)


shantanu-sardesai:
@jeaye
Is this a Windows skippable test?

compiler+runtime/src/cpp/jank/runtime/obj/transient_hash_set.cpp
```diff
@@ -2,6 +2,8 @@
 #include <jank/runtime/obj/persistent_hash_set.hpp>
 #include <jank/runtime/obj/persistent_vector.hpp>
 #include <jank/runtime/obj/nil.hpp>
+#include <jank/runtime/obj/number.hpp>
+#include <jank/runtime/rtti.hpp>
```
jeaye: Why were these needed?

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -516,14 +512,18 @@ namespace jank::runtime
...
+        [](auto const typed_s, object_ref const o) -> object_ref { return typed_s->disj(o); },
+        s,
+        o);
     }
     else
     {
-      throw std::runtime_error{ util::format("not disjoinable: {}", runtime::to_code_string(s)) };
+      throw std::runtime_error{ util::format("not disjoinable: {} - {}",
+                                             object_type_str(s->type),
+                                             runtime::to_code_string(s)) };
```
jeaye: It's a bad habit (that I started) to print values in errors, since we never know when one of them might be an infinite seq. Printing the type, and only the type, is a much better move.

jeaye:
> @jeaye Is this a Windows skippable test?

Perhaps we skip the clojure-test-suite entirely on Windows, if we're not already doing that. Exception unwinding from JIT compiled code, on Windows, doesn't currently work. That'll break basically every test which has negative cases that throw. In the bb script, we can just check for Windows and skip (with a warning).

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
shantanu-sardesai: I added this to debug but forgot to remove 🙈.

shantanu-sardesai:
Alright, will disable Windows tomorrow morning.

jeaye:
Nice work!

PR #761, opened 2026-05-09
Add source_location to IR

djblue:
I put together a quick prototype of how jank could emit source location to help debuggers navigate `.jank` source.

It appears to kinda be working, but I was only able to pass through a limited amount of source locations during IR generation. I think only a certain number of expressions have source location information.


shantanu-sardesai:
Curious, you mention it is already kind of working, which debuggers were you able to make it work with? I'm a bit surprised that this is all it required for it 🤔.

jeaye:
> Curious, you mention it is already kind of working, which debuggers were you able to make it work with? I'm a bit surprised that this is all it required for it 🤔.

This should work with any native debugger, since Clang will generate source info based on the generated pragmas. Chris and I have been discussing this separately, which is what led to this PR.

jeaye:
This seems like a reasonable approach, Chris. Are you open to building it out more? In particular:

1. If each AST expression has an `object_ref` for the underlying form, we could pull our source from there
  a. Seems like this could go into `analyze::expression` directly, so we don't even need to visit for it
2. I think we'll want to print the `source-location` instructions, just as we print everything else 
3. The creation of debug info should likely be behind a `--debug` flag
4. The `lein-jank` plugin should support `:debug true` as an option and the template should specify it for default builds

djblue:
@shantanu-sardesai I've only tested it with vscode. You can see what it looks like here: https://github.com/jank-lang/jank/pull/767.
@jeaye I am interested in building it out more 👍

djblue:
@jeaye Before I take this too far, I just want to confirm that the following commits  https://github.com/jank-lang/jank/pull/761/changes/b1f48464313947e398cc3a1d4bc4762ae50c705c https://github.com/jank-lang/jank/pull/761/changes/d6073986dd37c55257a111290e85b4dcca9637d1 https://github.com/jank-lang/jank/pull/761/changes/6a6e9c63e4ce8182630a2ec9ed00755c6deb10de are what you are expecting for step 1?

jeaye:
> @jeaye Before I take this too far, I just want to confirm that the following commits [b1f4846](https://github.com/jank-lang/jank/commit/b1f48464313947e398cc3a1d4bc4762ae50c705c) [d607398](https://github.com/jank-lang/jank/commit/d6073986dd37c55257a111290e85b4dcca9637d1) [6a6e9c6](https://github.com/jank-lang/jank/commit/6a6e9c63e4ce8182630a2ec9ed00755c6deb10de) are what you are expecting for step 1?

Yeah, we should be able to pass in a form for every AST expression. It should ideally be required, so we should update existing ctors to take it, rather than add a new ctor. Then we can update the creation of every AST expression. If ever there are any which don't reasonably have a form, that's ok, but I'd want to know why.

Once every AST expr has a form, we can hand that info down to the IR by having the base IR `instruction` store a `source_position`. The default source should be `unknown()`. We should then update the IR processor builder to have some state about the current source position. If we process an AST expression with a different source position, we automatically insert a new `:source-location` instruction. This can be done implicitly, by the builder, rather than explicitly for each `gen` function. This is how we handle `ret`, if the expression is in tail position. It allows us to not think about source positions when we're stitching together the IR, since the builder will just handle it for us.

By the time we get to codegen, our `:source-location` instructions will contain the minimal set of state changes, with no duplicate instructions. So there will be no logic we need to do there aside from just rendering them out.

djblue:
@jeaye I was able to thread the source form to ir gen via expression, please let me know if I did it wrong.

With the new approach, the source info we are producing doesn't seem to be working all the time. I still need to figure out why. I don't think I implemented the ir gen bit correctly 🤔 

For reference, https://github.com/djblue/jank/commit/5c2d366249aa25ac3cb4b95b0d11891c1c5b1e4c is the previous implementation that appears to produce more accurate source information.

jeaye:
> @jeaye I was able to thread the source form to ir gen via expression, please let me know if I did it wrong.
> 
> With the new approach, the source info we are producing doesn't seem to be working all the time. I still need to figure out why. I don't think I implemented the ir gen bit correctly 🤔
> 
> For reference, [djblue@5c2d366](https://github.com/djblue/jank/commit/5c2d366249aa25ac3cb4b95b0d11891c1c5b1e4c) is the previous implementation that appears to produce more accurate source information.

This looks great to me. Everything as I expected it would look. I'm not sure why it's not always working, but I have a guess. Some of the forms you're using are like `name` and `sym`, but those are not necessarily source forms. Ideally we want to be sure we're always pulling from the source forms.

djblue:
Ok, so it turns out that the `#line` directive will set an initial line / file, but every `C++` line after will increment the line counter. So the easiest thing to do is emit the `#line` directive for every line of `C++` we generate. In order to do so, I updated instruction to include source location, then in [cpp_processor.cpp](https://github.com/jank-lang/jank/pull/761/changes#diff-5fd09644fd7573aa33fa96b4b04a097f27fedcfa2f23a287bad516ff780cfcbaR1845-R1853) we can always print the location for every instruction.

With this change, the debugger is working exactly as I would expect 👍 Setting breakpoints also seems to work well 🙌

jeaye:
Great job, Chris!

PR #763, opened 2026-05-09
Math epic - Part 1

shantanu-sardesai:
~~Requires a Clojure test suite repoint (https://github.com/jank-lang/clojure-test-suite/pull/875)~~. Done.

Brings us to **125 of 231 tests**. Goal is to get to 145, covering all "math" `clojure.core` functions with tests in the test suite.

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
@@ -1062,13 +1062,6 @@ namespace jank::runtime
     return static_cast<i64>(l);
   }
 
-  i64 to_long(object_ref const l);
-  i64 to_long(obj::nil_ref const l);
-  i64 to_long(obj::integer_ref const l);
-  i64 to_long(obj::real_ref const l);
-  i64 to_long(i64 const l);
-  i64 to_long(f64 const l);
-
   f64 to_real(object_ref const o);
 
   template <typename T>
--- a/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
@@ -1062,6 +1062,13 @@ namespace jank::runtime
     return static_cast<i64>(l);
   }
 
+  i64 to_long(object_ref const l);
+  i64 to_long(obj::nil_ref const l);
+  i64 to_long(obj::integer_ref const l);
+  i64 to_long(obj::real_ref const l);
+  i64 to_long(i64 l);
+  i64 to_long(f64 l);
+
   f64 to_real(object_ref const o);
 
   template <typename T>
--- a/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
@@ -1110,8 +1110,8 @@ namespace jank::runtime
   bool is_nan(object_ref const o);
   bool is_infinite(object_ref const o);
 
-  i64 parse_long(object_ref const o);
-  f64 parse_double(object_ref const o);
+  object_ref parse_long(object_ref const o);
+  object_ref parse_double(object_ref const o);
 
   bool is_big_integer(object_ref const o);
   obj::big_integer_ref to_big_integer(object_ref const o);
--- a/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
@@ -1066,8 +1066,8 @@ namespace jank::runtime
   i64 to_long(obj::nil_ref const l);
   i64 to_long(obj::integer_ref const l);
   i64 to_long(obj::real_ref const l);
-  i64 to_long(i64 l);
-  i64 to_long(f64 l);
+  i64 to_long(i64 const l);
+  i64 to_long(f64 const l);
 
   f64 to_real(object_ref const o);
 
--- a/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
@@ -406,9 +406,23 @@ namespace jank::runtime
   {
     return visit_number_like(
       [](auto const typed_l) -> object_ref {
-        auto const ret{ make_box(typed_l->data + 1ll) };
-        object_ref r{ ret };
-        return r;
+        using T = typename decltype(typed_l)::value_type;
+
+        if constexpr(jtl::is_any_same<T, obj::integer, obj::small_integer>)
+        {
+          i64 res{};
+
+          if(__builtin_add_overflow(typed_l->data, 1ll, &res))
+          {
+            throw make_box("Overflow on increment.").erase();
+          }
+
+          return make_box<obj::integer>(res);
+        }
+        else
+        {
+          return make_box(typed_l->data + 1ll);
+        }
       },
       l);
   }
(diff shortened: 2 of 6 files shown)


compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -417,9 +417,23 @@ namespace jank::runtime
...
+        if constexpr(jtl::is_any_same<T, obj::integer, obj::small_integer>)
+        {
+          i64 res{};
+
+          if(__builtin_add_overflow(typed_l->data, 1ll, &res))
+          {
+            throw make_box("Overflow on increment.").erase();
+          }
+
+          return make_box<obj::integer>(res);
```
jeaye: Do not use this overload of `make_box`. It forces us to heap allocate integers, which is a HUGE performance hit. Just use `make_box(res)` and we'll get a `small_integer` if we can.

The same applies to EVERY usage in this PR. Do not do this.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -961,6 +993,36 @@ namespace jank::runtime
...
+    return to_int(l);
+  }
+
+  i64 to_long(i64 l)
+  {
+    return to_int(l);
+  }
+
+  i64 to_long(f64 l)
+  {
```
jeaye: These params can be const.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1052,12 +1144,45 @@ namespace jank::runtime
...
+      size_t parsed_upto{};
+      f64 parsed_double{};
+
+      try
+      {
+        parsed_double = std::stod(typed_o->data, &parsed_upto);
+      }
+      catch(...)
+      {
+        // FIXME: `jank_nil` get's converted to `0.00000`
```
jeaye: I would gather that you forgot to change the declaration of the function, so the compiler thinks it will be returning a `f64`, in another translation unit, but you return `jank_nil` instead.

This is not the first time a mentee has done this. Might've even been you last time! Always keep your declarations and your definitions in sync.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
shantanu-sardesai: Nope, as you can see in the `math.hpp`, I've updated the `return` type 🤔.

shantanu-sardesai: Would jank even compile if that were the case?

jeaye: Yes. Be sure that it's not forward declared somewhere else with the old signature.

shantanu-sardesai: Seem like it was not playing well with the way we were `intern_fn`'ing the same function in the `core_native`. Don't know why that doesn't work well but since we probably don't need this `intern_fn` call anyways I just removed it instead of investigating.

compiler+runtime/include/cpp/jank/runtime/core/math.hpp
```diff
@@ -1038,6 +1038,13 @@ namespace jank::runtime
   i64 to_int(i64 l);
   i64 to_int(f64 l);
 
+  i64 to_long(object_ref const l);
+  i64 to_long(obj::nil_ref const l);
+  i64 to_long(obj::integer_ref const l);
+  i64 to_long(obj::real_ref const l);
+  i64 to_long(i64 const l);
+  i64 to_long(f64 const l);
+
```
jeaye, line 1064: There is no need for these fns. They do the same thing as `to_int`. Let's just use `to_int`.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1038,12 +1100,42 @@ namespace jank::runtime
...
+  object_ref parse_long(object_ref const o)
   {
     auto const typed_o{ dyn_cast<obj::persistent_string>(o) };
     if(typed_o.is_some())
     {
-      return std::stoll(typed_o->data);
+      auto const str_size{ static_cast<size_t>(typed_o->data.size()) };
+      auto const max_limit{ std::numeric_limits<size_t>::max() };
+
+      if(max_limit < str_size)
```
jeaye: It is physically impossible for a `size_t` to hold more than its max. This will never be true. I believe you're looking for the max number of digits to represent an `i64`, which would be 20. The max value is `9223372036854775807`, which has 19 chars, plus one more for the sign.

But do we need to check this? Does `stoll` not handle this for us?

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -1052,12 +1144,42 @@ namespace jank::runtime
...
+  object_ref parse_double(object_ref const o)
   {
     auto const typed_o{ dyn_cast<obj::persistent_string>(o) };
     if(typed_o.is_some())
     {
-      return std::stod(typed_o->data);
+      auto const str_size{ static_cast<size_t>(typed_o->data.size()) };
+      auto const max_limit{ std::numeric_limits<size_t>::max() };
+
+      if(max_limit < str_size)
```
jeaye: Same thing here.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/obj/number.cpp
```diff
@@ -252,6 +253,11 @@ namespace jank::runtime::obj
 
   bool real::equal(object const &o) const
   {
+    if(is_nan(this) || is_nan(&o))
```
jeaye: This is UB, since we need `detail::untagged` for raw pointers, due to the recent NaN boxing changes. Please read the big comment block in `oref.hpp` for details.

We need it for both `this` and `&o`. Most likely, this will just crash right now.

Also, we shouldn't use `is_nan` for `this`, since that would require building an `object_ref`, visiting that, and then actually checking `isnan`. Since we have `data` already, let's just check directly. Arithmetic operations are one of the most important parts in the jank runtime to keep fast, so keep that in mind. That also means we could use `isnan` and avoid a duplicate visit on `o` if we do it later in the fn.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/obj/number.cpp
jeaye: On second thought, why is this not working by default?

shantanu-sardesai: Indeed it's working... hmm... weird prior to this it wasn't 🤔. Although I also found it weird that I needed to do this since I was also expecting the `data == r->data` to fail... Anyhow, I'll remove this and read the comment on `oref.hpp` anyways. Thanks!

compiler+runtime/src/cpp/jank/runtime/obj/number.cpp
```diff
@@ -1,5 +1,6 @@
 #include <cmath>
 
+#include <jank/runtime/core/math.hpp>
```
jeaye: I don't think this is needed.

jeaye (approved):

PR #766, opened 2026-05-09
Update c_api functions to work with tagged pointers

djblue:
`jank_truthy` was returning `1` for `nil`/`false`.  I'm not great at the art of C++ casting, so I did use a LLM to help me fix it, but it appears to work correctly after.

### DIFF
--- a/compiler+runtime/src/cpp/jank/c_api.cpp
+++ b/compiler+runtime/src/cpp/jank/c_api.cpp
@@ -924,7 +924,8 @@ extern "C"
   jank_bool jank_truthy(jank_object_ref const o)
   {
     auto const o_obj(reinterpret_cast<object *>(o));
-    return static_cast<jank_bool>(truthy(o_obj));
+    auto const o_ref(object_ref{ o_obj });
+    return static_cast<jank_bool>(truthy(o_ref));
   }
 
   jank_bool jank_equal(jank_object_ref const l, jank_object_ref const r)
--- a/compiler+runtime/src/cpp/jank/c_api.cpp
+++ b/compiler+runtime/src/cpp/jank/c_api.cpp
@@ -923,8 +923,7 @@ extern "C"
 
   jank_bool jank_truthy(jank_object_ref const o)
   {
-    auto const o_obj(reinterpret_cast<object *>(o));
-    auto const o_ref(object_ref{ o_obj });
+    object_ref const o_ref{ reinterpret_cast<object *>(o) };
     return static_cast<jank_bool>(truthy(o_ref));
   }
 
--- a/compiler+runtime/src/cpp/jank/c_api.cpp
+++ b/compiler+runtime/src/cpp/jank/c_api.cpp
@@ -61,7 +61,7 @@ extern "C"
 {
   jank_object_ref jank_eval(jank_object_ref const o)
   {
-    auto const o_obj(reinterpret_cast<object *>(o));
+    object_ref const o_obj(reinterpret_cast<object *>(o));
     return __rt_ctx->eval(o_obj).erase().data;
   }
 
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -924,7 +924,8 @@ extern "C"
   jank_bool jank_truthy(jank_object_ref const o)
   {
     auto const o_obj(reinterpret_cast<object *>(o));
```
jeaye: Let's just change `auto` here to `object_ref`. Nothing else needs to change.

Previously, looks like it was using the `truthy(bool)` overload. This is a new change based on the recent pointer tagging work which changed how we handle pointers.

Ultimately, every `auto const foo_obj(reinterpret_cast<object*>(foo));` in this file should become `object_ref const foo_obj{ reinterpret_cast<object*>(foo) };`. You don't have to do that, though.

djblue: I think I might though. I'm running into other segfaults that I think are also due to tagged pointers 👌

jeaye: Yeah, I can see that. Catching every pointer assumption across the codebase was quite difficult. Changing all of these to be `object_ref const` instead of `auto const` should help.

djblue: Ok, I think I implemented the fix you specified. Please let me know otherwise.

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -952,18 +952,18 @@ extern "C"
...
   jank_i64
   jank_shift_mask_case_integer(jank_object_ref const o, jank_i64 const shift, jank_i64 const mask)
   {
-    auto const o_obj(reinterpret_cast<object *>(o));
+    object_ref const o_obj(reinterpret_cast<object *>(o));
     auto integer{ to_integer_or_hash(o_obj) };
     if(mask != 0)
     {
-      if(o_obj->type == object_type::integer)
+      if(o_obj.get_type() == object_type::integer)
```
djblue, line 966: I think this is the correct fix here.

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -929,18 +929,18 @@ extern "C"
...
 
   jank_uhash jank_to_hash(jank_object_ref const o)
   {
-    auto const o_obj(reinterpret_cast<object *>(o));
+    object_ref const o_obj(reinterpret_cast<object *>(o));
     return to_hash(o_obj);
   }
 
-  static i64 to_integer_or_hash(object const *o)
+  static i64 to_integer_or_hash(object_ref const o)
```
djblue, line 943: Is this okay?

compiler+runtime/src/cpp/jank/c_api.cpp
jeaye: Correct.

compiler+runtime/src/cpp/jank/c_api.cpp
jeaye: Yep. We want to be working with `object_ref` only now, rather than raw pointers. Since we can't reasonably dereference them anyway.

jeaye (approved):
Works for me! Does this fix your crashes?

djblue:
@jeaye Yeah, this fixes my issues. The main symptom I was experiencing was in the gdb `pretty_printer.py` for seq/maps. The implementation relies on [jank_truthy](https://github.com/jank-lang/jank/pull/767/changes#diff-f107a0ba1af1272ac713fd3bedc7543bf13938f5c856321f4a6791acfb60cf7cR16-R18) to know when to step traversing a seq.

PR #767, opened 2026-05-10
Implement gdb/pretty_print.py

djblue:
I wanted to improve the debugging experience for myself in vscode, so I put together this pretty printer for gdb. Here is what the UI looks like when combined with https://github.com/jank-lang/jank/pull/761:

image

It also requires some setup of gdb to work well: https://gist.github.com/djblue/c0c992710897a78eed3688e984e76d7e is a vscode gdb setup.


djblue:
So when I tried testing the pretty printer against the most recent version of main, I ran into some segfaults. I think the commit https://github.com/jank-lang/jank/commit/a7cff49856d3df04b3ebc4d15b556fe7264ed0de was likely causing issues. I think this just reveals that the pretty printer was too coupled to the internals of jank's object system. I decided to add two additional c-api functions to help get the data I need from jank's objects for pretty printing:

-  char const *jank_to_string(jank_object_ref o); 
    - I need this to turn immutable_strings into cstrings so I can use them from python
- char const *jank_box_canonical_type(jank_object_ref o);
    - I need this to cast the data of opaque_box to leverage the automatic printing of native types

gdb/pretty_print.py
```diff
@@ -100,11 +100,13 @@ def children(self):
         except Exception as e:
             yield str(repr(e))
 
-def jank_opaque_box_type(val):
+def jank_box_canonical_type(val):
     address = jank_address(val)
-    canonical_type = gdb.parse_and_eval(f"((jank::runtime::obj::opaque_box*){address})->canonical_type.c_str()")
```
jeaye: The problem is that these addresses are untagged. Pointers are represented in NaN space now, so untagged pointers need to be tagged first. Check out the big comment block in `oref.hpp` for more details.

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -71,6 +71,12 @@ extern "C"
     return __rt_ctx->read_string(s_obj->data).erase().data;
   }
 
+  char const *jank_to_string(jank_object_ref const o)
+  {
+    auto const str(try_object<obj::persistent_string>(reinterpret_cast<object *>(o)));
+    return str->data.c_str();
```
jeaye: 1. I would expect this to do `.to_string()` on the object, so it works on any object
2. I think what you want is `to_code_string()`, which is different from `to_string` since it puts quotes around strings and `#'` in front of vars and such

compiler+runtime/src/cpp/jank/c_api.cpp
```diff
@@ -580,6 +586,13 @@ extern "C"
     return op_box->data;
   }
 
+  char const *jank_box_canonical_type(jank_object_ref o)
```
jeaye: I would prefer to not have this fn, if you can just use `untagged` to get your `object_ref`.

jeaye:
Thank you, Chris!

PR #779, opened 2026-05-19
Add `jank.shell`

djblue:
The `boost/process` library has a v1 and v2 version. I choose v1 as `clojure.jank.shell/sh` doesn't need all the low lever control v2 offers. Also, v1 has the advantage of being a header-only library, where as v2 doesn't. I tried using only interop to implement this, but ran into issues, so I added a little `cpp/raw` helper to ease interop.

```clojure
% jank repl
user=> (use 'clojure.jank.shell)
nil
user=> (sh "echo" "hi")
{:exit 0, :out "hi\n", :err ""}
user=> (sh "uname")
{:exit 0, :out "Linux\n", :err ""}
user=> (with-sh-env {"FOO" "BAR"} (sh "printenv"))
{:exit 0, :out "FOO=BAR\n", :err ""}
user=> (sh "pwd")
{:exit 0, :out "/home/chris/repos/jank/compiler+runtime\n", :err ""}
user=> (with-sh-dir "/" (sh "pwd"))
{:exit 0, :out "/\n", :err ""}
user=> (sh "ls" "-l" "/no-such-thing")
{:exit 2, :out "", :err "ls: cannot access '/no-such-thing': No such file or directory\n"}
```


jeaye:
> I tried using only interop to implement this, but ran into issues

What were the issues you saw? I would like to get them ticketed.

.gitmodules
```diff
@@ -34,3 +34,6 @@
 [submodule "compiler+runtime/third-party/doctest"]
 	path = compiler+runtime/third-party/doctest
 	url = https://github.com/jank-lang/doctest.git
+[submodule "compiler+runtime/third-party/boost-process"]
+	path = compiler+runtime/third-party/boost-process
+	url = https://github.com/boostorg/process.git
```
jeaye: We need a fork for this, for future stability. I have made one here: https://github.com/jank-lang/boost-process

compiler+runtime/src/jank/clojure/jank/shell.jank
```diff
@@ -0,0 +1,141 @@
...
+        result (cpp/clojure.jank.shell.sh
+                (cpp/cast std.string (str/join " " (map shell-escape cmd)))
+                (cpp/cast std.string (or (:dir opts) "."))
+                (cpp/cast std.string (or (:in opts) ""))
+                (if-let [env (:env opts)]
+                  (let [new-env (cpp/boost.process.v1.environment.)]
+                    (doseq [[k v] env]
+                      (.emplace new-env (cpp/cast std.string k) (cpp/cast std.string v)))
+                    (cpp/cast boost.process.v1.environment new-env))
+                  (cpp/cast boost.process.v1.environment (cpp/boost.this_process.environment))))]
```
jeaye: Why are these two casts needed?

compiler+runtime/src/jank/clojure/jank/shell.jank
```diff
@@ -0,0 +1,141 @@
...
+  [args]
+  (let [default-encoding "UTF-8" ;; see sh doc string
+        default-opts {:out-enc default-encoding :in-enc default-encoding :dir *sh-dir* :env *sh-env*}
+        [cmd opts] (split-with string? args)]
+    [cmd (merge default-opts (apply hash-map opts))]))
+
+(cpp/raw "namespace clojure::jank::shell
+{
+  struct result {
+    int exit;
```
jeaye: This should be default-initialized.

compiler+runtime/src/jank/clojure/jank/shell.jank
```diff
@@ -0,0 +1,141 @@
...
+           convert the input string specified by the :in option to the
+           sub-process's stdin.  Defaults to UTF-8.
+           If the :in option provides a byte array, then the bytes are passed
+           unencoded, and this option is ignored.
+  :out-enc option may be given followed by :bytes or a String. If a
+           String is given, it will be used as a character encoding
+           name (for example \"UTF-8\" or \"ISO-8859-1\") to convert
+           the sub-process's stdout to a String which is returned.
+           If :bytes is given, the sub-process's stdout will be stored
+           in a byte array and returned.  Defaults to UTF-8.
```
jeaye: Let's remove any docs for options we don't support.

compiler+runtime/src/jank/clojure/jank/shell.jank
```diff
@@ -0,0 +1,141 @@
...
+           String[] if you are a masochist).
+  :dir     override the process dir with a String or java.io.File.
+
+  You can bind :env or :dir for multiple operations using with-sh-env
+  and with-sh-dir.
+
+  sh returns a map of
+    :exit => sub-process's exit code
+    :out  => sub-process's stdout (as byte[] or String)
+    :err  => sub-process's stderr (String via platform default encoding)"
```
jeaye: Let's update docs to represent what we're doing, rather than the old JVM stuff.

jeaye (changes requested):
Very cool to see this come together!

jeaye:
@djblue I saw you called out that you used an LLM for #780. Please keep to the contribution policy and not AI-generate any code for jank. If this PR has AI-generated code, let me know.

compiler+runtime/src/jank/clojure/jank/shell.jank
djblue: It's so the if doesn't complain about mismatching types.

jeaye: `new-env` was created as a `(cpp/boost.process.v1.environment.)`, so this cast is doing nothing. Looks like `(cpp/boost.this_process.environment)` is returning a `native_environment`. Can we create `new-env` as a `native_environment`? If not, we can at least remove the duplicate cast.

djblue:
I did not use an LLM to generate the code for this PR. However, I did use an LLM to explore the different ways to spawn a process in C++, including: `popen`/`pclose`, boost v2 and boost v1. I also used test code from the boost repo itself for more examples of how to achieve this. They way I developed this namespace was as follows:

- I started with the clojure.java.shell as the base
- I deleted the bits that I didn't need
- I tried implementing a basic example with just support for `:exit`/`:out`/`:err` interactively at the REPL
    - Because boost is so template heavy, I ran into some interop issues.
    - Here are some of the errors I was getting with the different overloads I was trying:
        - `JIT session error: Failed to materialize symbols: { (main, { _ZN5boost7process2v15childC2EONS1_6detail5posix12child_handleERSt10shared_ptrISt6atomicIiEE }) }`
        - `JIT session error: Failed to materialize symbols: { (main, { _ZN5boost6fusion4nextINS0_15vector_iteratorINS0_5tupleIJNS_7process2v16detail5posix12exe_cmd_initIcEENS7_8env_initIcEEEEELi0EEEEEKNS0_9result_of4nextIT_E4typeERKSG_ }) }`
        - `JIT session error: Failed to materialize symbols: { (main, { _ZN5boost7process2v16detail5posix14start_dir_initIcED2Ev }) }`
- I added the `cpp/raw` function to sidestep the interop issue
- Once I got that working, I added support for `:in`
- Then I added support for `:env`, which uses the process's default env when not supplied
    - This is why I thought I needed to cast both branches of the if to `boost.process.v1.environment`
- I then tested the code in portal and ran into an issue where the process can have a `" "` in the name
- I explored passing in a `std::string` for the bin and `std::vector<std::string>` for args to avoid escaping issues
    - But when you provide arguments like this to boost, it uses a different mechanism to spawn the process that doesn't leverage your `$PATH` and assumes the bin is a full resolved binary
- I then added `shell-escape` to fix the `" "` issue, while still getting default `$PATH` bin resolution

compiler+runtime/src/jank/clojure/jank/shell.jank
djblue: I see the following error when trying to use a `native_environment`:

```
input_line_31:133:5: error: object of type 'boost::process::v1::basic_native_environment<char>' cannot be assigned because its copy assignment operator is implicitly deleted
  133 | s48 = v51;
```

jeaye: Ok, you could try to move it, then. `(cpp/jtl.move new-env)`

If not, keep it as an `environment` and just cast the second branch.

djblue: I get this error when trying to move:

```
Mismatched 'if' branch types 'jtl::remove_reference_t<boost::process::v1::basic_native_environment<char> &> &&' and 'boost::process::v1::basic_native_environment<char>'.
Each branch of an 'if' must have the same type.
```

jeaye: Ok, this is an analysis bug. Those are the same type except one has a `&&`, so we should strip references and get the canonical type for this comparison. I've made a note of that locally.

jeaye (dismissed):
The changes look good to me! The only thing I'm chewing on is the namespace. `clojure.jank.shell` feels weird, since it has both `clojure` and `jank` in the name. I'm thinking we just do `jank.shell`.

If you want to apply that change, I can merge once it's done (also update the `cpp/raw` namespace). Otherwise, I'll apply it myself once I hit a good stashing point tonight or tomorrow morning.

Thanks, Chris!

djblue:
For the namespace, I think Clojure puts all their platform specific stuff under [`clojure.java`](https://github.com/clojure/clojure/tree/master/src/clj/clojure/java), I think the CLR does something similar under [`clojure.clr`](https://github.com/clojure/clojure-clr/tree/master/Clojure/Clojure.Source/clojure/clr). In mirroring that we get some amount predictability and perhaps when [clojuredocs](https://github.com/nubank/clojuredocs/pull/37) adds support for more dialects, jank could slot in more easily. However, I have seen dialects like basilisp do something else, [`basilisp.shell`](https://github.com/basilisp-lang/basilisp/blob/main/src/basilisp/shell.lpy). Let me know if you still want me to change it. Perhaps `clojure.cpp.shell` or `clojure.native.shell`?

jeaye:
Thanks for doing the research.

> Perhaps `clojure.cpp.shell` or `clojure.native.shell`?

I've considered the same two, but neither is actually jank. Let's go with `jank.shell`.

jeaye (approved):

PR #786, opened 2026-05-23
Rewrite jank.perf in pure jank + improvements

kylc:
Replaces the minimal `jank.perf` implementation, which was mostly C++ code, with a more full-featured version written in jank. The intent is to provide built-in functionality similar to criterium in Clojure, but more closely resembling BenchmarkTools.jl from Julia.

I have left some high-level TODOs in the main module which we can leave for later PRs.

Note that this PR bumps the bdwgc submodule commit. For CI and before merge, @jeaye will need to push this commit up to the `jank-lang/bdwgc` fork master branch.

Example output and usage:

image

```clojure
(ns example
  (:require [jank.perf :as perf])
  (:include "random"))

(perf/bench
 {:label   "make garbage"
  :warmups 0
  :epochs  5000}
 (str "i" "like" "turtles" (rand))
 (apply + (range 0 1000))
 (vec (range 0 5000)))

(println "-------")

(defn fib [n]
  (if (<= n 1) n
      (+ (fib (- n 1))
         (fib (- n 2)))))

(perf/bench
 {:label   "fib(25)"
  :warmups 1
  :epochs  100}
 (fib 25))

(println "-------")

(defn sleep-normrnd [μ σ]
  ;; simulate some work
  (let [t0 (cpp/std.chrono.steady_clock.now)
        ;; Box-Muller to compute Z in the standard distribution.
        U1 (cpp/double (rand 1.0))
        U2 (cpp/double (rand 1.0))
        Z  (* (cpp/sqrt (* -2.0 (cpp/log U1))) (cpp/cos (* 2.0 cpp/std.numbers.pi U2)))

        ;; Transform into our normal distribution.
        X  (-> Z (* (cpp/double σ)) (+ (cpp/double μ)))]
    ;; Sleep such that the total execution time of this method is approx. equal
    ;; to X.
    (cpp/std.this_thread.sleep_until
     (cpp/+ t0 (cpp/std.chrono.nanoseconds (cpp/int (* 1e6 X)))))))

(perf/quick-bench (sleep-normrnd 5.0 2.5))

(println "-------")

(perf/bench
 {:label   "sleep-normrnd(μ = 5ms, σ = 2.5ms)"
  :warmups 1
  :epochs  1000}
 (sleep-normrnd 5.0 2.5))

(println "-------")

(perf/bench
 {:label   "sleep-normrnd(μ = 5ms, σ = 2.5ms)"
  :warmups 1
  :epochs  2500}
 (sleep-normrnd 5.0 2.5))

(println "-------")

(->
 (perf/quick-bench-to-data (sleep-normrnd 5.0 2.5))
 (dissoc :samples)
 (println))

(println "-------")

(perf/bench-compare
 {:label  "sleep-normrnd 5ms vs. 8ms"
  :epochs 1000}
 (sleep-normrnd 5.0 2.5)
 (sleep-normrnd 8.0 1.5))
```


kylc:
Ready for review @jeaye

compiler+runtime/src/cpp/jank/analyze/expr/cpp_member_call.cpp
```diff
@@ -1,11 +1,8 @@
+#include <CppInterOp/CppInterOp.h>
+
 #include <jank/analyze/expr/cpp_member_call.hpp>
 #include <jank/detail/to_runtime_data.hpp>
 
-namespace CppImpl
-{
-  void *GetFunctionReturnType(void *);
-}
-
```
kylc, line 8: This was causing a Windows build failure because it's missing the dllexport attribute as it is actually declared in CppInterop. I replaced it with the include, but I'm not sure how invested you are in keeping the include out of here.

I imagine the build failures all came up in this PR because of some transitive imports that were getting pulled in via the perf.hpp/perf_native.hpp includes, which don't exist anymore.

jeaye (approved):
Looks great to me, dude! Thanks for the new rich benchmarking API!

compiler+runtime/src/cpp/jank/analyze/expr/cpp_member_call.cpp
jeaye: Yeah, I've seen this failure crop up before as well. I'm surprised we still had some uses of `CppImpl`, since I had fixed a few of those already.

PR #788, opened 2026-05-24
Fix shuffle/re-find vector ctor usage

kylc:
`shuffle` and `re-find` native implementations were improperly calling the `vector(std::initializer_list)` constructor with 2 elements rather than the `vector(Iter first, Iter last)` constructor.

Before:
```
user=> (shuffle (range 0 10))
[0.000000 0.000000]
user=> (re-find #"(\d+) (\d+) (\d+)" "1 2 3 4 5 6 7")
[0.000000 0.000000]
```

After:
```
user=> (shuffle (range 0 10))
[4 9 2 0 5 1 6 8 3 7]
user=> (re-find #"(\d+) (\d+) (\d+)" "1 2 3 4 5 6 7")
["1 2 3" "1" "2" "3"]
```

Note there are some other similar instances in core that might need a closer look.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -1185,7 +1185,7 @@ namespace jank::runtime
         std::shuffle(vec.begin(), vec.end(), g);
 
         return make_box<obj::persistent_vector>(
-          runtime::detail::native_persistent_vector{ vec.begin(), vec.end() });
+          runtime::detail::native_persistent_vector(vec.begin(), vec.end()));
       },
       coll);
   }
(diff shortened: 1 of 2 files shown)


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
kylc, line 737: I guess this usage might be OK because these iterator types (whatever they may be) are not implicitly convertible to `object_ref`s, so the `initializer_list<object_ref>` ctor does not match? But that seems precarious.

jeaye: Right, there's no way for these to become `object_ref`. This is the key downside of uniform initialization. I don't generally want initializer_list ctors.

jeaye (approved):
Thanks!

jeaye:
Oh, I can't merge because your commit is not signed. Would you mind please signing your commits?

https://github.com/jank-lang/jank/blob/main/doc/git-gpg.md

kylc:
Should be good now.

jeaye:
Thanks!

PR #789, opened 2026-05-24
Implement `char`

djblue:
I initially implemented `char` without consideration for Unicode. However, I don't know Unicode outside of ASCII all that well, and the [spec](https://datatracker.ietf.org/doc/html/rfc3629#section-3) helped a bit, but I assumed other cpp projects had this problem as well. I decided to "borrow" the implementation from [`imgui`](https://github.com/ocornut/imgui/blob/master/imgui.cpp#L2691-L2724) as it's a well known popular project. The code is licensed under MIT, so I assume it is ok to use in jank?

Jank:
```clojure
% jank repl
user=> (char 65)
\A
user=> (char 65.0)
\A
user=> (char 0x1F600)
\😀
```

Clojure:
```clojure
% clj
Clojure 1.12.2
user=> (char 65)
\A
user=> (char 65.0)
\A
user=> (char 0x1F600)
Execution error (IllegalArgumentException) at user/eval5 (REPL:1).
Value out of range for char: 128512
user=> (String. (Character/toChars 0x1F600))
"😀"
```

The one thing I noticed is that for Clojure, `char` blows up much sooner since I think the underlying jvm strings are 2 bytes per "char". Do we want jank to have the same limitation, or should we support more because we can and this is more of a platform limitation?

### DIFF
--- a/compiler+runtime/include/cpp/jtl/immutable_string.hpp
+++ b/compiler+runtime/include/cpp/jtl/immutable_string.hpp
@@ -64,11 +64,44 @@ namespace jtl
 
     static constexpr size_type npos{ std::numeric_limits<size_type>::max() };
 
+    static constexpr i64 max_unicode_value{ 0x10FFFF };
+
     friend struct jtl::string_builder;
 
     constexpr immutable_string() noexcept
     {
       set_small_size(0);
+    }
+
+    // Taken from imgui.cpp (ImTextCharToUtf8)
+    explicit constexpr immutable_string(i32 code_point) noexcept
+    {
+      if(code_point < 0x80)
+      {
+        init_small_fill(0, 1);
+        store.small[0] = static_cast<value_type>(code_point);
+      }
+      else if(code_point < 0x800)
+      {
+        init_small_fill(0, 2);
+        store.small[0] = static_cast<value_type>(0xc0 + (code_point >> 6));
+        store.small[1] = static_cast<value_type>(0x80 + (code_point & 0x3f));
+      }
+      else if(code_point < 0x10000)
+      {
+        init_small_fill(0, 3);
+        store.small[0] = static_cast<value_type>(0xe0 + (code_point >> 12));
+        store.small[1] = static_cast<value_type>(0x80 + ((code_point >> 6) & 0x3f));
+        store.small[2] = static_cast<value_type>(0x80 + (code_point & 0x3f));
+      }
+      else if(code_point <= 0x10FFFF)
+      {
+        init_small_fill(0, 4);
+        store.small[0] = static_cast<value_type>(0xf0 + (code_point >> 18));
+        store.small[1] = static_cast<value_type>(0x80 + ((code_point >> 12) & 0x3f));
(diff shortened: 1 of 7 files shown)


compiler+runtime/include/cpp/jtl/immutable_string.hpp
```diff
@@ -64,13 +64,46 @@ namespace jtl
...
+
     friend struct jtl::string_builder;
 
     constexpr immutable_string() noexcept
     {
       set_small_size(0);
     }
 
+    // Taken from imgui.cpp (ImTextCharToUtf8)
+    explicit constexpr immutable_string(i32 code_point) noexcept
```
djblue: I'm not sure what the implication of putting this code here is, and I'm not sure if there is a better type for code_point (although I should probably add `const`). I assumed since we need access to the underlying storage to initialize the string, this would be appropriate to add here. Sorry if I've made a lot of bad assumption 😅

djblue: I switched it to `u32 const code_point` as I feel that's a bit better.

djblue:
Another interesting case I ran into:

Jank:
```clojure
% jank repl
user=> (seq "😀")
(\� \� \� \�)
```

Clojue:
```clojure
% clj
Clojure 1.12.2
user=> (seq "😀")
(\? \?)
```

It looks like jank assumes 8 bytes per char, while the jvm assumes 16 bytes per char? Am I making a bad assumption that jank strings are utf8?

compiler+runtime/include/cpp/jtl/immutable_string.hpp
```diff
@@ -64,13 +64,46 @@ namespace jtl
...
     friend struct jtl::string_builder;
 
     constexpr immutable_string() noexcept
     {
       set_small_size(0);
     }
 
+    // Taken from imgui.cpp (ImTextCharToUtf8)
+    explicit constexpr immutable_string(u32 const code_point) noexcept
+    {
```
jeaye: I believe the correct fn for this is https://en.cppreference.com/c/string/multibyte/wcrtomb

See the opposite direction, which we do in the lexer:

```cpp
  static jtl::result<codepoint, error_ref>
  convert_to_codepoint(jtl::immutable_string_view const sv, movable_position const &pos)
```

Also, this should be a ctor of `obj::character`, not for `immutable_string`.

jeaye (changes requested):
Thanks, Chris! Hopefully I provided some good direction.

compiler+runtime/include/cpp/jtl/immutable_string.hpp
djblue: Yeah, that makes more sense 😂

jeaye:
> It looks like jank assumes 8 bytes per char, while the jvm assumes 16 bytes per char? Am I making a bad assumption that jank strings are utf8?

jank does indeed assume UTF-8. This is how we want it.

jeaye (approved):
Nice!

PR #790, opened 2026-05-24
Implement `int` for `char`

djblue:
Jank:
```clojure
% jank repl
user=> (int \newline)
10
user=> (int \A)
65
```
Clojure:
```clojure
% clj
Clojure 1.12.2
user=> (int \newline)
10
user=> (int \A)
65
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
@@ -954,6 +954,11 @@ namespace jank::runtime
 
   i64 to_int(object_ref const l)
   {
+    if(l.get_type() == object_type::character)
+    {
+      return expect_object<obj::character>(l)->to_integer();
+    }
+
     return visit_number_like([](auto const typed_l) -> i64 { return typed_l->to_integer(); }, l);
   }
 
--- a/compiler+runtime/include/cpp/jank/runtime/obj/character.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/character.hpp
@@ -26,6 +26,7 @@ namespace jank::runtime::obj
     jtl::immutable_string to_code_string() const override;
     uhash to_hash() const override;
 
+    /* Character does not fully support `behavior::number_like`, but can be converted to an integer. */
     i64 to_integer() const;
 
     /*** XXX: Everything here is immutable after initialization. ***/
--- a/compiler+runtime/include/cpp/jank/runtime/obj/character.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/character.hpp
@@ -26,6 +26,8 @@ namespace jank::runtime::obj
     jtl::immutable_string to_code_string() const override;
     uhash to_hash() const override;
 
+    i64 to_integer() const;
+
     /*** XXX: Everything here is immutable after initialization. ***/
 
     /* Holds the raw form of the character bytes. Supports Unicode. */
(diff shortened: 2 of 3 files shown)


compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -954,6 +954,11 @@ namespace jank::runtime
 
   i64 to_int(object_ref const l)
   {
+    if(l.get_type() == object_type::character)
+    {
+      return expect_object<obj::character>(l)->to_integer();
+    }
+
```
jeaye, line 961: This is not needed, if you make `character` number-like. That involves adding adding `to_real` as well:

```cpp
  template <typename T>
  concept number_like = requires(T * const t) {
    { t->to_integer() } -> std::convertible_to<i64>;
    { t->to_real() } -> std::convertible_to<f64>;
  };
```

Let's do this instead of having a special case here (and wherever else we need it). That's why we have the behaviors.

For `to_real`, just do `static_cast<f64>(to_integer());`.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
djblue: I tried that initially and it didn’t appear to work.

jeaye: You'll need to add `character` to `visit_number_like`.

djblue: Do I need to update this guy:

```cpp
  template <typename F1, typename... Args>
  [[gnu::always_inline, gnu::flatten, gnu::hot]]
  auto visit_number_like(F1 const &fn, object_ref const erased, Args &&...args)
  {
    return visit_number_like(
      fn,
      [=]() -> decltype(fn(obj::integer_ref{}, std::forward<Args>(args)...)) {
        throw std::runtime_error{ "not a number: " + to_code_string(erased) };
      },
      erased,
      std::forward<Args>(args)...);
  }
```

How do I add character to it?

jeaye: No, that overload just calls the main `visit_number_like`. You only need to update the one. Adding `character` to the main one just involves adding a case for `character`.

```cpp
      case object_type::character:
        return fn(expect_object<obj::character>(erased), std::forward<Args>(args)...);
```

djblue: Maybe I'm doing it work, but when I add that I get the following:

```cpp
/home/chris/repos/jank/compiler+runtime/src/cpp/jank/runtime/obj/big_decimal.cpp:150:25: error: invalid operands to binary expression ('const native_big_decimal' (aka 'const boost::multiprecision::number<boost::multiprecision::cpp_dec_float<100>, boost::multiprecision::et_off>') and 'jtl::immutable_string')
  150 |         return abs(data - typed_o->data) < std::numeric_limits<f64>::epsilon();
      |                    ~~~~ ^ ~~~~~~~~~~~~~
```

jeaye: Ah, I see. We can't treat this as a number, since that would break in other places. Ok. Please add a comment above `to_integer` in `character.hpp` saying that characters are not full number-likes, but that they can be converted to integers.

jeaye (approved):
Thanks!

PR #793, opened 2026-05-24
Fix `(= (seq "hi") (seq "hi"))`

djblue:
Before:
```clojure
% jank repl
user=> (= (seq "hi") (seq "hi"))
false
```

After:
```clojure
% jank repl
user=> (= (seq "hi") (seq "hi"))
true
```

Clojure:
```clojure
% clj
Clojure 1.12.2
user=> (= (seq "hi") (seq "hi"))
true
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/equal.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/equal.cpp
@@ -13,7 +13,7 @@ namespace jank::runtime
     }
 
     obj::character_ref const typed_rhs{ expect_object<obj::character>(rhs) };
-    return typed_rhs->data[0] == lhs;
+    return typed_rhs->data.size() == 1 && typed_rhs->data[0] == lhs;
   }
 
   bool equal(object_ref const lhs, object_ref const rhs)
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/cpp/jank/runtime/core/equal.cpp
```diff
@@ -12,8 +12,8 @@ namespace jank::runtime
       return false;
     }
 
-    auto const typed_rhs{ expect_object<obj::character>(rhs) };
-    return typed_rhs->to_hash() == static_cast<uhash>(lhs);
```
djblue, line 16: I think the way [`jtl::immutable_string`](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/include/cpp/jtl/immutable_string.hpp#L648-L662) computes its hash is different from how `static_cast<uhash>` works.

jeaye: `static_cast<uhash>` doesn't do anything aside from shift the type. It uses the value directly. So, yeah, we're not even hashing `lhs` here.

jeaye:
Thanks!

Issue #566, opened 2025-10-24
Add jank build system tooling


jeaye:
# Motivation
Many, if not most, jank libraries will not be pure Clojure. Instead, they will be wrapping a native library (C, C++, etc) of some sort. This intrinsically adds a dependency to this native library which then needs to be realized when a user tries to use the jank library. For example:

1. I write a `glfw-jank` library to wrap the GLFW native library
2. Sally wants to use my library, so she adds it as a dep in her `project.clj`
3. leiningen pulls down my library from Clojars for her to use

At this point, Sally will have a copy of the jank code which wraps GLFW, but she may not have GLFW installed on her system. If it is installed, perhaps it's not the correct version. We need to be able to detect this. Furthermore, I may have opted to just compile GLFW from source, rather than deal with system versions, so leiningen will need to trigger a compilation of GLFW from source prior to trying to load it in jank.

Compiling a native library from source may involve CMake, Meson, Bazel, pkg-config, or other build tools. It'll also involve finding an appropriate C or C++ compiler, includes, and other transient native dependencies.

# Design proposal
jank will support an optional top-level `build.clj` alongside the `project.clj`. This `build.clj` will be a Clojure file which can run with babashka and will have access to a `jank.build` namespace (as an implicit Clojars dep) which has functions present to aid in common operations:

1. Finding system libraries
2. Building native libraries from source (CMake, Autotools, etc)
3. Generating files
4. Exposing this information back up to leiningen (lib files, lib paths, include paths, object files, etc)

When lein-jank starts any operation, it'll first scan the module path for any `build.clj` files and ensure that they have been executed. Their outputs will be cached and the script will not run again unless a new version of the library is pulled. If the script failed, there will be no cache and lein-jank will try to run the script again.

# Security considerations
jank will not support using packaged binaries, either as object files or static/dynamic libaries. Each user who pulls down a library will compile that library from source. Library build scripts may choose to target system dependencies, which will be binaries, but those are fair game since they're already installed on a user's machine. For example libc, libz, libgl, etc.

This may change at a later date, but it's how we're starting, both for security concerns and practical portability concerns.

# References
* Cargo build script output format: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

jeaye:
This is done, thanks to Kyle.

PR #807, opened 2026-06-07
Add native build support to lein-jank

kylc:
Adds initial support to `lein-jank` for for jank projects which either include their own native build step specified by `jank-build.bb`, or have dependencies which do.

What works:
- Native jars are downloaded, extracted, build scripts are run, and relevant jank flags are included in the main executable
- If a dependency does not change, and none of its inputs change, then its output will be cached
- Build scripts are run in a jail constructed by bubblewrap on Linux
- A build script in the project root is always run

What is TODO:
- Build out standard libraries for use in build scripts to interface with CMake, pkg-config, etc.
- The build cache fingerprint should include more things like $CC/$CXX, etc. (see TODO in code)
- Wire up more of the build options to lein CLI flags like `lein --verbose-build run` (is this possible in lein?)
- Wire up more project.clj flags to the build metadata (like `:static-build`)
- Will not work on non-Linux platforms due to bubblewrap, will need to add a `--disable-sandbox` option, eventually add Mac support with seatbelt
- Eventually, needs a book chapter on how to write build scripts

To test, first `lein install` in this checkout of `lein-jank`. Then check out https://github.com/kylc/lein-jank-playground and do a `lein run` inside `json-formatter`. You will need a C++ build toolchain, CMake, and glfw3 (at least). This repo shows off some different build scripts, and a hierarchy of native dependencies in `jank-json-sys` -> `jank-json` -> `json-formatter`.

Implements most of https://github.com/jank-lang/jank/issues/566

### DIFF
--- a/lein-jank/src/leiningen/jank/sandbox/core.clj
+++ b/lein-jank/src/leiningen/jank/sandbox/core.clj
@@ -1,12 +1,28 @@
 (ns leiningen.jank.sandbox.core
   (:require [babashka.process :as proc]
-            [leiningen.jank.sandbox.bwrap :as bwrap]))
+            [leiningen.jank.sandbox.bwrap :as bwrap]
+            [leiningen.core.main :as lmain]))
 
 (defn process
   "Pass `cmd` and `sh-opts` to `babashka.process/process` running inside of a
-  sandboxed container specified by `sandbox-opts`."
-  [sandbox-opts cmd sh-opts]
+  sandboxed container specified by `sandbox-opts`.
+
+  Setting the `enable?` flag to false allows sandboxing to be skipped entirely
+  and the command will be run as a standard shell process."
+  [enable? sandbox-opts cmd sh-opts]
   ;; TODO: handle Windows and MacOS, either by a fake sandbox or using their
   ;; native constructs for containerization.
-  (let [bwrap-prefix (bwrap/bwrap (into bwrap/standard-binds sandbox-opts))]
-    (proc/process (concat bwrap-prefix cmd) sh-opts)))
+  (cond
+    (not enable?)
+    (proc/process cmd sh-opts)
+
+    (not (bwrap/which-bwrap))
+    (lmain/abort
+     (str "No 'bwrap' executable found. If bubblewrap is not available on your"
+          " platform then you can build with the lein option :disable-sandbox,"
+          " however this allows potentially nefarious build scripts free rein"
+          " over your system. Use with care!"))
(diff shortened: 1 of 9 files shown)


lein-jank/src/leiningen/jank/sandbox/core.clj
```diff
@@ -0,0 +1,12 @@
...
+  (:require [babashka.process :as proc]
+            [leiningen.jank.sandbox.bwrap :as bwrap]))
+
+(defn process
+  "Pass `cmd` and `sh-opts` to `babashka.process/process` running inside of a
+  sandboxed container specified by `sandbox-opts`."
+  [sandbox-opts cmd sh-opts]
+  ;; TODO: handle Windows and MacOS, either by a fake sandbox or using their
+  ;; native constructs for containerization.
+  (let [bwrap-prefix (bwrap/bwrap (into bwrap/standard-binds sandbox-opts))]
```
jeaye: Right now, this PR will fail on non-Linux, since `bwrap` won't exist. Are you open to conditionally just running the command without the sandbox, for non-Linux platforms? That way people can still use `lein-jank` on macOS and Windows, just without the sandboxing.

I'm more hesitant to merge something that is only going to work on Linux.

jeaye (changes requested):
Thank you, Kyle! Overall, I am really pleased with how this turned out. My only request is that we merge this as something which can *work* on at least Linux and macOS, even without sandboxing.

If this is something you're open to doing, please update the PR. If not, I will jump in and apply the changes once I wrap up my current task.

lein-jank/src/leiningen/jank/sandbox/core.clj
kylc: I added a `:disable-sandbox` argument which skips bwrap. The flow is now:
1. If you have bwrap and `lein run` then the build will run inside a sandbox.
2. If you don't have bwrap and `lein run` you will get a message instructing you to either install it or run with `lein run :disable-sandbox`.
3. If you run with `lein run :disable-sandbox` then no containerization is applied to the build, and you receive a warning that you are doing something which is potentially unsafe.

jeaye (dismissed):
Thanks, Kyle! Any changes you want done before this gets merged?

kylc:
OK this is ready to go. I'm sure there will be plenty of followup work to address the TODOs and improve robustness.

jeaye:
What a great start to an incredible feature. Thank you, Kyle!

PR #808, opened 2026-06-10
jank's math epic - part 2

shantanu-sardesai:
> [!important]
> ~~Requires a Clojure test suite repoint. [Clojure test suite PR](https://github.com/jank-lang/clojure-test-suite/pull/906)~~. Done.

- Fixes jank's print for reals.
- Catches us up on a bunch of math functions.

Brings the current tally to **150 tests**.

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core/math.hpp
@@ -1242,6 +1242,7 @@ namespace jank::runtime
   object_ref number(object_ref const o);
 
   bool is_integer(object_ref const o);
+  bool is_integral(object_ref const o);
   bool is_real(object_ref const o);
   bool is_ratio(object_ref const o);
   bool is_boolean(object_ref const o);
--- a/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
@@ -19,6 +19,11 @@ namespace jank::runtime
       return i->data;
     }
 
+    if(auto const bi{ dyn_cast<obj::big_integer>(o) }; bi.is_some())
+    {
+      return bi->to_integer();
+    }
+
     throw std::runtime_error{ util::format("An integer was required here, but a {} was provided.",
                                            object_type_str(o.get_type())) };
   }
--- a/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
@@ -539,6 +539,11 @@ namespace jank::runtime
     return detail::is_tagged_small_int(o.raw()) || o.get_type() == object_type::integer;
   }
 
+  bool is_integral(object_ref const o)
+  {
+    return is_integer(o) || is_big_integer(o);
+  }
+
   bool is_real(object_ref const o)
   {
     return detail::is_tagged_small_real(o.raw()) || o.get_type() == object_type::real;
--- a/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
@@ -677,8 +677,12 @@ namespace jank::runtime
         {
           return make_box<obj::big_integer>(typed_o->data);
         }
-        else if constexpr(
-          jtl::is_any_same<T, obj::real, obj::small_real, obj::ratio, obj::big_decimal>)
+        else if constexpr(jtl::is_any_same<T, obj::real, obj::small_real, obj::big_decimal>)
+        {
+          return make_box<obj::big_integer>(
+            native_big_decimal(typed_o->data).convert_to<native_big_integer>());
+        }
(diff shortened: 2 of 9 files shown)


compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -539,6 +539,11 @@ namespace jank::runtime
     return detail::is_small_int(o.raw()) || o.get_type() == object_type::integer;
   }
 
+  bool is_integral(object_ref const o)
+  {
+    return is_integer(o) || is_big_integer(o);
+  }
+
```
jeaye, line 551: You have made this distinction here, between integral and integer, but I don't think it's correct. I think the correct change is to update `is_integer` to include big integers and then to update the various uses of `is_integer` to ensure that big integers are supported. For example, lookup in vectors should support big integers: `([:a :b] 1N)`. With this change, it will not.

jeaye (changes requested):

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
shantanu-sardesai: Clojure distinguishes between [int?](https://github.com/clojure/clojure/blob/6a4ba6aedc8575768b2fff6d9c9c7e6503a0a93a/src/clj/clojure/core.clj#L1417-L1420) & [integer?](https://github.com/clojure/clojure/blob/6a4ba6aedc8575768b2fff6d9c9c7e6503a0a93a/src/clj/clojure/core.clj#L1393-L1398) so we'll probably have to keep this distinction around. 🤔

> For example, lookup in vectors should support big integers: ([:a :b] 1N). With this change, it will not.

The support for using big integer values doesn't exist on `main` either. I can add support for that here but I expect some other test in the `clojure-test-suite` to require this change. Just trying to keep this PR small. What do you think?

compiler+runtime/include/cpp/jank/runtime/core/math.hpp
```diff
@@ -1199,13 +1199,9 @@ namespace jank::runtime
...
       return 0;
     }
-    else if constexpr(jtl::is_same<L, object_ref>)
-    {
-      return l.to_integer();
-    }
     else
     {
-      return l->to_integer();
+      return visit_number_like([](auto const typed_l) -> i64 { return typed_l.to_integer(); }, l);
```
shantanu-sardesai: I know we don't like doing this but currently jank is unable to compile certain cases:

```jank
user=> (char nil)
─ runtime/unsupported-behavior ────────────────────────────────────────────────────
error: Objects of type 'nil' do not support the 'number_like' behavior.

─────┬─────────────────────────────────────────────────────────────────────────────
     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-yal32G
─────┼─────────────────────────────────────────────────────────────────────────────
  1  │ (char nil)
     │ ^^^^^^^^^^ Found here.
─────┴─────────────────────────────────────────────────────────────────────────────

user=> (char [])
─ runtime/unsupported-behavior ────────────────────────────────────────────────────
error: Objects of type 'persistent_vector' do not support the 'number_like'
       behavior.

─────┬─────────────────────────────────────────────────────────────────────────────
     │ /var/folders/xd/9v1dcg0d46l7d19830vq0lhc0000gn/T/jank-repl-yal32G
─────┼─────────────────────────────────────────────────────────────────────────────
  1  │ (char [])
     │ ^^^^^^^^^ Found here.
─────┴─────────────────────────────────────────────────────────────────────────────

user=> (def x [])  ;; Needed to erase the type.
#'user/x

user=> (int x)
 ─ runtime/unsupported-behavior ────────────────────────────────────────────────────
error: Objects of type 'persistent_vector' do not support the 'number_like'
       behavior.
```

shantanu-sardesai: Is there a better way to guard against this?

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
jeaye: I know we don't support this in `main`. I am just trying to make sure we go the right direction. With this distinction, we'll later need to update existing usages of `is_integer` to be `is_integral`, as well as the expectations which come after those checks.

compiler+runtime/include/cpp/jank/runtime/core/math.hpp
jeaye: > Is there a better way to guard against this?

Yes. Do not do this. Do not ever visit an already typed object.

We already have the complete type. We're already in a template. Take a look at `detail::valid_boxed_math<L>` and how we're using it in other templates.

shantanu-sardesai: Continuing on Slack.

shantanu-sardesai: Removed the change.

jeaye (approved):
Nice!

PR #809, opened 2026-06-12
Support build-time dependencies for jank-build.bb scripts.

kylc:
Adds support for build dependencies: dependencies for your `jank-build.bb` script. For example, a common `jank-build-cmake` library which includes helpers for executing CMake, to be used by all `jank-build.bb` scripts in the wild which drive a CMake build.

This mimics the Cargo [build-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#build-dependencies) feature.

I have updated my lein-jank-playground to demo this:
1. jank-json-sys declares a `:build-dependencies` entry in [its project.clj file](https://github.com/kylc/lein-jank-playground/blob/main/jank-json-sys/project.clj#L6)
2. It uses the dependency in [its `jank-build.bb` file](https://github.com/kylc/lein-jank-playground/blob/main/jank-json-sys/jank-build.bb#L5)
3. This dependency is resolved via Maven as usual, in this case to an artifact on clojars (source: https://github.com/kylc/lein-jank-playground/tree/main/jank-build-cmake). It is added to the babashka classpath when the build script is executed.

Also includes some misc. fixups and some integration tests for the build system.

### TODO
- ~Validate (build?)-dependencies-of-build-dependencies behavior~
- ~Think about whether using this non-standard pom.xml dependency scope `jank-build` will cause any problems down the line~

### DIFF
--- a/lein-jank/src/leiningen/jank.clj
+++ b/lein-jank/src/leiningen/jank.clj
@@ -173,20 +173,35 @@
           (with-meta result (apply deep-merge-metadata maps))
           result)))))
 
-(defn verbatim->filespecs [{:keys [root verbatim-paths] :as project}]
+(defn verbatim->filespecs
+  "Specify filespecs to copy files or directories in the :verbatim-paths project
+  key into the output archive."
+  [{:keys [root verbatim-paths] :as project}]
   ;; TODO: I couldn't find a good way to insert paths verbatim into the output
   ;; jar. With resource-paths etc. lein always seems to strip the first
   ;; directory from the source path when copying. Need to find an alternative or
   ;; implement a better version of verbatim-paths.
   (for [path  verbatim-paths
-        f     (fs/glob (fs/path (:root project) path) "**")
-        :when (not (fs/directory? f))]
+        f     (if (fs/directory? path)
+                (fs/glob (fs/path (:root project) path) "**")
+                [(fs/path (:root project) path)])
+        :when (fs/regular-file? f)]
     {:type  :bytes
      :path  (str (fs/relativize (:root project) f))
      :bytes (fs/read-all-bytes f)}))
+
+(defn build-dependencies->dependencies
+  "Compute regular :dependencies coordinates from the jank-specific
+  :build-dependencies coordinates.
+
+  We simply add a :scope 'jank-build' to the end of the coordinate, which
+  designates it as a build-time dependency."
+  [{:keys [build-dependencies] :as project}]
+  (mapv #(conj % :scope "jank-build") build-dependencies))
 
 (defn middleware
   "Inject jank project details into your current project."
   [project]
   (-> (deep-merge default-project project)
-      (update :filespecs concat (verbatim->filespecs project))))
+      (update :filespecs concat (verbatim->filespecs project))
+      (update :dependencies concat (build-dependencies->dependencies project))))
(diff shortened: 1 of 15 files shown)


jeaye:
This looks great. I will ask technomancy if he sees any issue with the scope usage for build deps.

lein-jank/src/leiningen/jank.clj
```diff
@@ -60,9 +56,7 @@
         project        (update project :jank ljb/merge-native-flags native-flags)]
     (if (:main project)
       (ljc/shell-out! project cp-str "compile" [(:main project)] args)
-      (do
-        (lmain/warn "No :main entrypoint for project.")
-        (lmain/exit 1)))))
+      (lmain/warn "No :main entrypoint for project."))))
```
kylc, line 59: Got rid of the `exit` calls here because some standard lein tasks call out to these unconditionally. For example `lein install` seems to always `lein compile` (that makes sense for building JVM JARs). In our case, if a library does not have a main, this will cause `lein install` to abort, which we don't want.

lein-jank/src/leiningen/jank/build.clj
```diff
@@ -154,34 +174,55 @@
...
+            ;; Output is fingerprinted on the source jar contents and all of the
+            ;; build steps which produce its descendant outputs. If any of these
+            ;; change then a fresh build is triggered.
+            out-fprint (fingerprint subtree-ops)
+            out-dir    (target-subdir target-dir "out" jar-name out-fprint)]
+        (into subtree-ops
+              [{:op     :extract-src
+                :dep    dep
+                :jar    (str src-jar)
+                :fprint (fingerprint-file src-jar)
```
kylc, line 219: There is a bit more to figure out with the fingerprinting to make sure rebuilds trigger whenever something changes. Consider this a mostly functional placeholder for now.

kylc:
> This looks great. I will ask technomancy if he sees any issue with the scope usage for build deps.

It's nice because the scope fits into the standard leiningen dependency tree functionality, so we can easily pick out build dependencies without any additional code. But if it's a no-go then we can implement our own logic to read the dependencies' project.clj and pull out the build-dependencies key.

kylc:
This is ready for review. I shoehorned some other improvements into this same PR, but they are somewhat organized as separate commits.

jeaye:
I asked technomancy and some others, on IRC, and heard back that our usage of dependency scopes here should be just fine.

lein-jank/test/leiningen/jank/integration_test.clj
```diff
@@ -0,0 +1,78 @@
+(ns leiningen.jank.integration-test
```
jeaye, line 1: Would you mind hooking this into CI for this PR? I think we'll need a new job for the lein-jank project. Maybe `.github/workflows/test-lein-jank.yml`.

Otherwise, I'll never remember to run this.

jeaye (changes requested):
Looks great to me! Just hoping to get the tests into CI.

lein-jank/test/leiningen/jank/integration_test.clj
kylc: Added a CI job to run `lein test`

jeaye (approved):
Incredible work, Kyle! I'm loving this direction and progress.

PR #817, opened 2026-06-19
Various jank-build fixups from initial testing

kylc:
- Use built-in lein argument parsing, now pass `--disable-sandbox` instead of `:disable-sandbox`
- Use a randomized /tmp build directory to avoid conflicts
- Implicitly include `jank-build.bb` in the jar if it exists in the project root


jeaye (approved):
Thank you!

PR #818, opened 2026-06-19
Initial `clojure.data.json` support

djblue:
- Add `read-str`
- And `write-str`

### DIFF
--- a/compiler+runtime/include/cpp/clojure/data/json_native.hpp
+++ b/compiler+runtime/include/cpp/clojure/data/json_native.hpp
@@ -21,9 +21,8 @@ namespace clojure::data::json_native
 
   struct write_options
   {
-    bool escape_unicode{ true };
-    bool escape_js_separators{ true };
-    bool escape_slash{ true };
+    // TODO: Add more read options supported by clojure.data.json
+    // https://github.com/clojure/data.json/blob/master/src/main/clojure/clojure/data/json.clj#L796-L806
     bool indent{ false };
     object_ref date_formatter{};
     object_ref key_fn{};
--- a/compiler+runtime/include/cpp/clojure/data/json_native.hpp
+++ b/compiler+runtime/include/cpp/clojure/data/json_native.hpp
@@ -6,6 +6,18 @@
 
 namespace clojure::data::json_native
 {
-  jtl::immutable_string write_str(jank::runtime::object_ref const x);
-  jank::runtime::object_ref read_str(jtl::immutable_string const &string);
+  using namespace ::jank::runtime;
+
+  struct read_options
+  {
+    bool eof_error{ true };
+    object_ref eof_value{ jank_nil };
+    bool bigdec{ false };
+    object_ref key_fn{ jank_nil };
+    object_ref value_fn{ jank_nil };
+    object_ref extra_data_fn{ jank_nil };
+  };
+
+  jtl::immutable_string write_str(object_ref const x);
+  object_ref read_str(jtl::immutable_string const &string, read_options const &opts);
 }
--- a/compiler+runtime/include/cpp/clojure/data/json_native.hpp
+++ b/compiler+runtime/include/cpp/clojure/data/json_native.hpp
@@ -11,11 +11,10 @@ namespace clojure::data::json_native
   struct read_options
   {
     bool eof_error{ true };
-    object_ref eof_value{ jank_nil };
     bool bigdec{ false };
-    object_ref key_fn{ jank_nil };
-    object_ref value_fn{ jank_nil };
-    object_ref extra_data_fn{ jank_nil };
+    object_ref eof_value{};
+    object_ref key_fn{};
+    object_ref value_fn{};
   };
 
   jtl::immutable_string write_str(object_ref const x);
--- a/compiler+runtime/include/cpp/clojure/data/json_native.hpp
+++ b/compiler+runtime/include/cpp/clojure/data/json_native.hpp
@@ -17,6 +17,18 @@ namespace clojure::data::json_native
     object_ref value_fn{};
   };
 
-  jtl::immutable_string write_str(object_ref const x);
   object_ref read_str(jtl::immutable_string const &string, read_options const &opts);
+
+  struct write_options
+  {
+    bool escape_unicode{ true };
+    bool escape_js_separators{ true };
+    bool escape_slash{ true };
+    bool indent{ false };
+    object_ref date_formatter{};
+    object_ref key_fn{};
+    object_ref value_fn{};
+  };
+
+  jtl::immutable_string write_str(object_ref const x, write_options const &opts);
 }
--- a/compiler+runtime/include/cpp/clojure/data/json_native.hpp
+++ b/compiler+runtime/include/cpp/clojure/data/json_native.hpp
@@ -1,0 +1,11 @@
+#pragma once
+
+#include <jtl/immutable_string.hpp>
+
+#include <jank/runtime/object.hpp>
+
+namespace clojure::data::json_native
+{
+  jtl::immutable_string write_str(jank::runtime::object_ref const x);
+  jank::runtime::object_ref read_str(jtl::immutable_string const &string);
+}
--- a/compiler+runtime/src/cpp/clojure/data/json_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/data/json_native.cpp
@@ -109,10 +109,16 @@ namespace clojure::data::json_native
 
           for(auto const &[k, v] : json.items())
           {
-            auto const key(make_box<obj::persistent_string>(k));
-            auto const value(read(v, opts));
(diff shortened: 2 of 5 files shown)


compiler+runtime/src/cpp/clojure/data/json_native.cpp
```diff
@@ -0,0 +1,128 @@
...
+    return visit_object(
+      [](auto const typed_o) -> nlohmann::json {
+        using T = typename decltype(typed_o)::value_type;
+
+        if constexpr(std::same_as<T, obj::nil>)
+        {
+          return {};
+        }
+
+        if constexpr(std::same_as<T, obj::boolean> || std::same_as<T, obj::persistent_string>
```
djblue: This is my best guess for doing this.

compiler+runtime/src/jank/clojure/data/json.jank
```diff
@@ -0,0 +1,42 @@
...
+(defn read-str
+  "Reads one JSON value from input String. Options are the same as for
+  read."
+  [string & {:as options}]
+  (cpp/clojure.data.json_native.read_str string))
+
+(defn write-str
+  "Converts x to a JSON-formatted string. Options are the same as
+  write."
+  ^String [x & {:as options}]
```
djblue: Still need to plugin options

compiler+runtime/src/jank/clojure/data/json.jank
```diff
@@ -0,0 +1,42 @@
...
+  (cpp/clojure.data.json_native.write_str x))
+
+(comment
+  (read-str "null")
```
djblue: How do we want to test this namespace? Is this something that should be in clojure-test-suite, or should the tests live in this repo?

compiler+runtime/src/jank/clojure/data/json.jank
```diff
@@ -1,11 +1,75 @@
...
+
 (defn read-str
   "Reads one JSON value from input String. Options are the same as for
   read."
   [string & {:as options}]
-  (cpp/clojure.data.json_native.read_str string))
+  (cpp/clojure.data.json_native.read_str
+   string
+   (let [options (merge default-read-options options)]
+     (cpp/clojure.data.json_native.read_options.
```
djblue, line 49: Is this a good approach for modeling these options for the cpp side of the code?

compiler+runtime/include/cpp/clojure/data/json_native.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace clojure::data::json_native
+{
+  using namespace ::jank::runtime;
+
+  struct read_options
+  {
+    bool eof_error{ true };
+    object_ref eof_value{ jank_nil };
+    bool bigdec{ false };
```
jeaye, line 14: Let's put the bools next to each other, so we don't waste 14 bytes for padding.

Also, there's no need for `{ jank_nil }` on these. `object_ref` has a default ctor which will initialize to nil. Explicitly doing it is just more code to compile.

compiler+runtime/src/cpp/clojure/data/json_native.cpp
```diff
@@ -0,0 +1,139 @@
...
+                     || std::same_as<T, obj::integer> || std::same_as<T, obj::small_integer>
+                     || std::same_as<T, obj::real> || std::same_as<T, obj::small_real>)
```
jeaye: Does Clojure support big int, big dec, and ratio here?

compiler+runtime/src/cpp/clojure/data/json_native.cpp
jeaye: `jtl::is_any_same` would be cleaner.

compiler+runtime/src/cpp/clojure/data/json_native.cpp
```diff
@@ -0,0 +1,139 @@
...
+        return jank_nil;
+      case nlohmann::json::value_t::string:
+        return make_box<obj::persistent_string>(json.get<std::string>());
+      case nlohmann::json::value_t::number_unsigned:
+      case nlohmann::json::value_t::number_integer:
+        return make_box(json.get<jtl::i64>());
+      case nlohmann::json::value_t::number_float:
+        return make_box(json.get<jtl::f64>());
+      case nlohmann::json::value_t::boolean:
+        return json.get<bool>() ? jank_true : jank_false;
```
jeaye: This is what `make_box(json.get<bool>())` will do.

compiler+runtime/src/jank/clojure/data/json.jank
```diff
@@ -0,0 +1,106 @@
+(ns clojure.data.json
+  (:refer-clojure :exclude (read))
+  (:include "clojure/data/json_native.hpp"))
+
+(def default-read-options {:bigdec false
+                           :key-fn nil
+                           :value-fn nil})
+
+(defn read
```
jeaye: Let's remove this fn. We don't have `Reader` and I don't think it makes sense for jank. That's just leaking Java into Clojure's APIs.

compiler+runtime/src/jank/clojure/data/json.jank
```diff
@@ -0,0 +1,106 @@
...
+      (:eof-value options)
+      (cpp/jank.runtime.truthy (:bigdec options))
+      (:key-fn options)
+      (:value-fn options)
+      (:extra-data-fn options)))))
+
+(defn write-str
+  "Converts x to a JSON-formatted string. Options are the same as
+  write."
+  ^String [x & {:as options}]
```
jeaye: Please remove all type hints. jank doesn't support them now, but if we ever do, the Java class types certainly won't work.

compiler+runtime/src/jank/clojure/data/json.jank
jeaye: The clojure-test-suite is the place to test this.

jeaye (changes requested):
Overall, looking good! Nice work, Chris.

compiler+runtime/src/cpp/clojure/data/json_native.cpp
djblue: Yes. It looks like [ratios are turned into doubles](https://github.com/clojure/data.json/blob/master/src/main/clojure/clojure/data/json.clj#L757-L758) and [big int, big dec are turned into strings](https://github.com/clojure/data.json/blob/master/src/main/clojure/clojure/data/json.clj#L703-L704). Looks like they also support writing UUIDs and insts. Should I do the same for jank as well?

jeaye: Yes please! Let's leave a TODO here for inst, since we don't yet support those.

djblue: We do support inst, [kinda](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/runtime/obj/inst.cpp#L18-L58) 😂 Since we have the object type already, should we just add it now?

jeaye: Ah, yep. We can add it then.

compiler+runtime/src/jank/clojure/data/json.jank
djblue: Are we okay with merging this before that is available?

compiler+runtime/include/cpp/clojure/data/json_native.hpp
```diff
@@ -0,0 +1,34 @@
...
+    object_ref eof_value{};
+    object_ref key_fn{};
+    object_ref value_fn{};
+  };
+
+  object_ref read_str(jtl::immutable_string const &string, read_options const &opts);
+
+  struct write_options
+  {
+    bool escape_unicode{ true };
```
djblue: I only implemented support for `indent`, `date_formatter`, `key_fn` and `value_fn`. I think this is still very usable without the other options, and I'm not sure exactly about their semantics. Do you think we need to implement them now, or can we wait until we have some tests in the clojure-test-suite? If we can defer the work, should I remove them from this struct?

compiler+runtime/src/jank/clojure/data/json.jank
jeaye: Yes.

compiler+runtime/include/cpp/clojure/data/json_native.hpp
jeaye: Let's remove them from the struct and replace them with a TODO. Leaving helpful TODOs is a great way to defer work for later without losing the intention.

djblue: Ok, I think this PR is in a good state and is ready for merge. Please let me know otherwise.

jeaye (approved):
Works for me!

PR #819, opened 2026-06-19
Update string_native to use specific types where possible

djblue:
Now that I've learned how trait analysis works in jank, I thought this would be a good namespace to cleanup.

### DIFF
--- a/compiler+runtime/src/jank/clojure/string.jank
+++ b/compiler+runtime/src/jank/clojure/string.jank
@@ -139,12 +139,12 @@
 (defn upper-case
   "Converts string to all upper-case."
   [s]
-  (cpp/clojure.string_native.upper_case s))
+  (cpp/clojure.string_native.upper_case (.to_string s)))
 
 (defn lower-case
   "Converts string to all lower-case."
   [s]
-  (cpp/clojure.string_native.lower_case s))
+  (cpp/clojure.string_native.lower_case (.to_string s)))
 
 (defn capitalize
   "Converts first character of the string to upper-case, all other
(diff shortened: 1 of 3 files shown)


djblue:
Once this is accepted, I will have another follow up for implementing `clojure.string/replace`.

compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -139,12 +139,12 @@
 (defn upper-case
   "Converts string to all upper-case."
   [s]
-  (cpp/clojure.string_native.upper_case s))
+  (cpp/clojure.string_native.upper_case (.to_string s)))
```
djblue, line 142: A subtlety in these `.to_string` coercions between jank and Clojure is how `nil` is handled. Calling `.toString` on `null` on the jvm throws an exception, therefore technically `nil` is not an accepted argument, but for jank we would return `NIL` for `str/upper-case`. This is not new behavior introduced by this PR. What behavior do we want for jank here? Do we want to add a special case to the [clojure-test-suite for jank for `nil`](https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/string_test/upper_case.cljc#L8)?

jeaye: Nope, we don't want special cases for these things. jank just needs to change its behavior to be more conforming.

jeaye: Is this change something which blocks this PR?

djblue: Since this PR isn't enabling the tests yet, and this is existing behavior, it's not a blocker in my opinion. However, when we do enable them, which I plan on doing in a future PR soon, we will need to have the explicit throws, which I don't mind adding then 👍

jeaye (approved):
Thanks!

jeaye:
I've made a note to add an IR optimization pass to make it so that when we convert from an object `O` into some native value `N`, we don't do it more than once. For example:

```clojure
(let [data (get-data!)]
  (cpp/foo data) ; takes a jtl::immutable_string
  (cpp/bar data) ; takes a jtl::immutable_string
)
```

In this case, we just want to do the conversion once and then re-use the result. Similarly, if we're going from native to jank object, we also want to dedup the conversions. That will ensure that APIs with unboxed types won't negatively affect performance.

PR #820, opened 2026-06-19
Implement `delay?` and cleanup return types

djblue:


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -5027,8 +5027,7 @@
 (defn delay?
   "returns true if x is a Delay created with delay"
   [x]
-  ;; (instance? clojure.lang.Delay x)
-  (throw "TODO: port delay?"))
+  (cpp/== cpp/jank.runtime.object_type.delay (.get_type x)))
```
djblue, line 5030: Do we prefer this to having the same function implemented in cpp?

jeaye: Yep.

jeaye (approved):
Thanks!

PR #821, opened 2026-06-19
Work around lein dep resolution deduplication

kylc:
If you have an actual dep tree like this:

```
 [my-gl-demo "2026.06-1"]
   [org.jank-lang.commons/gl-sys "2026.06-1"]
     [org.jank-lang.commons/jank-build-pkg-config "2026.06-1" :scope "jank-build"]
   [org.jank-lang.commons/glfw-sys "2026.06-1"]
     [org.jank-lang.commons/jank-build-pkg-config "2026.06-1" :scope "jank-build"]
```

Then `lein deps :tree` and also lein's internal dependency resolution functions will perform some deduplication on the tree and give you this:

```
 [my-gl-demo "2026.06-1"]
   [org.jank-lang.commons/gl-sys "2026.06-1"]
   [org.jank-lang.commons/glfw-sys "2026.06-1"]
     [org.jank-lang.commons/jank-build-pkg-config "2026.06-1" :scope "jank-build"]
```

Presumably this is to prevent duplicate classpath entires on the JVM (?). For jank, this causes the build to fail because gl-sys will not have its necessary build inputs. This PR updates the dependency resolution logic to use aether directly and avoid the deduplication. Adds a test with this duplicated structure to verify.

This fix can result in some duplicate flags passed to jank (e.g. `-lGL -lGL`), but I don't think this is a problem.

### DIFF
--- a/lein-jank/src/leiningen/jank/build.clj
+++ b/lein-jank/src/leiningen/jank/build.clj
@@ -4,8 +4,8 @@
             [clojure.java.io :as io]
             [babashka.fs :as fs]
             [leiningen.core.main :as lmain]
-            [leiningen.core.classpath :as lcp]
-            [leiningen.jank.sandbox.core :as sandbox])
+            [leiningen.jank.sandbox.core :as sandbox]
+            [leiningen.jank.resolve :as resolve])
   (:import (java.security MessageDigest)
            (java.util HexFormat)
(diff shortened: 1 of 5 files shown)


jeaye (approved):
Great catch! Thanks for the test, too.

PR #822, opened 2026-06-19
Remove second pass of pr-str for `gdb/jank.py`

djblue:
The current version of `gdb/jank.py` appears to be `pr-str`'ing values twice. You can see it most clearly in the below screenshot for the keys `:query-string`, `:uri` and `:body` since they are already strings. This PR fixes that by using `jank_to_code_string` directly.

Before:
before
After:
after


jeaye (approved):
Thanks!

PR #825, opened 2026-06-20
Set executable runpath to include all library dirs

kylc:
Allows executables to find the shared libraries with which they were compiled without needing to manually set `LD_LIBRARY_PATH`.

Tested with an imgui hello world project, which now automatically locates `libimgui.so` etc. from the `target/` directory.


jeaye (approved):
Ah, perfect. I had a TODO for this.

PR #826, opened 2026-06-20
Add --output flag support to lein-jank

kylc:


lein-jank/src/leiningen/jank/core.clj
```diff
@@ -26,6 +26,9 @@
     :output-dir
     ["--output-dir" (fs/path value project-name)]
 
+    :output
```
jeaye: Thanks! Right now, we're defaulting to `a.out`, which goes into the root project dir. I think the default should probably be `<output dir>/<project name>`. Thoughts on that?

jeaye: Updated the PR to contain this, using the middleware.

jeaye: We still write it to the project dir, but that's because `--output` currently expects a path instead of a name. We might want a jank change to always assume `--output` is a file within `--output-dir`, but that would rule out the possibility of being able to write it anywhere we want. I need to chew more on that.

kylc: Cargo ref: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries

In cargo you can only specify a binary filename, not a path. It always goes into the profile-specific target directory. That seems like a reasonable approach. Maybe we can call the parameter `--output-name` to make it more clear.

jeaye: Yep. I'll tackle that and update the book accordingly. I'll put everything in this PR so it all gets changed together.

jeaye: Hm, we'll need to reorganize a bit to make this happen. Right now, under the target dir, we include a dir named after the project. Within that dir, we also have a dir named within the project, assuming the project's namespace has the same name as the project. This means we can't put in a file with that name into either of those directories. Here's what we have now, from building a fresh `hello_lein` project (which failed, due to not being able to write `target/debug/hello_lein/hello_lein`, since it's already a dir).

```
❯ tree
.
├── LICENSE
├── project.clj
├── src
│   └── hello_lein
│       └── main.jank
├── target
│   ├── debug
│   │   └── hello_lein
│   │       └── hello_lein
│   │           └── main.o
│   └── stale
│       └── leiningen.core.classpath.extract-native-dependencies
└── test
    └── hello_lein
        └── main_test.jank

10 directories, 6 files
```

Here's what Cargo has, for the same name.

```
❯ tree
.
├── Cargo.lock
├── Cargo.toml
├── src
│   └── main.rs
└── target
    ├── CACHEDIR.TAG
    └── debug
        ├── build
        ├── deps
        │   ├── hello_lein-d5caa92a3c048014
        │   └── hello_lein-d5caa92a3c048014.d
        ├── examples
        ├── hello_lein
        ├── hello_lein.d
        └── incremental
            └── hello_lein-0cfdvw3znkol6
                ├── s-hjot1c45f1-1hhea5e-45jfxhhsiyklffmjqz3no8g3o
                │   ├── 19f35l5vfe0e2fubm9o1yp8av.o
                │   ├── 35d6vy0m2ciuf0uvtdu2833n3.o
                │   ├── 3ngwzvgqjylvnlb4noi5ijzz5.o
                │   ├── 5mivi67sd3shlm2ejkbzx2zku.o
                │   ├── a9aw5twrgk3cyi7t4a66frq2x.o
                │   ├── dep-graph.bin
                │   ├── e6jqflaapw8j2o1siesbnqi70.o
                │   ├── query-cache.bin
                │   └── work-products.bin
                └── s-hjot1c45f1-1hhea5e.lock

10 directories, 18 files
```

So, Cargo nests one level deeper before it starts putting intermediate files. Perhaps we should do the same. Maybe we just call that dir `cache`? So we have `target/debug/cache/...` for object files and then `target/debug/hello_lein` is the executable.

You're more in tune with this system, so let me know if you have some more insightful solution for this.

jeaye: My general feeling here is that we're not stuck with any of the current behavior and right now is the best time to make any changes we want. If it helps to lean even further into Cargo's design, we can do that.

kylc: `cache` seems fine. We may want to add some gobbledygook to the name to avoid conflicts in case someone wanted to name their binary `cache` as well. Will we also need to make sure that directory is placed on the jank module path so it picks up the .o files?

jeaye: I'll have an updated on this coming soon. Also, I have been testing this with the imgui packages in commons and have some fixes for them, too. Just so you're not also doing the same thing!

jeaye: Ok, I have revised jank's current design to more closely follow Cargo.

* `--output-dir` is now `--target-dir`, with a deprecation warning for the old `:output-dir` usage
  * We don't have any serious lein-jank users, that I know of, so I'm not concerned about annoying anyone with this
* There's a new `--build-dir` flag, which defaults to `<target dir>/_cache` in jank and `<target dir>/_cache/<project name>` in lein-jank
  * `_cache` ought to be unique enough to not cause trouble, but we can deal with it later if needed
  * All dep files are written into `_cache` via lein-jank and the jank compiler also has explicit support for this now
  * By having lein-jank set `--build-dir` to be nested, we address your original concerns about collisions between deps and the root project
*  The `-o` flag has been replaced with a `--name` flag, which names the output within the target dir
* lein-jank and lein-jank-template have been bumped to the new versioning scheme
  * I'll publish to Clojars once we merge

kylc: > By having lein-jank set --build-dir to be nested, we address your original concerns about collisions between deps and the root project

I think I am running into this issue again on main. My target directory tree looks like this:

```
target
├── release
│   ├── _cache
│   │   ├── ftxui-sys-7.0.0-out-nqpVYdXUNR-Ll8fhGwlaSw
│   │   │   ├── include
│   │   │   ├── jank-build-cache.txt
│   │   │   └── lib64
│   │   ├── ftxui-sys-7.0.0-src-UAIBIiQQYGX-DAIPf8DLYg
│   │   │   ├── jank-build.bb
│   │   │   ├── lib
│   │   │   └── META-INF
│   │   └── minitui
│   │       ├── demo.o
│   │       └── minitui
│   └── minitui
```

If I change a build flag, then jank-build runs (as it should) but then jank sees the flag change (new include and link dirs), prints the warning that it is removing build files, and then deletes the entire `_cache` directory including the jank-build artifacts. Effectively you have to build the project twice when changing flags to reach build flag fixity.

Perhaps we can separate the jank object files from the dep builds more like Cargo? Something like:

```
target
├── release
│   ├── _deps
│   │   ├── ftxui-sys-7.0.0-out-nqpVYdXUNR-Ll8fhGwlaSw
│   │   │   ├── include
│   │   │   ├── jank-build-cache.txt
│   │   │   └── lib64
│   │   ├── ftxui-sys-7.0.0-src-UAIBIiQQYGX-DAIPf8DLYg
│   │   │   ├── jank-build.bb
│   │   │   ├── lib
│   │   │   └── META-INF
│   ├── _cache
│   │   └── minitui
│   │       ├── demo.o
│   │       └── minitui
│   └── minitui
```

jeaye: Ah, crap. I was thinking about project files colliding with deps, but that's only half the problem. The other half is that automatic cleaning should only clean the project files. I will address this and note down that we need to have functional tests for cases like this.

Side note, ftxui-sys is a good candidate for commons!

jeaye: Ah, no. We just need to clean the build dir, not the target dir.

PR #827, opened 2026-06-20
Enable tap tests

djblue:
I think this PR will fail until [#808](https://github.com/jank-lang/jank/pull/808/changes#diff-b050f815fdacd8071666fb6bd2b8fae1523c8ad92095c27acd79a38fb70a5749R38) is merged or if I also disable `clojure.core-test.char`.

### DIFF
--- a/compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
+++ b/compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
@@ -206,7 +206,7 @@
     clojure.core-test.take-last
     ; clojure.core-test.take-nth ; FIXME: Failing tests.
     clojure.core-test.take-while
-    ; clojure.core-test.taps ; analyze/unresolved-symbol error: Unable to resolve symbol 'clojure.lang.IPending'.
+    clojure.core-test.taps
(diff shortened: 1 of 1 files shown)


jeaye (dismissed):
Thanks! I love seeing more jank progress on the clojure-test-suite.

djblue:
@jeaye how do you think we should address `clojure.core-test.char`? It fails with the following:

```clojure
Testing clojure.core-test.char

FAIL in (nil) (:)
unicode 4+ byte characters throw
expected: (= (first "𐅦") (char 65895))
  actual: (not (= \� \𐅧))
```

This is very much related to https://github.com/jank-lang/jank/pull/789#issue-4510377440.

> The one thing I noticed is that for Clojure, char blows up much sooner since I think the underlying jvm strings are 2 bytes per "char". Do we want jank to have the same limitation, or should we support more because we can and this is more of a platform limitation?

Should I just update char to throw like Clojure? Should the `clojure-test-suite` be encoding platform limitations as well?

jeaye:
Hm, looks like we have a failure due to updating the submodule.

```
FAIL in (nil) (:)
unicode 4+ byte characters throw
expected: (= (first "𐅦") (char 65895))
  actual: (not (= \� \𐅧))
```

We can disable this test for this PR, for further investigation, but we can't merge until CI is green.

djblue:
@jeaye I tried looking into a quick fix for the test but was unsuccessful. I disabled it for now as you suggested.

jeaye (approved):
Thanks!

PR #828, opened 2026-06-21
Add `clojure.string` tests from `clojure-test-suite`

djblue:


compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
```diff
@@ -227,6 +227,15 @@
     ; clojure.core-test.with-precision ; TODO: port with-precision, INFO: SKIP - defprotocol
     clojure.core-test.zero-qmark
     clojure.core-test.zipmap
+
+    ; clojure.string-test.blank-qmark ; error: String contains invalid escape sequence '\u'.
+    ; clojure.string-test.capitalize  ; Uncaught exception: wstring_convert::from_bytes
+    clojure.string-test.ends-with-qmark
+    ; clojure.string-test.escape ; Uncaught exception: Can't convert character to integer.
```
djblue, line 234: I think this is a new issue as before [`(\A) -> 65`](https://github.com/jank-lang/jank/pull/790) used to work. Now I get the following:

```clojure
% jank repl
user=> (int \A)
Uncaught exception: Can't convert character to integer.
user=> (#(int %) \A) ;; but this works still
65
```

jeaye (approved):
Nice!

compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
jeaye: Yep. Without the tests enabled, the behavior likely changed indirectly. We'll get it fixed and then get the test enabled.

PR #829, opened 2026-06-21
Implement `clojure.string/replace`

djblue:
Depends on https://github.com/jank-lang/clojure-test-suite/pull/913. ~I still need to figure out how I want to add regex support, but strings/char are working currently.~


compiler+runtime/include/cpp/clojure/string_native.hpp
```diff
@@ -25,6 +26,19 @@ namespace clojure::string_native
   object_ref
   replace_first(object_ref const s, object_ref const match, object_ref const replacement);
 
+  jtl::immutable_string replace(jtl::immutable_string const &s,
```
djblue, line 29: Something new I'm trying in this PR is dispatching directly from jank into these overloads. Please let me know if this is a bad idea 😂

compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -54,17 +53,36 @@
...
-      (instance? Pattern match) (if (instance? CharSequence replacement)
-                                  (.replaceAll (re-matcher match s)
-                                               (.toString replacement))
-                                  (replace-by s match replacement))
-      :else (throw (IllegalArgumentException. (str "Invalid match arg: " match)))))
-  (throw "TODO: port clojure.string/replace"))
+      (char? match)
+      (cpp/clojure.string_native.replace
+       s
+       (cpp/jank.runtime.to_char match)
```
djblue, line 50: I tried casting to a `character_ref` here, but jank failed saying they overload was ambiguous. Using a function which returns a `character_ref` did however work.

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -170,6 +171,79 @@ namespace clojure::string_native
...
   }
 
+  jtl::immutable_string replace(jtl::immutable_string const &s,
+                                jtl::immutable_string const &match,
+                                jtl::immutable_string const &replacement)
+  {
+    auto const s_size(s.size());
+    auto const match_size(match.size());
+
+    if(match_size == 0)
```
djblue, line 192: Initially, I assumed this should return the original string, but the jvm appears to interpose the `replacement` between each character in `s` 🤷

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -170,6 +171,79 @@ namespace clojure::string_native
...
+
+  jtl::immutable_string replace(jtl::immutable_string const &s,
+                                obj::character_ref const match,
+                                obj::character_ref const replacement)
+  {
+    return replace(s, match->data, replacement->data);
+  }
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
```
jeaye: Please don't use this. Just remove the names from the params here in the definition.

compiler+runtime/src/jank/clojure/string.jank
jeaye: Yeah, that's more of a Java-style OOP cast. We don't currently use casting for that, but I can see how someone would want to. Noting this down as a UX improvement.

compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -54,17 +53,38 @@
...
+       (cpp/jank.runtime.to_char replacement))
+
+      (string? match)
+      (cpp/clojure.string_native.replace
+       s
+       (cpp/jank.runtime.to_string match)
```
jeaye: `.to_string` is a member on every object. We want to call that instead. The `jank.runtime.to_string` fn just does that. It's a leftover from when we had a closed object model.

compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -54,17 +53,38 @@
   (clojure.string/replace \"Almost Pig Latin\" #\"\\b(\\w)(\\w+)\\b\" \"$2$1ay\")
   -> \"lmostAay igPay atinLay\""
   [s match replacement]
-  #_
-  (let [s (.toString s)]
+  (when (nil? s)
+    (throw "replace does not support nil"))
+  (let [s (cpp/cast jtl.immutable_string (str s))]
     (cond
```
jeaye, line 46: Seems like a `case` on the type might be a cleaner way to handle this.

jeaye (changes requested):
Nice start!

compiler+runtime/src/jank/clojure/string.jank
djblue: I tried doing a case based on type, something like:

```clojure
(case (.get_type match)
  cpp/jank.runtime.object_type.character ...
  cpp/jank.runtime.object_type.persistent_string ...)
```

And I got the following error:

```clojure
ERROR in (nil) (:)
Replace can produce an empty string
expected: (= "" (str/replace "x" "x" ""))
  actual: "No matching clause: 9"
```

It looks like `persistent_string` has index `9` of the `object_type` enum, so this must be a type mismatch issue, no?

djblue:
I think perhaps we can scope this PR to char/string, then add support for regex in a follow-up PR.

jeaye:
> I think perhaps we can scope this PR to char/string, then add support for regex in a follow-up PR.

Sure, however you'd like to handle this on the jank side.

djblue:
@jeaye I forgot to comment that I added regex support when I pushed d7c0d6150da6b41baf53bf89dd061180c0bc8fc0. So all the `clojure.string/replace` tests should work for jank now.

jeaye:
> @jeaye I forgot to comment that I added regex support when I pushed [d7c0d61](https://github.com/jank-lang/jank/commit/d7c0d6150da6b41baf53bf89dd061180c0bc8fc0). So all the `clojure.string/replace` tests should work for jank now.

Nice! So, anything else we need to tackle here?

djblue:
Just pushed e0ee81ce48ba56a7fe9a8e0af734c86c8ab4697a to enable `clojure.string-test.replace`, I thought I already did that, but I guess not. So once we get a :green_circle: CI, we should be good.

jeaye (approved):
Thanks, Chris!

PR #831, opened 2026-06-21
Implement clojure.core/format on std::format

kylc:
Adds support for `clojure.core/format` via the `std::format` library. This PR supports all features of `std::format` as of C++20, including variable width and precision specifiers and manually indexed arguments.

We work around the `std::format` limitation that the format args arity must be known at compile time by parsing the format string ahead of time into individual top-level format specifiers, and then passing them into `std::format` one-by-one.

Example usage:

```
(defn p [fmt & args]
  (println (apply format fmt args)))

(p "Look {2} {1} {0}" "these" "are" "backwards")
;; => "Look backwards are these"
(p "repeat {0} {0} {0}" "again")
;; => "repeat again again again"

(p "{{ escaped }}")
;; => "{ escaped }"
(p "Hello {}" "kyle")
;; => "Hello kyle"
(p "pi is {:+.{}f}" cpp/std.numbers.pi 20)
;; => "pi is +3.14159265358979311600"
(p "{:-<10}{:+>10}" "left" "right")
;; => "left------+++++right"

(p "hex: {:#06x}" 1234)
;; => "hex: 0x04d2"
(p "bin: {:#06b}" 1234)
;; => "bin: 0b10011010010"
(p "sci: {:.2e}" 100000.0)
;; => "sci: 1.00e+05"

(p "{:^40}" "---------------")
(p "{:^40}" "| hello world |")
(p "{:^40}" "---------------")
;; => "            ---------------             "
;; => "            | hello world |             "
;; => "            ---------------             "

(p "localized: {:L}" 1000000)
;; => "localized: 1,000,000"

(p "vector: {:>30}" [1 2 3])
;; => "vector:                        [1 2 3]"
(p "set:    {:>30}" #{1 2 3})
;; => "set:                          #{1 3 2}"
(p "map:    {:>30}" {:a 1 "b" 2})
;; => "map:                       {:a 1, b 2}"
(p "list:   {:>30}" '(1 2 "hello"))
;; => "list:                      (1 2 hello)"

(p "{0:},{0:+},{0:-},{0: }" 1)
;; => "1,+1,1, 1"
```

### DIFF
--- a/compiler+runtime/include/cpp/clojure/core_native.hpp
+++ b/compiler+runtime/include/cpp/clojure/core_native.hpp
@@ -76,7 +76,5 @@ namespace clojure::core_native
   jtl::u32 hash_unordered(object_ref const coll);
   jtl::immutable_string jank_version();
 
-  jtl::immutable_string format(jtl::immutable_string const &format, object_ref const args);
-
   obj::delay_ref delay(object_ref const fn);
 }
--- a/compiler+runtime/include/cpp/clojure/core_native.hpp
+++ b/compiler+runtime/include/cpp/clojure/core_native.hpp
@@ -76,5 +76,7 @@ namespace clojure::core_native
   jtl::u32 hash_unordered(object_ref const coll);
   jtl::immutable_string jank_version();
 
+  jtl::immutable_string format(jtl::immutable_string const &format, object_ref const args);
+
   obj::delay_ref delay(object_ref const fn);
 }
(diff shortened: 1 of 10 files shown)


kylc:
This is ready for review. Since last time we talked I added support for manual argument indexing in the format specifiers. It should now match C++20 std::format behavior in all cases, as if you passed the arguments statically, unless there are bugs.

jeaye (approved):
Looks great! I pushed some cleanup, but otherwise good if CI is green.

PR #836, opened 2026-06-24
Improve jank-build input change detection/caching

kylc:
Fixes various issues culminating in build scripts not rerunning when they should. For example
- A JAR input changing contents, but not changing name
- Build flag changes, such as optimization level

Changes the hash computation and string encoding function to more closely match that of Nix, mostly to avoid very long path names which can cause problems on some systems.

Also establishes a predefined set of inputs which will be passed from the build system into the build script, so build script authors know what to expect.

Cleans up verbose build handling to reuse the existing -v flag. It's not a build input, so we don't want to trigger a rebuild if just that flag changes.

### DIFF
--- a/lein-jank/src/leiningen/jank/build.clj
+++ b/lein-jank/src/leiningen/jank/build.clj
@@ -7,19 +7,21 @@
             [leiningen.jank.sandbox.core :as sandbox]
             [leiningen.jank.resolve :as resolve])
   (:import (java.security MessageDigest)
-           (java.util HexFormat)
+           (java.util Base64)
            (java.util.jar JarFile)))
 
 (def ^:dynamic *disable-sandbox* false)
+
+(def ^:dynamic *verbose-build* false)
 
 (def jank-build-file "jank-build.bb")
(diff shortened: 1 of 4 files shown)


lein-jank/src/leiningen/jank/build.clj
```diff
@@ -7,25 +7,26 @@
...
 (def jank-build-cache-file "jank-build-cache.txt")
 
 (def default-build-opts
   {:target-dir         "target"
-   :optimization-level 2
-   :static-build       true
-   :verbose-build      false})
+   :optimization-level 3
+   ;; TODO: enable when jank can link to .a files via -L and -l flags.
+   :static-build       false})
```
kylc, line 24: This is about the discussion we had on Slack where once jank supports linking static libraries by `-l:xyz` then we will add a `jank-build::link-static-lib=xyz` build directive. Once that's all done maybe we will enable static linking by default.

jeaye: Yep, makes sense. I agree, static linking is a better default.

jeaye (approved):
Nice work!

PR #837, opened 2026-06-24
Only clean build dir when binary version changes

jeaye:
@kylc I have tested this with your `improve-build-hashing` branch and the double-build issue is gone.

The one key effect of this is that final artifacts, in the target dir, are not cleaned. I think maybe that's a good thing, though.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/module/loader_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/module/loader_dynamic.cpp
@@ -122,7 +122,7 @@ namespace jank::runtime::module
 
   static jtl::immutable_string binary_version_path()
   {
-    auto const path{ util::format("{}/.jank-binary-version", util::cli::opts.target_dir) };
+    auto const path{ util::format("{}/.jank-binary-version", util::cli::opts.build_dir) };
     return path;
   }
 
@@ -138,7 +138,7 @@ namespace jank::runtime::module
 
   static binary_version_status check_binary_version_status()
   {
-    if(!std::filesystem::exists(util::cli::opts.target_dir.c_str()))
+    if(!std::filesystem::exists(util::cli::opts.build_dir.c_str()))
     {
       return binary_version_status::needs_write;
     }
@@ -164,7 +164,7 @@ namespace jank::runtime::module
 
   static void write_binary_version()
   {
-    std::filesystem::create_directories(util::cli::opts.target_dir.c_str());
+    std::filesystem::create_directories(util::cli::opts.build_dir.c_str());
 
     auto const &binary_version{ util::binary_version() };
     auto const &path{ binary_version_path() };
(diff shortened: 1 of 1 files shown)


compiler+runtime/src/cpp/jank/runtime/module/loader_dynamic.cpp
```diff
@@ -164,17 +164,17 @@ namespace jank::runtime::module
...
     auto const &path{ binary_version_path() };
     std::ofstream ofs{ path.c_str() };
     ofs << binary_version.c_str();
   }
 
-  static void clean_target_directory()
+  static void clean_build_directory()
   {
-    std::filesystem::remove_all(util::cli::opts.target_dir.c_str());
+    std::filesystem::remove_all(util::cli::opts.build_dir.c_str());
```
kylc, line 177: Is there any restriction on the build_dir, like that it's a child of the CWD? If I run `jank --build-dir .` in my project root will it delete the project source tree? 😰

jeaye: There is no such protection right now. The jank binary has no notion of projects, so we'd just be guessing at the project dir (i.e. CWD).

Maybe this whole thing is cleaner (no pun intended) if we move build dir cleaning into lein-jank? That way, without lein-jank, jank itself will clean nothing. The difficulty here is that the binary version is used to determine whether or not to clean and currently the binary version comes from the jank binary. Maybe we add a command to print the binary version, given a set of flags, and then lein-jank just uses that to manage everything?

Then we could provide some assurance that target-dir and build-dir are subdirs of the current project. The only problem I foresee with that is folks who want their build dir to be in `/tmp` or something. At that point, all bets are off. Maybe we could later permit that with an additional flag like `:allow-out-of-source-build? true` in the project.

kylc: > Maybe this whole thing is cleaner (no pun intended) if we move build dir cleaning into lein-jank?

Will this cause problems if you aren't using lein-jank, just jank directly?

Something else to consider, in your cargo example from the other PR, there is some hashing going on in the prefix path of the object files:

```
...
    └── debug
        ├── build
        ├── deps
        │   ├── hello_lein-d5caa92a3c048014
        │   └── hello_lein-d5caa92a3c048014.d
        ├── examples
        ├── hello_lein
        ├── hello_lein.d
        └── incremental
            └── hello_lein-0cfdvw3znkol6 <-------- here (and below)
                ├── s-hjot1c45f1-1hhea5e-45jfxhhsiyklffmjqz3no8g3o
                │   ├── 19f35l5vfe0e2fubm9o1yp8av.o
                │   ├── 35d6vy0m2ciuf0uvtdu2833n3.o
                │   ├── 3ngwzvgqjylvnlb4noi5ijzz5.o
                │   ├── 5mivi67sd3shlm2ejkbzx2zku.o
                │   ├── a9aw5twrgk3cyi7t4a66frq2x.o
                │   ├── dep-graph.bin
                │   ├── e6jqflaapw8j2o1siesbnqi70.o
                │   ├── query-cache.bin
                │   └── work-products.bin
                └── s-hjot1c45f1-1hhea5e.lock
```

Maybe jank doesn't need to delete files at all, it just loads/dumps them into a binary version hash -affixed directory like `<opts.build_dir>/<binary_version_hash>/thing.o`. If the hash changes then the old incompatible files are effectively forgotten.

This discussion doesn't need to block this PR though.

PR #840, opened 2026-06-25
Enable `keys`/`vals` tests

djblue:


compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -2292,19 +2292,23 @@
...
+   (persistent!
+    (reduce (fn [acc kv]
+              (conj! acc (first kv)))
+            (transient [])
+            m))))
 
 (defn vals
   "Returns a sequence of the map's values, in the same order as (seq map)."
   [m]
   ; TODO: Use a proper val seq instead.
```
jeaye, line 2305: We removed the TODO for `keys`. Should we remove this one?

djblue: I don't think I removed the TODO for `keys` 🤔

jeaye: 🤦 You're right. I misread.

jeaye (approved):
Works for me!

PR #841, opened 2026-06-26
Enable `clojure.core-test.realized-qmark`

djblue:
I had to update some implementation details to get the tests passing for [`clojure.core-test.realized-qmark`](https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/realized_qmark.cljc):

- Calling `realized?` immediately after `future-cancel` on a `future` needs to return `true` if it was successful
  - the current implementation would eventually return true, but not immediately after.
- `lazy_sequence` had an incorrect implementation of `is_realized`
  - `fn.is_nil()` is `true` after the lazy-seq has been realized
- `future-cancel` should return a `bool` indicating success or failure
  - not strictly needed for the tests to pass, but a detail I observed in getting the tests working

Depends on https://github.com/jank-lang/clojure-test-suite/pull/917 for tests to work/pass.

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core.cpp
@@ -971,17 +971,22 @@ namespace jank::runtime
     return ret;
   }
 
-  void cancel_future(obj::future_ref const future)
+  bool cancel_future(obj::future_ref const future)
   {
     /* We need to hold this lock the whole time we're checking, to ensure the thread
      * doesn't finish while we're here checking. */
-    auto const locked_state{ future->state.rlock() };
+    auto const locked_state{ future->state.wlock() };
     if(locked_state->status == obj::future_status::running)
     {
       auto const locked_thread{ future->thread.wlock() };
       auto const thread_handle{ locked_thread->native_handle() };
-      pthread_cancel(reinterpret_cast<pthread_t>(thread_handle));
+      if(pthread_cancel(reinterpret_cast<pthread_t>(thread_handle)) == 0)
+      {
+        locked_state->status = obj::future_status::cancelled;
+        return true;
+      }
     }
+    return false;
   }
 
   bool is_future_cancelled(obj::future_ref const future)
--- a/compiler+runtime/include/cpp/jank/runtime/core.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/core.hpp
@@ -134,7 +134,7 @@ namespace jank::runtime
   object_ref remove_watch(object_ref const reference, object_ref const key);
 
   obj::future_ref future(object_ref const fn);
-  void cancel_future(obj::future_ref const future);
+  bool cancel_future(obj::future_ref const future);
   bool is_future_cancelled(obj::future_ref const future);
 
   obj::promise_ref promise();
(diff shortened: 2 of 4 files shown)


compiler+runtime/src/cpp/jank/runtime/core.cpp
```diff
@@ -971,17 +971,22 @@ namespace jank::runtime
...
     if(locked_state->status == obj::future_status::running)
     {
       auto const locked_thread{ future->thread.wlock() };
       auto const thread_handle{ locked_thread->native_handle() };
-      pthread_cancel(reinterpret_cast<pthread_t>(thread_handle));
+      if(pthread_cancel(reinterpret_cast<pthread_t>(thread_handle)) == 0)
+      {
+        locked_state->status = obj::future_status::cancelled;
+        return true;
+      }
```
jeaye, line 987: So, the reason we didn't do this before is that simply calling `pthread_cancel` doesn't actually mean anything gets cancelled. Certainly not synchronously. From the docs:

> The above steps happen asynchronously with respect to the
> pthread_cancel() call; the return status of pthread_cancel()
> merely informs the caller whether the cancelation request was
> successfully queued.

Is there a requirement from the test suite that we mark this as cancelled right away?

djblue: Yeah, right [here](https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/realized_qmark.cljc#L46-L49).

djblue: If this is a platform constraint, we could update the test for jank to poll for cancel with a timeout. Would that be better? I think I misunderstood the return value from pthread_cancel for sure.

jeaye: Ok, gotcha. Ultimately, this is not a portable test. In a VM, with a managed thread implementation, they can control cancellation in a synchronous way. However, in the native world, that's not actually viable. In fact, on macOS, it's not even possible to know if a thread was cancelled at all (see `is_future_cancelled`).

So, if we accept the change you have, we're not actually saying the thread was cancelled. We're saying that we tried to cancel the thread. It could still be running, or it could be done but not cancelled. So far, in these cases, I have avoided optimistically setting state, since I wanted the state to represent reality, rather than intention.

For example, with the test you linked, the thread is sleeping. If we "cancel" it and then say the thread is realized, and then we try to deref it, we may actually end up blocking for the entire sleep. This is even worse when we imagine a thread which is infinitely looping, whereupon a deref on it will block forever. So I would rather keep the state as close to reality as possible, to avoid these scenarios.

djblue: One thing I've learned is that the following code is just not interruptible:

```clojure
(cpp/std.this_thread.sleep_for
 (cpp/std.chrono.milliseconds
  (cpp/jank.runtime.to_int ms)))
```

So if we want the future/sleep/cancel test to work for jank, we need to provide a sleep implementation that supports `pthread_cancel` via [`pthread_testcancel`](https://man7.org/linux/man-pages/man3/pthread_testcancel.3.html) which creates a cancelation point. Do we want to do that or mark the test as platform specific?

jeaye: Hm, I'd be open to a sleep fn which can be cancelled, and you're welcome to explore it. If that doesn't sound appealing right now, we can skip this test for jank. Outside of a VM, this sort of thing is much harder to accomplish.

djblue: Ok, I'll close this for now, and may open a PR at a later time.

jeaye: Thank you for the hard work, Chris!

PR #844, opened 2026-06-26
Enable `clojure.core-test.drop/-while`

djblue:


jeaye (dismissed):
Good catch!

jeaye:
There's a conflict, due to all of the merged PRs. :grin:

PR #848, opened 2026-06-27
Explicitly fail `rem` when div is 0

djblue:
Before:
```clojure
% jank repl        
nREPL server started on port 44830 on host 127.0.0.1 - nrepl://127.0.0.1:44830
user=> (rem 1 0)
Assertion failed! is_tagged_pointer(val)
...
zsh: abort      jank repl
```

After:

```clojure
% jank repl
nREPL server started on port 43504 on host 127.0.0.1 - nrepl://127.0.0.1:43504
user=> (rem 1 0)
Uncaught exception: Illegal divide by zero in 'rem'

...

user=> ;; jank process is fine
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/math.cpp
@@ -177,8 +177,15 @@ namespace jank::runtime
             }
             else
             {
-              auto const l_real{ to_real(typed_l_data) };
               auto const r_real{ to_real(typed_r->data) };
+
+              if(r_real == 0)
+              {
+                throw std::runtime_error{ "Illegal divide by zero in 'rem'" };
+              }
+
+              auto const l_real{ to_real(typed_l_data) };
+
               return make_box(std::fmod(l_real, r_real)).erase();
             }
           },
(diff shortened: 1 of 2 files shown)


compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -177,8 +177,15 @@ namespace jank::runtime
             }
             else
             {
-              auto const l_real{ to_real(typed_l_data) };
               auto const r_real{ to_real(typed_r->data) };
+
+              if(r_real == 0)
+              {
+                throw std::runtime_error{ "Illegal divide by zero in 'rem'" };
```
shantanu-sardesai, line 184: @jeaye

This is more of a comment for me but I've been wondering, which type of exception should we `throw` inside the `jank::runtime`? Should it be a jank string or a `std::exception` inheritor?

shantanu-sardesai (commented):
A question for Jeaye from my side. 🙂

djblue:
If https://github.com/jank-lang/clojure-test-suite/pull/919 gets merged, we can also enable `clojure.core-test.take-nth`.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
jeaye: The answer is based on what we catch in `JANK_CATCH` in `util/try.cpp`. In general, the preference should be:

1. An `error_ref`, is there is a corresponding error we can create
2. A child of `std::exception`, otherwise
3. Avoid throwing anything else

jeaye (approved):
Nice!

PR #856, opened 2026-06-29
Enable `clojure.core-test.get-in`

djblue:
The fix to get-in mirrors the jvm implementation [here](https://github.com/clojure/clojure/blob/clojure-1.12.5/src/clj/clojure/core.clj#L6218).

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
@@ -597,7 +597,7 @@ namespace jank::runtime
 
   object_ref get_in(object_ref const m, object_ref const keys, object_ref const fallback)
   {
-    static auto const sentinel(make_box<obj::cons>(jank_nil, jank_nil));
+    static auto const sentinel(make_box<obj::reduced>(jank_true));
 
     if(m.has_behavior(object_behavior::get))
     {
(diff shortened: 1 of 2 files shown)


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -597,19 +597,20 @@ namespace jank::runtime
 
   object_ref get_in(object_ref const m, object_ref const keys, object_ref const fallback)
   {
+    static auto const sentinel(make_box<obj::cons>(jank_nil, jank_nil));
```
djblue: I assume this is the easiest way to construct a sentinel object.

jeaye: We don't have a good solution for creating sentinels. Maybe a helper would be nice? Something like `jank::runtime::create_sentinel`.

Using `cons` works. Something lighter like `var_unbound_root` or `reduced` or `volatile_` has just one pointer within.

jeaye (dismissed):
Works for me! Up to you if you want to add a helper for this, based on my comment.

compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
djblue: I think we can add `jank::runtime::create_sentinel` if we find ourselves needing a sentinel elsewhere.

jeaye (approved):
Thanks!

PR #858, opened 2026-06-29
Enable `clojure.core-test.compare`

djblue:
I had to re-work compare a little to handle `small_` numbers as follows:

| Left                | Right                | Operation                       |
|----------------|-----------------|--------------------------- |
|`small_integer` | `small_integer`| `(r < l) - (l < r)`                |
|`small_integer` | `small_real`      | `(r < l) - (l < r)`                |
|`small_integer` | `object_ref`      | `l.compare(*o.ptr())`        |
|`small_real`      | `small_integer` | `(r < l) - (l < r)`                |
|`small_real`      | `small_real`      | `(r < l) - (l < r)`                 |
|`small_real`      | `object_ref`      | `l.compare(*o.ptr())`         |
|`object_ref`      | `small_integer`| `ptr()->compare(i)`           |
|`object_ref`      | `small_real`     | `ptr()->compare(r)`           |
|`object_ref`      | `object_ref`     | `ptr()->compare(*o.ptr())` |

It's mostly the same as the existing code, just slightly reordered to handle the following issues:

Before:
```clojure
% jank repl
user=> (compare 0 -100N) ;; we were swapping L and R here
-1
user=> (compare 1.0 0) ;; we weren't handling this case
Assertion failed! is_tagged_pointer(val)
```

After
```clojure
% jank repl               
user=> (compare 0 -100N)
1
user=> (compare 1.0 0)
1
```

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/visit.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/visit.hpp
@@ -262,9 +262,9 @@ namespace jank::runtime
     }
     else
     {
-      jtl::panic("invalid object type: {}, raw value {}",
-                 object_type_str(erased.get_type()),
-                 static_cast<int>(erased.get_type()));
+      throw std::runtime_error{ util::format("invalid object type: {}, raw value {}",
+                                             object_type_str(erased.get_type()),
+                                             static_cast<int>(erased.get_type())) };
     }
   }
 
--- a/compiler+runtime/src/cpp/jank/runtime/core/equal.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/core/equal.cpp
@@ -32,11 +32,6 @@ namespace jank::runtime
 
   i64 compare(object_ref const l, object_ref const r)
   {
-    if(l == r)
-    {
-      return 0;
-    }
-
     if(l.is_some())
     {
       if(r.is_nil())
--- a/compiler+runtime/include/cpp/jank/runtime/obj/character.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/character.hpp
@@ -26,6 +26,12 @@ namespace jank::runtime::obj
     jtl::immutable_string to_code_string() const override;
     uhash to_hash() const override;
 
+    /* behavior::comparable */
+    i64 compare(object const &) const override;
+
+    /* behavior::comparable extended */
+    i64 compare(character const &) const;
+
     /* Character does not fully support `behavior::number_like`, but can be converted to an integer. */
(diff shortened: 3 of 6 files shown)


compiler+runtime/include/cpp/jank/runtime/visit.hpp
```diff
@@ -262,9 +262,9 @@ namespace jank::runtime
     }
     else
     {
-      jtl::panic("invalid object type: {}, raw value {}",
-                 object_type_str(erased.get_type()),
-                 static_cast<int>(erased.get_type()));
+      throw std::runtime_error{ util::format("invalid object type: {}, raw value {}",
```
djblue, line 265: I'm assuming we don't need to panic here.

compiler+runtime/src/cpp/jank/runtime/core/equal.cpp
```diff
@@ -32,11 +32,6 @@ namespace jank::runtime
 
   i64 compare(object_ref const l, object_ref const r)
   {
-    if(l == r)
```
djblue, line 35: This was causing issues for the following code: 

```clojure
(compare #{0} #{0}) ;; 0
````

because it was bypassing the `behavior::comparable` check.

compiler+runtime/include/cpp/jank/runtime/visit.hpp
jeaye: If we get an invalid object type, we have undefined behavior. At that point, memory is corrupted for at least one object and we have no idea what else is wrong. That's why we panic.

jeaye (changes requested):
I think we should keep the panic. Otherwise, looks good to me!

compiler+runtime/include/cpp/jank/runtime/visit.hpp
djblue: The main issue is that [`persistent_vector::compare`](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/runtime/obj/persistent_vector.cpp#L133-L137) will panic anytime we try to compare it to a non-vector. Should we update `persistent_vector::compare` instead?

jeaye: Oh, my bad. This is for `visit_type`. That should just behave as `try_object`. In fact, `try_object` is a much better tool. I was too visit-happy when I wrote `visit_type`.

So, yes, we can throw in this case, but ideally we should actually just use `try_object`.

jeaye (approved):

PR #862, opened 2026-06-30
Enable `clojure.core-test.update`

djblue:


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -549,7 +549,7 @@ namespace jank::runtime
         else
         {
           throw std::runtime_error{ util::format("not associatively writable: {}",
-                                                 typed_m.to_code_string()) };
+                                                 object_type_str(typed_m.get_type())) };
```
djblue, line 552: So the test was hanging because we were printing an infinite seq. I wonder if we should just avoid printing values for exception messages, or truncate printing for infinite seqs 🤔 `clojure.core-test.juxt` has the exact same problem.

jeaye: Yeah, this is a known issue with our errors. We should be printing object types, not code strings. I've been updating them as I get to them, but there are a lot like this.

jeaye (approved):
Nice work!

PR #863, opened 2026-06-30
Enable `clojure.core-test.ifn-qmark`

djblue:


jeaye (approved):
I didn't realize symbols could do this. Thanks!

PR #864, opened 2026-06-30
Enable `clojure.core-test.mapcat`

djblue:


compiler+runtime/src/cpp/jank/runtime/core/call.cpp
```diff
@@ -140,29 +141,9 @@ namespace jank::runtime
...
-                                   a2,
-                                   a3,
-                                   a4,
-                                   a5,
-                                   a6,
-                                   a7,
-                                   a8,
-                                   a9,
-                                   a10,
-                                   try_object<obj::persistent_list>(*(++it)));
```
djblue, line 153: The test was hanging because were were trying to turn an infinite seq into a persistent list.

jeaye: Hm, that's not what this code does. `try_object` just checks and object's type and does a cast. It doesn't construct a new list or iterate a sequence.

compiler+runtime/src/cpp/jank/runtime/core/call.cpp
```diff
@@ -140,29 +141,9 @@ namespace jank::runtime
...
-                                 a3,
-                                 a4,
-                                 a5,
-                                 a6,
-                                 a7,
-                                 a8,
-                                 a9,
-                                 a10,
-                                 obj::persistent_list::create((++it).data));
+              return source.call(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, seq((++it).data));
```
jeaye: More likely this is where it was hanging. I don't think this is correct, though. Firstly, because you're doing the same thing in both branches now. So there's no point in our branch anymore. The branch is what I just added yesterday to fix the arg packing issue with variadic fns.

djblue: They are applying seq on different parts of the args iterator: `seq(*(++it))` vs `seq((++it).data)` so the fix is still applied. I can revert the other branch and see if it still works 👍

jeaye: Doesn't `operator*` on the iterator just return `data`? I would expect these to do the same.

jeaye: Oh, `operator*` calls `first`.

djblue: I'm not sure, but I assumed one was returning the element at the position and the other was returning the head. Is that an invalid assumption?

jeaye: Yeah, I think we can/should keep the `try_object`, since we are expecting a list there. We're expecting it has been packed during analysis.

djblue: Yeah, seq is only needed for the second branch. Should I revert other branch?

djblue: We actually don't even need seq, we can use `(++it).data` directly, is that better?

jeaye: Yes.

jeaye (approved):
Works for me!

PR #865, opened 2026-06-30
Stop cleaning build dir; use binary version

jeaye:
This change removes the binary version check from jank, so we'll no
longer try to detect out-of-date build directories. Instead, we'll just
append the binary version to our build directly and happily put things
in it. If our binary version changes, we'll use the new directory,
leaving the old one in place. `lein clean` with lein-jank will remove
the whole target directory, which will include all build directories
with all binary versions.

Here's an example lein-jank target tree, with two binary versions for
the same debug profile.

```
target/
├── debug
│   ├── _cache
│   │   └── json-query
│   │       ├── x86_64-unknown-linux-gnu-211eea4be7942af843e902b894e93b6cc665b2aa6943f36f5bca4ddde080972b
│   │       │   └── json_query
│   │       │       └── main.o
│   │       └── x86_64-unknown-linux-gnu-5e3948b46dee3bb4f17fc699e821182481fb5465155bdbc359678f7e6a8db9da
│   │           └── json_query
│   │               └── main.o
│   └── json-query
└── stale
    └── leiningen.core.classpath.extract-native-dependencies
```

Here's an example of using jank to build a program directly, without
lein-jank.

```
.
├── hello.jank
└── target
    ├── a.out
    └── _cache
        └── x86_64-unknown-linux-gnu-deb515fb6fcde162c8cefee396e3090f348d5ec8e2d1a863abf325fa79682321
            └── hello.o
```

To aid in working with these binary versions, this PR also adds the following:

* The ability to override jank's binary version, with the `--force-binary-version` flag, for when having a stable path is important (we use this in CMake)
* The ability to query the binary version, with the `print-binary-version` command, for when having dynamic binary versions is desired but some tooling needs to know where to find things

Aside from a safety check for target/build dirs not being parents of any dirs on the module path, I believe this directory structure and implementation should last us a while, with lein-jank.

### DIFF
--- a/compiler+runtime/src/cpp/jank/util/environment_static.cpp
+++ b/compiler+runtime/src/cpp/jank/util/environment_static.cpp
@@ -11,6 +11,6 @@ namespace jank::util
 
   jtl::immutable_string build_dir()
   {
-    throw error::runtime_static_feature_disabled("eval");
+    throw error::runtime_static_feature_disabled("build_dir");
   }
 }
--- a/compiler+runtime/src/cpp/jank/util/environment_static.cpp
+++ b/compiler+runtime/src/cpp/jank/util/environment_static.cpp
@@ -9,7 +9,7 @@ namespace jank::util
     return res;
   }
 
-  jtl::immutable_string build_version()
+  jtl::immutable_string build_dir()
   {
     throw error::runtime_static_feature_disabled("eval");
   }
--- a/compiler+runtime/src/cpp/jank/util/environment_static.cpp
+++ b/compiler+runtime/src/cpp/jank/util/environment_static.cpp
@@ -1,4 +1,5 @@
 #include <jank/util/environment.hpp>
+#include <jank/error/runtime.hpp>
 
 namespace jank::util
 {
@@ -6,5 +7,10 @@ namespace jank::util
   {
     static jtl::immutable_string const res{ "static" };
     return res;
+  }
+
+  jtl::immutable_string build_version()
+  {
+    throw error::runtime_static_feature_disabled("eval");
   }
 }
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -180,34 +180,34 @@ set(
   "${CMAKE_SOURCE_DIR}/src/jank/jank/nrepl/server/handler/lookup.jank"
 )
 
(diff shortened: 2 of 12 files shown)


jeaye:
@kylc Would you mind taking a look at the PR description and providing input? I'm not concerned about the code, just making sure the design is good.

compiler+runtime/src/cpp/jank/util/environment_static.cpp
```diff
@@ -7,4 +8,9 @@ namespace jank::util
     static jtl::immutable_string const res{ "static" };
     return res;
   }
+
+  jtl::immutable_string build_dir()
+  {
+    throw error::runtime_static_feature_disabled("eval");
```
kylc: "build_dir"?

kylc (approved):
LGTM. This does mean that the target directory will continuously grow if you switch between different compiler versions/flags (especially since lein-jank has the same behavior). But Cargo does the same and they haven't found a good solution yet, so we are in good company.

kylc:
The long file paths caused by the encoded binary version like `x86_64-unknown-linux-gnu-5e3948b46dee3bb4f17fc699e821182481fb5465155bdbc359678f7e6a8db9da` might be a problem. I think Windows has a pretty short max path length. In `lein-jank` we use a path-safe base64 variant on a 128 bit hash to try to shorten up the hash string.

compiler+runtime/src/cpp/jank/util/environment_static.cpp
jeaye: I've been wondering how granular I want to get with these. I've been logically grouping some of them under `eval`, but maybe it's more useful to just have a per-fn name.

jeaye:
> LGTM. This does mean that the target directory will continuously grow if you switch between different compiler versions/flags (especially since lein-jank has the same behavior). But Cargo does the same and they haven't found a good solution yet, so we are in good company.

Yep, exactly. This is also how things like `~/.m2` work. Thanks for the review!

jeaye:
> The long file paths caused by the encoded binary version like `x86_64-unknown-linux-gnu-5e3948b46dee3bb4f17fc699e821182481fb5465155bdbc359678f7e6a8db9da` might be a problem. I think Windows has a pretty short max path length. In `lein-jank` we use a path-safe base64 variant on a 128 bit hash to try to shorten up the hash string.

Good thinking. I will make a note to see how we can improve this, ideally for both lein-jank and jank, in a consistent fashion. Not going to block the PR for it.

PR #867, opened 2026-07-01
Enable `clojure.core-test.atom`

djblue:


compiler+runtime/include/cpp/jank/runtime/obj/atom.hpp
```diff
@@ -55,5 +63,9 @@ namespace jank::runtime::obj
     /* Since watches is a `persistent_hash_map`, there in no guarantee in which
      * order watches are invoked. */
     folly::Synchronized<persistent_hash_map_ref> watches{};
+    folly::Synchronized<object_ref> validator{};
+
+  protected:
+    lazy_meta meta;
```
jeaye: Let's make this private, please.

compiler+runtime/include/cpp/jank/runtime/behavior/ref_like.hpp
```diff
@@ -4,6 +4,8 @@ namespace jank::runtime::behavior
 {
   template <typename T>
   concept ref_like = requires(T * const t) {
+    { t->set_validator(object_ref{}) } -> std::convertible_to<void>;
+    { t->get_validator() } -> std::convertible_to<object_ref>;
```
jeaye, line 8: This can work, since it's good to unblock this test and there's already a concept in use. However, the way forward is to utilize the open object model and add a new object behavior for `ref_like`. That way we can stop using `visit_object` and instead just use runtime behavior checks.

jeaye (changes requested):
Nice work, Chris!

compiler+runtime/include/cpp/jank/runtime/behavior/ref_like.hpp
djblue: Does this involve updating oref.h to include these methods with default implementations of unsupported?

jeaye: The default implementation belongs in the base type, which is `object`. The `oref` container just acts as a proxy for that and forwards the call to the object. For example: https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/runtime/object.cpp#L214

jeaye (approved):
Yeah! That's how we do it.

Later on, we can also remove the `ref_like` concept and any uses of it.

PR #869, opened 2026-07-01
Enable `clojure.core-test.sort`

djblue:


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -1171,6 +1176,39 @@ namespace jank::runtime
...
+  {
+    if(coll.is_nil())
+    {
+      return obj::persistent_list::empty();
+    }
+
+    return visit_seqable(
+      [=](auto const typed_coll) -> object_ref {
+        native_vector<object_ref> vec;
+        for(auto const &e : make_sequence_range(typed_coll))
```
jeaye: No need for `&` here, since `e` is an `object_ref`. Adding another layer of indirection may negatively impact performance. `object_ref` is asserted, at compile-time, to be 64 bits, so we know it can fit in a register and should just be copied around.

jeaye (changes requested):
Thank you, Chris! Small change worth pointing out.

jeaye (approved):
Thanks!

PR #870, opened 2026-07-01
Fix AOT compilation on Intel (x86-64) macOS

### LINKED ISSUE #866: AOT compilation crashes with "illegal text-relocations" on Intel (x86_64) macOS
## Environment
- macOS 26.5.1 (Build 25F80), Intel x86_64 (i7-9750H)
- jank `main` @ `fea82dbeef712b6c4ee46e9c28fdabfc4956f9bd`
- LLVM/Clang: Homebrew `llvm` 22.1.8 (keg-only, `/usr/local/opt/llvm`)
- Built per `compiler+runtime/doc/build.md` "Release" recipe, with
  `SDKROOT`, `PATH`, `LDFLAGS`, `CPPFLAGS` pointed at the Homebrew LLVM
  install (Intel Homebrew prefix is `/usr/local`, not `/opt/homebrew`)

## Summary
On Intel macOS, `jank`'s AOT compilation path fails at link time with
`ld: Found illegal text-relocations`, referencing symbols from
`__GLOBAL__sub_I_incr_module_*` (the incrementally JIT-compiled core
module init code). Forcing a non-PIE link (`-Wl,-no_pie`) avoids the
link error but the resulting binary crashes on startup (SIGBUS, before
any output) with no further diagnostics.

Regular JIT execution (`jank run`, interactive eval) works correctly.
Only the AOT path (`jank compile`, and `check-health`'s internal AOT
self-test) is affected.

## Steps to reproduce
1. Fresh clone + `git submodule update --init --recursive --jobs 8`.
2. `brew install entr double-conversion boost llvm` (curl/git/etc.
   already present via Xcode CLT).
3. ```bash
   export SDKROOT=$(xcrun --sdk macosx --show-sdk-path)
   export PATH="/usr/local/opt/llvm/bin:${PATH}"
   export LDFLAGS="-L/usr/local/opt/llvm/lib"
   export CPPFLAGS="-I/usr/local/opt/llvm/include"
   export CC=clang CXX=clang++
   cd compiler+runtime
   ./bin/configure -GNinja -DCMAKE_BUILD_TYPE=Release
   ./bin/compile
   ```
4. ```bash
   export JANK_EXTRA_FLAGS="-L/usr/local/lib"   # see note below
   ./build/jank check-health
   ```

## Actual result
```
  text-relocation in '__GLOBAL__sub_I_incr_module_4'+0x5A (.../health.o) to '___cxx_global_var_init.7'
  ... (several more, same pattern)
ld: Found illegal text-relocations
error: Clang failed with errors.error:  linker command failed with exit code 1
─ ❌ jank cannot aot compile working binaries
```

Forcing `JANK_EXTRA_FLAGS="-L/usr/local/lib -Wl,-no_pie"` makes the
link succeed, but running the produced binary directly (also
reproduced with a minimal standalone `-main` program compiled via
`jank compile`) crashes immediately:
```
$ ./target/a.out
[no output]
$ echo $?
138   # SIGBUS
```

## What was ruled out
- **Not the Homebrew prefix mismatch** (separate, smaller bug — see
  below): reproduces identically even after supplying the correct
  `-L` for Intel Homebrew.
- **Not a missing `-fPIC` on a third-party static lib**: reconfigured
  with `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` globally and did a full
  rebuild. Identical failure, same symbol pattern
  (`__GLOBAL__sub_I_incr_module_*`). This points at the incremental
  JIT/CppInterOp compilation pipeline itself emitting non-PIC-relocatable
  object code (fine for in-process JIT execution, since it doesn't need
  ASLR-safe relocations there) that then gets reused/relinked into the
  final AOT executable, which macOS's linker rejects for a PIE target.

## Secondary bug found along the way
`src/cpp/jank/aot/processor.cpp` unconditionally hardcodes
```cpp
if constexpr (jtl::current_platform == jtl::platform::macos_like)
{
  compiler_args.push_back(strdup("-L/opt/homebrew/lib"));
}
```
This assumes the Apple Silicon Homebrew prefix. On Intel macOS,
Homebrew lives at `/usr/local`, so this never resolves `-lzstd`
(and would presumably miss any other Homebrew-provided AOT link
dependency) unless the user manually passes
`JANK_EXTRA_FLAGS="-L/usr/local/lib"`. Should probably resolve the
Homebrew prefix dynamically (e.g. via `brew --prefix` at configure
time, or checking both `/opt/homebrew/lib` and `/usr/local/lib`).

## Related issues
Searched open + closed issues for "text-relocation", "illegal
text-relocation", "SIGBUS", "relocation model", "incr_module",
"non-PIC", "no_pie", "PIE macOS", "AOT macOS/Intel/x86_64" before
filing. No existing issue reports this exact failure. Two issues are
relevant context, not duplicates:

- **#645 "Homebrew package is only for aarch64, not x86"** (open,
  `help wanted`). Different symptom (the published Homebrew bottle is
  arm64-only, so `jank` won't even run — `bad CPU type in
  executable`), but the maintainer's reply confirms the underlying
  gap: *"we only have aarch64 binaries right now... I'm not familiar
  with how to get both added... but we definitely want to do that."*
  Independent confirmation that Intel macOS isn't part of jank's
  tested/shipped matrix today.
- **#515 "Fix AOT builds on Nix"** (closed, apparently without landing
  a fix on `main`). Different root cause (the `clang::Driver` CLI-arg
  translation and `JANK_DEPS_LIBRARY_DIRS` misbehave under Nix's
  compiler wrapper), but the same *shape* of problem: AOT compilation
  breaking on a platform/toolchain combination outside the primary
  macOS-arm64 / Linux-x86_64 CI matrix. Read together with #645, this
  suggests AOT-on-non-primary-platform is a recurring soft spot.

Also checked and ruled out as unrelated: #504 (Bad `CppInterOp`
codegen — a `cpp/raw` template/JIT-eval crash, not AOT/linking), #582
(ODR violation for duplicate `cpp/raw` definitions during AOT), #632
(AOT not passing user `-l` flags to clang, fixed by PR #631), #379
(dylib rpath resolution for external libs, closed as not-a-jank-bug).

## Question
Is x86_64 macOS an actively tested/supported target for AOT
compilation? If so, any pointers on where in the incremental
compilation pipeline (CppInterOp / Cling-derived) the relocation
model would need to be forced to PIC for AOT-reused object code?
Happy to test patches — this reproduces reliably on a real Intel Mac.

## Context
Found this while trying to build `jank-engine`, a small jank+OpenGL
game engine (private repo), which is itself produced via this exact
AOT pipeline. It currently cannot be built on Intel macOS with a
locally-built jank; JIT-only usage (`jank run`, `jank repl`) is
unaffected.

burinc:
Fixes #866.

## Problem

On Intel (x86-64) macOS, `jank`'s AOT compilation path fails at link
time with `ld: Found illegal text-relocations`, and forcing a non-PIE
link instead crashes at startup (SIGBUS). Every other platform we
support (arm64 Darwin, Linux) is unaffected. Regular JIT execution
(`jank run`, interactive eval) also works correctly — only the AOT
path (`jank compile`, and `check-health`'s AOT self-test) is broken.

## Root cause

`CodeModel::Large` combined with `Reloc::PIC_` produces object code
with absolute-address text relocations on x86-64 Darwin. The linker
rejects that when the code is reused during AOT linking of a
position-independent executable. This affects two call sites that
both construct target machines for jank-compiled code:

- `jit/processor_dynamic.cpp` — the JIT interpreter's own target
  machine (`Cpp::CreateInterpreter`'s `CodeModel` argument).
- `runtime/context_dynamic.cpp` — the AOT `compile-module
  --output-target object` codegen path.

I also found and fixed a smaller, related issue while diagnosing
this: `aot/processor.cpp` unconditionally passes `-L/opt/homebrew/lib`
during AOT linking, which is only the Apple Silicon Homebrew prefix.
On Intel macOS, Homebrew installs to `/usr/local`, so
Homebrew-provided AOT link dependencies (e.g. `-lzstd`) were never
found without manually passing an extra `-L` flag.

## Fix

Use `CodeModel::Small` instead of `Large`, scoped narrowly to
x86-64+macOS (`#if defined(__x86_64__)` combined with
`jtl::current_platform == jtl::platform::macos_like`), leaving every
other platform untouched. I deliberately didn't change the code model
globally — `Large` may genuinely be needed elsewhere for bigger
binaries, and this bug only reproduces on this one platform
combination.

Also add `-L/usr/local/lib` alongside the existing
`-L/opt/homebrew/lib` in the AOT linker args; a nonexistent `-L`
directory is a harmless no-op for the linker.

## Verification

- `./build/jank check-health` — all green including
  `jank can aot compile working binaries`, using **stock** Homebrew
  LLVM 22.1.8 (no custom LLVM build required).
- Minimal repro: `jank compile` on a trivial `-main` namespace, then
  running the resulting binary — prints correctly instead of crashing.
- Full real-world test: built and ran a jank+OpenGL game engine
  project (private repo, uses ENet networking, ozz animation, etc.)
  end-to-end on this Intel Mac for the first time — client/server
  connected and exchanged messages correctly over real UDP
  networking.

Environment: macOS 26.5.1, Intel x86-64 (i7-9750H), Homebrew LLVM
22.1.8.

### DIFF
--- a/compiler+runtime/src/cpp/jank/aot/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/aot/processor.cpp
@@ -224,7 +224,11 @@ int main(int argc, const char** argv)
 
     if constexpr(jtl::current_platform == jtl::platform::macos_like)
     {
+      /* Homebrew's prefix differs by architecture: /opt/homebrew on Apple
+       * Silicon, /usr/local on Intel. Add both; an absent -L path is a
+       * harmless no-op for the linker. */
       compiler_args.push_back(strdup("-L/opt/homebrew/lib"));
+      compiler_args.push_back(strdup("-L/usr/local/lib"));
     }
 
     for(auto const &library_dir : util::cli::opts.library_dirs)
--- a/compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
@@ -221,9 +221,24 @@ namespace jank::jit
 
     //util::println("jit flags {}", args);
 
+    /* CodeModel::Large combined with Reloc::PIC_ produces object code with
+     * absolute-address text relocations on x86-64 Darwin, which the linker
+     * rejects ("illegal text-relocations") when that code is reused during
+     * AOT linking of a position-independent executable. Every other target
+     * we support handles Large+PIC_ fine, so we only narrow the code model
+     * here. */
+    constexpr auto code_model{
+#if defined(__x86_64__)
+      jtl::current_platform == jtl::platform::macos_like ? llvm::CodeModel::Small
+                                                          : llvm::CodeModel::Large
+#else
+      llvm::CodeModel::Large
+#endif
+    };
+
     /* We don't actually own this interpreter. CppInterOp does. */
     interpreter.reset(static_cast<CppInternal::Interpreter *>(
-      Cpp::CreateInterpreter(args, {}, vfs, static_cast<int>(llvm::CodeModel::Large))));
+      Cpp::CreateInterpreter(args, {}, vfs, static_cast<int>(code_model))));
 
     /* Enabling perf support requires registering a couple of plugins with LLVM. These
      * plugins will generate files which perf can then use to inject additional info
--- a/compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
@@ -230,7 +230,7 @@ namespace jank::jit
     constexpr auto code_model{
 #if defined(__x86_64__)
       jtl::current_platform == jtl::platform::macos_like ? llvm::CodeModel::Small
-                                                          : llvm::CodeModel::Large
+                                                         : llvm::CodeModel::Large
 #else
       llvm::CodeModel::Large
 #endif
--- a/compiler+runtime/src/cpp/jank/runtime/context_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/context_dynamic.cpp
@@ -306,7 +306,7 @@ namespace jank::runtime
           constexpr auto code_model{
 #if defined(__x86_64__)
             jtl::current_platform == jtl::platform::macos_like ? llvm::CodeModel::Small
-                                                                : llvm::CodeModel::Large
+                                                               : llvm::CodeModel::Large
 #else
             llvm::CodeModel::Large
 #endif
--- a/compiler+runtime/src/cpp/jank/runtime/context_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/context_dynamic.cpp
@@ -289,15 +289,28 @@ namespace jank::runtime
           //module->print(llvm::outs(), nullptr);
 
           auto const target_triple{ util::default_target_triple() };
+          llvm::Triple const triple{ target_triple.c_str() };
           std::string target_error;
-          auto const target{
-            llvm::TargetRegistry::lookupTarget(llvm::Triple{ target_triple.c_str() }, target_error)
-          };
+          auto const target{ llvm::TargetRegistry::lookupTarget(triple, target_error) };
           if(!target)
           {
             return err(target_error);
           }
           llvm::TargetOptions const opt;
+          /* CodeModel::Large combined with Reloc::PIC_ produces object code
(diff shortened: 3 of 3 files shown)


jeaye:
Hey Burin. Apologies for the long delay. I need to verify this fix with an LLVM dev, since I don't understand it. Using a small code model should result in other issues, with invalid 32bit locations, which is exactly why we're using a large code model. We didn't start by using a large code model; I had to patch LLVM to make that possible. So reverting this for macOS x86 is concerning.

PR #872, opened 2026-07-03
Enable `clojure.core-test.juxt`

djblue:


jeaye (approved):
Nice! More code string printing biting us in the butt. We've learned our lesson, but we're still paying the price. 😂

PR #874, opened 2026-07-03
Enable `clojure.core-test.subvec`

djblue:


jeaye (approved):
Ah, that's a good catch.

PR #876, opened 2026-07-03
Enable `clojure.core-test.shuffle`

djblue:


compiler+runtime/src/cpp/jank/runtime/core/seq.cpp
```diff
@@ -1227,6 +1227,19 @@ namespace jank::runtime
   {
     return visit_seqable(
       [](auto const typed_coll) -> object_ref {
+        using T = typename jtl::decay_t<decltype(typed_coll)>::value_type;
+
+        if constexpr(!behavior::collection_like<T>
+                     || jtl::is_any_same<T,
+                                         obj::persistent_hash_map,
+                                         obj::persistent_array_map,
+                                         obj ::persistent_sorted_map>)
```
jeaye: I think we want `map_like` for this.

jeaye (changes requested):

jeaye (approved):
Nice!

PR #878, opened 2026-07-03
Enable `clojure.core-test.eval`

djblue:


compiler+runtime/src/cpp/jank/analyze/processor.cpp
```diff
@@ -4810,13 +4797,11 @@ namespace jank::analyze
...
+        // https://clojure.org/reference/evaluation
+        // > Any object other than those discussed above will evaluate to itself.
         else
         {
-          return error::internal_analyze_failure(
-            util::format("Unimplemented analysis for object type '{}'.",
-                         object_type_str(typed_o.get_type())),
-            object_source(o),
-            latest_expansion(macro_expansions));
+          return analyze_primitive_literal(o, current_frame, position, fn_ctx, needs_box);
```
djblue, line 4804: So this is to fix `(eval (fn [x] x))`, but a fn is technically not a primitive, I think? However, it behaves the same as one in the context of eval (self evaluating).

jeaye: Hm, if the tests pass, this works for me.

jeaye (approved):
Thanks!

PR #882, opened 2026-07-05
Enable `clojure.string-test.reverse`

djblue:
Before:
```clojure
% jank repl
nil
user=> (str/reverse "hello")
"olleh"
user=> (str/reverse "🫲  hello 🫱")
"���� olleh  ����"
```

After:
```clojure
% jank repl              
nil
user=> (str/reverse "hello")
"olleh"
user=> (str/reverse "🫲  hello 🫱")
"🫱 olleh 🫲"
```

### DIFF
--- a/compiler+runtime/include/cpp/jtl/utf8.hpp
+++ b/compiler+runtime/include/cpp/jtl/utf8.hpp
@@ -1,29 +1,30 @@
-#include <ranges>
-
 #include <jtl/immutable_string.hpp>
 
 namespace jtl
 {
-  struct reverse_iterator
+  struct utf8_iterator
   {
-    using reverse_iterator_category = std::input_iterator_tag;
+    using iterator_category = std::bidirectional_iterator_tag;
     using difference_type = std::ptrdiff_t;
     using value_type = immutable_string;
     using reference = value_type const &;
 
-    reverse_iterator() = default;
-    reverse_iterator(immutable_string const &s);
-    reverse_iterator(immutable_string const &s, value_type::size_type const i);
+    utf8_iterator() = default;
+    utf8_iterator(immutable_string const &s);
+    utf8_iterator(immutable_string const &s, value_type::size_type const i);
 
     value_type operator*() const;
-    reverse_iterator &operator++();
-    reverse_iterator operator++(int);
-    bool operator==(reverse_iterator const &it) const;
+    utf8_iterator &operator++();
+    utf8_iterator operator++(int);
+    utf8_iterator &operator--();
+    utf8_iterator operator--(int);
+    bool operator==(utf8_iterator const &it) const;
+
+    utf8_iterator begin();
+    utf8_iterator end();
 
     value_type data;
     value_type::size_type i{};
     value_type::size_type n{};
   };
-
-  std::ranges::subrange<reverse_iterator> utf8_reverse_range(immutable_string const &s);
 }
--- a/compiler+runtime/include/cpp/jtl/utf8.hpp
+++ b/compiler+runtime/include/cpp/jtl/utf8.hpp
@@ -1,110 +1,29 @@
+#include <ranges>
+
 #include <jtl/immutable_string.hpp>
 
 namespace jtl
 {
-  constexpr bool
-  is_continuation_byte(immutable_string const &s, immutable_string::size_type const i)
+  struct reverse_iterator
   {
-    if(i == immutable_string::npos)
-    {
-      throw std::runtime_error{ "Invalid unicode string." };
-    }
+    using reverse_iterator_category = std::input_iterator_tag;
+    using difference_type = std::ptrdiff_t;
+    using value_type = immutable_string;
+    using reference = value_type const &;
 
-    return (s[i] & 0xC0) == 0x80;
-  }
+    reverse_iterator() = default;
+    reverse_iterator(immutable_string const &s);
+    reverse_iterator(immutable_string const &s, value_type::size_type const i);
 
-  struct utf8_reverse_range
-  {
-    utf8_reverse_range(immutable_string const &s)
-      : data{ s }
-    {
-    }
+    value_type operator*() const;
+    reverse_iterator &operator++();
+    reverse_iterator operator++(int);
+    bool operator==(reverse_iterator const &it) const;
 
-    struct iterator
-    {
-      using iterator_category = std::input_iterator_tag;
-      using difference_type = std::ptrdiff_t;
-      using value_type = immutable_string;
-      using reference = value_type const &;
-
-      iterator()
-        : data{}
-        , i{}
-        , n{}
-      {
-      }
-
-      iterator(immutable_string const &s)
-        : data{ s }
-        , i{ s.size() }
-        , n{ prev_char_size() }
-      {
-      }
-
-      constexpr value_type::size_type prev_char_size() const
-      {
-        if(!is_continuation_byte(data, i - 1))
-        {
-          return 1;
-        }
-        else if(!is_continuation_byte(data, i - 2))
-        {
-          return 2;
-        }
-        else if(!is_continuation_byte(data, i - 3))
-        {
-          return 3;
-        }
-        else if(!is_continuation_byte(data, i - 4))
-        {
-          return 4;
-        }
-        else
-        {
-          throw std::runtime_error{ "Invalid unicode string." };
-        }
-      }
-
-      value_type operator*() const
-      {
-        return data.substr(i - n, n);
-      }
-
-      iterator &operator++()
-      {
-        i -= n;
-        if(i > n)
-        {
-          n = prev_char_size();
-        }
-        return *this;
-      }
-
-      bool operator==(iterator const &it) const
-      {
-        return i == it.i;
-      }
-
-      value_type data;
-      value_type::size_type i;
-      value_type::size_type n;
-    };
-
-    iterator begin() const
-    {
-      if(data.empty())
-      {
-        return {};
-      }
-
-      return { data };
-    }
-
-    iterator end() const
-    {
-      return {};
-    }
-
-    immutable_string data;
+    value_type data;
+    value_type::size_type i{};
+    value_type::size_type n{};
   };
+
+  std::ranges::subrange<reverse_iterator> utf8_reverse_range(immutable_string const &s);
 }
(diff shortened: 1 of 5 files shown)


compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -24,9 +24,48 @@ namespace clojure::string_native
...
+
+    return (s[i] & 0xC0) == 0x80;
+  }
+
   jtl::immutable_string reverse(jtl::immutable_string const &s)
   {
-    return { s.rbegin(), s.rend() };
+    usize i{ s.size() };
+    jtl::string_builder buff{ s.size() };
+    while(i > 0)
```
djblue: Not sure if we want this code here or in `jtl::immutable_string`.

I do think this could potentially be how we implement the reverse code point iterator for `immutable_string`, and do the opposite for the forward iterator (would be useful for `persistent_string_sequence`). But I wasn't sure how/where to add the code point iterator. Does it live in `immutable_string.hpp`, is it a separate struct or is it in another jtl file entirely (`<jtl/utf8.hpp>`)?

jeaye: We could put it into `immutable_string_view`, since it will be used by `immutable_string` and `immutable_string_view`. I would nest two structs within `immutable_string_view`.

* `utf8_iterator`
* `utf8_reverse_iterator`

Something like this ought to get you going:

```cpp
struct utf8_iterator
{
    using difference_type = std::ptrdiff_t;
    using value_type = immutable_string_view;

    immutable_string_view operator*() const; // TODO
    utf8_iterator& operator++(); // TODO
    utf8_iterator operator++(int)
    {
        auto tmp{ *this };
        ++*this;
        return tmp;
    }

    bool operator==(const utf8_iterator&) const; // TODO
};
```

djblue: I think I finally realized what I actually wanted was to be able to do the following:

```cpp
jtl::string_builder buff{ s.size() };
for(auto const &c : jtl::utf8_reverse_range(s))
{
  buff(c);
}
return buff.release();
```

After taking a look at [`sequence_range`](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/include/cpp/jank/runtime/sequence_range.hpp), I added [`utf8_reverse_range`](https://github.com/jank-lang/jank/pull/882/changes#diff-bafaff7a01c2bb7ea62f955bc1065e87dba0a005c5e6d9b90e179124f15da38e). It's probably not in the correct place, but it was easier to iterate in a new file. Is it a reasonable way to solve this problem? Are there issues with this type of solution to be aware of?

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -26,7 +28,12 @@ namespace clojure::string_native
 
   jtl::immutable_string reverse(jtl::immutable_string const &s)
   {
-    return { s.rbegin(), s.rend() };
```
djblue, line 29: It would be kind of cool if I could do the following:
```cpp
jtl::utf8_range const r{ s };
return { r.rbegin(), r.rend() };
```

jeaye: This is absolutely something we could do.

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,110 @@
+#include <jtl/immutable_string.hpp>
+
+namespace jtl
```
jeaye, line 3: We don't want all of these fn definitions to be in the header. That's going to slow down compilation and bloat our PCH further, which will then slow down jank's startup time. Please make a corresponding `src/cpp/jtl/utf8.cpp` for the implementation.

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      bool operator==(iterator const &it) const
+      {
+        return i == it.i;
+      }
+
+      value_type data;
+      value_type::size_type i;
+      value_type::size_type n;
```
jeaye: Please always default initialize primitives.

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,110 @@
...
+      using difference_type = std::ptrdiff_t;
+      using value_type = immutable_string;
+      using reference = value_type const &;
+
+      iterator()
+        : data{}
+        , i{}
+        , n{}
+      {
+      }
```
jeaye: There is no need for this, if you default initialize them where they're defined. Instead you can just do `iterator() = default;`.

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,110 @@
+#include <jtl/immutable_string.hpp>
+
+namespace jtl
+{
+  constexpr bool
+  is_continuation_byte(immutable_string const &s, immutable_string::size_type const i)
+  {
+    if(i == immutable_string::npos)
+    {
+      throw std::runtime_error{ "Invalid unicode string." };
```
jeaye: Technically UTF-8.

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      iterator(immutable_string const &s)
+        : data{ s }
+        , i{ s.size() }
+        , n{ prev_char_size() }
+      {
+      }
+
+      constexpr value_type::size_type prev_char_size() const
```
jeaye: This is not `prev`, it's `next`, since we're going in reverse.

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,110 @@
...
+        if(i > n)
+        {
+          n = prev_char_size();
+        }
+        return *this;
+      }
+
+      bool operator==(iterator const &it) const
+      {
+        return i == it.i;
```
jeaye: This will mean to iterators pointing at different strings, at the same index, will be considered equal. We need to include more. We can also compare `data.data()`, so we check the pointer of our string, rather than comparing the whole string. You'll need to update your `begin` and `end` to pass in `data`, though.

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+    return (s[i] & 0xC0) == 0x80;
+  }
+
+  struct utf8_reverse_range
+  {
+    utf8_reverse_range(immutable_string const &s)
+      : data{ s }
+    {
+    }
```
jeaye: We don't really need to make the range ourselves. The C++ stdlib helps us with that. If we just have an iterator pair (begin/end), we can use `std::ranges::subrange` to get a range.

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,110 @@
...
+        return data.substr(i - n, n);
+      }
+
+      iterator &operator++()
+      {
+        i -= n;
+        if(i > n)
+        {
+          n = prev_char_size();
+        }
```
jeaye: This feels incorrect to me, but let me check with you. Let's assume we have a 4 byte string, with three characters.

```
[ 1 byte ] [ 1 byte ] [ 2 bytes ]
```

In reverse, we'll start with `i = 4` and `n = 2`. When we increment, we'll drop to `i = 2`, but `i > n` is false, so we will keep `n = 2` and our next increment will skip past the middle character.

jeaye (changes requested):
Thank you, Chris!

compiler+runtime/include/cpp/jtl/utf8.hpp
djblue: Should I add some tests for this?

jeaye: Seems like an excellent sort of test case for the clojure-test-suite. To help a brother out:

* One byte: `a`
* Two bytes: `¡`
* Three bytes: `ষ`
* Four bytes: `𐅦`

compiler+runtime/include/cpp/jtl/utf8.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+    value_type operator*() const;
+    utf8_iterator &operator++();
+    utf8_iterator operator++(int);
+    utf8_iterator &operator--();
+    utf8_iterator operator--(int);
+    bool operator==(utf8_iterator const &it) const;
+
+    utf8_iterator begin() const;
+    utf8_iterator end() const;
```
jeaye: Iterators do not have `begin`/`end`. Ranges do. These two things should not be conflated.

jeaye: Below, when you use this with a reverse view, what we actually want is `utf8_range{ s } | ...`. So the fix here is to add a new `struct utf8_range` which holds a string and provides `begin`/`end` (and optionally `rbegin`/`rend`).

jeaye (changes requested):
Thanks, Chris!

compiler+runtime/include/cpp/jtl/utf8.hpp
djblue: Sorry for all the code churn. I'm still learning about iterators + range based for loops. I think I finally landed in a good place. Initially we talked about having two separate iterators forwards/reverse, but it felt better if we could have one iterator that could go forwards and backwards.

compiler+runtime/include/cpp/jtl/utf8.hpp
djblue: I added more examples in https://github.com/jank-lang/clojure-test-suite/pull/936. I can bump the clojure-test-suite once they get it.

jeaye (dismissed):
Nice!

jeaye:
Merged https://github.com/jank-lang/clojure-test-suite/pull/936. Want to pull that in here?

djblue:
@jeaye when I bump to latest clojure-test-suite, existing enabled tests break. Should we address bumping clojure-test-suite in another PR, or should I disable the failing tests?

jeaye:
> @jeaye when I bump to latest clojure-test-suite, existing enabled tests break. Should we address bumping clojure-test-suite in another PR, or should I disable the failing tests?

Let's disable the failing tests.

jeaye (approved):
Nice!

PR #846, opened 2026-06-26
Add initial Arch CI jobs

jeaye:
As part of this PR, we:

* Add an Arch CI job, which runs on main, to publish tarballs for Arch packaging
* Add (unreleased) LLVM 23 support and bump CI to use LLVM 23 by default (except on Windows)
* Add support for ADL calls for inline friend fns, which became necessary with recent libc++ changes
* Fix double registration of GC threads (this closes #708 and #722)

PR #884, opened 2026-07-05
build-clang(win): make MSYS2 patching more resileint when upgrading LLV version

ikappaki:
Hi,

Could you please consider this patch, which introduces a more resilient way of applying the MSYS2 LLVM package patches to jank's pinned LLVM repository checkout? It addresses the Windows build failures reported in #846 after upgrading to LLVM 23.

The approach treats the MSYS2 LLVM package `PKGBUILD` as a template containing placeholder values. The `bin/build-clang` script replaces these placeholders with the version of the pinned LLVM checkout read directly from the cmake version file, allowing the patches to work as intended taken the correct version into account. This change depends on https://github.com/jank-lang/MINGW-packages/pull/2, which adds support for the placeholder values.

I've also added documentation at the top of the script to make it easier for future contributors to understand how it works and update it when needed.

Comments were generated in collaboration with AI.

Thanks


jeaye:
> Could you please consider this patch, which introduces a more resilient way of applying the MSYS2 LLVM package patches to jank's pinned LLVM repository checkout? It addresses the Windows build failures reported in #846 after upgrading to LLVM 23.

Thanks for the patch. At this build `bin/build-clang` is entirely different for Windows. What do you think about separating it into `bin/build-clang-msys2`? That way msys2 changes don't end up busting the cache for all of our non-Windows CI jobs, too. Already, this is going to cost us several hours of CI compilation because `bin/build-clang` has changed, but the actual artifact for macOS, Ubuntu, Arch will be the same.
 
> The approach treats the MSYS2 LLVM package `PKGBUILD` as a template containing placeholder values. The `bin/build-clang` script replaces these placeholders with the version of the pinned LLVM checkout read directly from the cmake version file, allowing the patches to work as intended taken the correct version into account. This change depends on [jank-lang/MINGW-packages#2](https://github.com/jank-lang/MINGW-packages/pull/2), which adds support for the placeholder values.

Nice, this should address the issue with us wanting to use unreleased versions.

> I've also added documentation at the top of the script to make it easier for future contributors to understand how it works and update it when needed.
> 
> Comments were generated in collaboration with AI.

Please remove these comments. jank has a policy against AI-generated content, including comments. It is covered by our contributing guide and the CLA you've signed.

ikappaki:
> Thanks for the patch. At this build `bin/build-clang` is entirely different for Windows. What do you think about separating it into `bin/build-clang-msys2`? That way msys2 changes don't end up busting the cache for all of our non-Windows CI jobs, too. Already, this is going to cost us several hours of CI compilation because `bin/build-clang` has changed, but the actual artifact for macOS, Ubuntu, Arch will be the same.

I've split it into `bin/build-clang-msys2`. Although there is a fair amount of duplication, it should be straightforward to keep it in sync , as long as we remember to reflect changes from the main script.

> > I've also added documentation at the top of the script to make it easier for future contributors to understand how it works and update it when needed.
> > Comments were generated in collaboration with AI.
> 
> Please remove these comments. jank has a policy against AI-generated content, including comments. It is covered by our contributing guide and the CLA you've signed.

It was just the top level comment in the MSYS2 script which I since rewritten by hand. 

I've reread the AI policy, and the rule itself is clear, I was pushing my luck on this. What I'm missing is the rationale behind it. Is the primary concern copyright, reviewability, or something else? It might be worth briefly explaining the reasoning in the policy, as I think that would help contributors better understand the motivation behind it. Sorry to bother you with this. Thanks

ikappaki:
Hi, just to confirm my understanding of the expected process for this PR, as I may not have been clear:
1. The CI workflow runs successfully, confirming that the `mingw-w64-package` patch inhttps://github.com/jank-lang/MINGW-packages/pull/2 works as expected. We then merge that PR into `master`.
2. I update the commit hash in `bin/build-clang-msys2` to point to the new `master` commit.
3. We then merge this PR with the updated hash.

I wasn't expecting the CI workflow on this PR to require approval before running, but I assume that's maybe part of the effort to reduce CI costs?

Thanks!

compiler+runtime/doc/build.md
```diff
@@ -80,12 +80,20 @@ Now configure and build jank as normal, but pass `-Djank_local_clang=on` when yo
 
 For Windows:
 
-Run the same commands from the [MSYS2](https://www.msys2.org/) CLANG64 shell.
+Run the same commands from the [MSYS2](https://www.msys2.org/) CLANG64 shell, but use the **bin/build-clang-msys2** script instead.
```
jeaye: Please use back ticks here (`) instead of asterisks.

compiler+runtime/bin/build-clang-msys2
```diff
@@ -0,0 +1,189 @@
...
+# --- Configuration, edit as needed.
+
+# LLVM repo to clone.
+llvm_url="https://github.com/jank-lang/llvm-project.git"
+llvm_branch="jank"
+llvm_commit="8164f1a0c17b192e133817436bdb07598b7402a3"
+
+# Patched version of the MSYS2-package mingw-w64-llvm used to build
+# LLVM from source. Tracks the MSYS2-packages mingw-w64-llvm package.
+llvm_mingw_pkg_url="https://github.com/ikappaki/MINGW-packages"
```
jeaye: This will need to change to be the jank-lang repo.

jeaye (changes requested):
> 1. The CI workflow runs successfully, confirming that the `mingw-w64-package` patch inhttps://github.com/[mingw-w64-llvm: update package to latest, use placeholder for better resilience when updating MINGW-packages#2](https://github.com/jank-lang/MINGW-packages/pull/2) works as expected. We then merge that PR into `master`.
> 
>     2. I update the commit hash in `bin/build-clang-msys2` to point to the new `master` commit.
> 
>     3. We then merge this PR with the updated hash.

Yeah, this process sounds right. The MSYS PR currently has conflicts which need to be resolved before we can merge. Then we can focus on this PR.

> I wasn't expecting the CI workflow on this PR to require approval before running, but I assume that's maybe part of the effort to reduce CI costs?

This is a security measure, since jank CI uses various secrets. If we run arbitrary CI jobs from outside contributors without review, we're risk exposing those secrets.

compiler+runtime/doc/build.md
ikappaki: Updated

compiler+runtime/bin/build-clang-msys2
ikappaki: Updated to point to master latest.

ikappaki:
Hi @jeaye, I've updated the build script to point to the master MINGW-packages commit , merged from main, and updated to point to the same llvm23 commit as the main script. Can you please have another look.

thanks

ikappaki:
Hi @jeaye, I've updated the `bin/jank/util.clj llvm-version` to 23 for all archs since the windows build failed looking for llvm v22, i wasn't aware about this conditional, could you please restart the CI or suggest otherwise, thanks. 

(Hopefully, llvm packages are now cached)

jeaye (approved):
Thank you!

PR #886, opened 2026-07-07
Add rerun-if[-env]-changed to jank-build

kylc:
Eliminates unnecessary reruns of top-level jank-builds when the source does not change. Conversely, prevents stale builds when some external env var changes (like $CC or $TARGET).

See: https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection

### DIFF
--- a/lein-jank/src/leiningen/jank/changed.clj
+++ b/lein-jank/src/leiningen/jank/changed.clj
@@ -1,0 +1,14 @@
+(ns leiningen.jank.changed
+  (:require [babashka.fs :as fs]))
+
+(defn getenv [k]
+  (System/getenv k))
+
+(defn maximum-mtime
+  "Read the mtime of the given file, or the recursive maximum if given a
+  directory."
+  [path]
+  (if (fs/directory? path)
+    (apply max (map maximum-mtime (fs/list-dir path)))
+    (fs/file-time->millis (fs/last-modified-time path))))
+
--- a/lein-jank/src/leiningen/jank/changed.clj
+++ b/lein-jank/src/leiningen/jank/changed.clj
@@ -1,14 +1,34 @@
 (ns leiningen.jank.changed
-  (:require [babashka.fs :as fs]))
+  "For tracking changes in the environment and in build source files."
+  (:require [babashka.fs :as fs]
+            [leiningen.jank.fingerprint :refer [fingerprint]]))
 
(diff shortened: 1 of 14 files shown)


lein-jank/src/leiningen/jank/changed.clj
```diff
@@ -0,0 +1,14 @@
...
+
+(defn getenv [k]
+  (System/getenv k))
+
+(defn maximum-mtime
+  "Read the mtime of the given file, or the recursive maximum if given a
+  directory."
+  [path]
+  (if (fs/directory? path)
+    (apply max (map maximum-mtime (fs/list-dir path)))
```
jeaye: If this follows symlinks, it might loop forever. If not, might be worth a comment saying so.

lein-jank/test-data/test-change-detection/project.clj
```diff
@@ -0,0 +1,7 @@
+(defproject org.jank-lang/test-change-detection "0.1.0"
+  :description "FIXME: write description"
+  :url "https://example.com/FIXME"
+  :license {:name "MPL 2.0"
+            :url  "https://www.mozilla.org/en-US/MPL/2.0/"}
+  :plugins [[org.jank-lang/lein-jank "0.7"]]
```
jeaye: Should we install a local version of `lein-jank` for test running so that we don't need to always keep these up to date? The most recent version is `2026.06-2`, but it'll be bumped again based on your recent PRs.

jeaye: I suppose, based on how we're currently testing, it won't matter.

kylc: I don't think it matters, but I've updated these to say "LATEST" to avoid any confusion.

lein-jank/src/leiningen/jank/changed.clj
kylc: Added a note about symlinks not being followed, but their self mtime being respected.

jeaye:
@kylc would you mind pulling in main to fix the conflicts? If there's anything else you'd like to do for this PR, please let me know. Otherwise, we can push it on through. :)

kylc:
Ready to merge :+1:

jeaye (approved):
Thank you!

PR #887, opened 2026-07-07
Add options for `lein compile-module` + cleanup

kylc:


lein-jank/src/leiningen/jank.clj
```diff
@@ -34,25 +34,43 @@ Calls the -main function in the given namespace."
...
   (let [[opts args] (ljc/parse-opts #'compile! args ljc/standard-options)]
     (if-let [main (:main project)]
       (dispatch-jank project opts "compile" [main] args)
       (lmain/warn "No :main entrypoint for project."))))
 
+(def compile-module-cli-options
+  [["-m" "--module NAMESPACE" "the jank namespace to compile"
+    :missing "Must provide namespace to compile"]
+   ["-o" "--output FILE" "the output file name (*.cpp, *.o)"
+    :missing "Must provide output file name"]
```
jeaye: Hm, jank doesn't actually support `-o` for module or for executables now. We just have `--name` for executables. I think we probably want to remove this flag.

lein-jank/src/leiningen/jank.clj
```diff
@@ -34,25 +34,43 @@ Calls the -main function in the given namespace."
...
+:main is specified."
   [project & args]
   (let [[opts args] (ljc/parse-opts #'repl! args ljc/standard-options)
         main        (:main project)]
     (dispatch-jank project opts "repl" (if main [main] []) args)))
 
 (defn compile!
-  "Compile your project to an executable."
+  "Ahead of time compile your project to an executable, starting at the :main
+entrypoint."
```
jeaye: We could consider adding `--name` here, in case people want to override it for one invocation. I don't think it's a big deal either way.

lein-jank/src/leiningen/jank.clj
kylc: `-o` here becomes `--name` on L69 before passing to jank. We can just call it `-n/--name` to match the jank flag though.

lein-jank/src/leiningen/jank.clj
kylc: BTW there is some odd behavior which I think is inherited from `jank` where `jank compile` puts the executable in the target dir, but `jank compile-module` puts the module file in the cwd.

lein-jank/src/leiningen/jank.clj
kylc: Fixed

lein-jank/src/leiningen/jank.clj
jeaye: `--name` currently only applies to the executable, within the target dir.

However, `jank compile-module` should use the same target and build dirs as anything else. If it's showing up in the CWD then that's indeed a bug. I've made a note to look into that. Feel free to investigate if you'd like.

jeaye:
Let's wait for main to finish building Clang in CI. Then we can use the cached version.

jeaye:
Ok, if you merge in the latest main and push, we should be able to use the CI cache now.

jeaye (approved):
Thank you!

PR #892, opened 2026-07-09
Add ability to link static libs into JIT runtime

jeaye:
With this PR, jank will support linking static libs via `-lfoo`. If both static and dynamic are present, jank will choose the dynamic lib, following Clang's behavior. To force the static lib, jank supports the `-l:foo` syntax. If `-l:foo` is provided and there's not static lib, but there is a dynamic lib, the dynamic lib will not be chosen.

### DIFF
--- a/compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
@@ -52,9 +52,15 @@ namespace jank::jit
 #endif
 
   static jtl::immutable_string static_lib_name(jtl::immutable_string const &lib)
+#if defined(JANK_WINDOWS_LIKE)
+  {
+    return util::format("lib{}.lib", lib);
+  }
+#else
   {
     return util::format("lib{}.a", lib);
   }
+#endif
 
   static bool is_static_lib(jtl::immutable_string const &lib)
   {
@@ -524,20 +530,21 @@ namespace jank::jit
 
   jtl::option<jtl::immutable_string> processor::find_lib(jtl::immutable_string const &lib) const
   {
+    if(lib.starts_with("./"))
+    {
+      if(std::filesystem::is_regular_file(lib.c_str()))
+      {
+        return std::filesystem::absolute(lib.c_str());
+      }
+      return none;
+    }
+
     for(auto const &lib_dir : library_dirs)
     {
-      auto default_lib_abs_path{ util::format("{}/{}", lib_dir.string(), lib) };
-      if(std::filesystem::exists(default_lib_abs_path.c_str()))
+      auto lib_abs_path{ util::format("{}/{}", lib_dir.string(), lib) };
+      if(std::filesystem::is_regular_file(lib_abs_path.c_str()))
       {
-        return default_lib_abs_path;
-      }
-      else
-      {
-        auto lib_abs_path{ util::format("{}/{}", lib_dir.string(), lib) };
-        if(std::filesystem::exists(lib_abs_path.c_str()))
-        {
-          return lib_abs_path;
-        }
+        return lib_abs_path;
       }
     }
 
@@ -549,8 +556,15 @@ namespace jank::jit
   {
     for(auto const &lib : libs)
     {
-      if(std::filesystem::path{ lib.c_str() }.is_absolute())
+      /* If we have an absolue path, there's nothing more to search for. Just load.
+       * Example: -l/foo/bar/libfoo.so */
+      std::filesystem::path const lib_path{ lib.c_str() };
+      if(lib_path.is_absolute())
       {
+        if(!std::filesystem::is_regular_file(lib_path))
+        {
+          return err(util::format("Failed to find library '{}'.", lib));
+        }
         if(is_static_lib(lib))
         {
           load_static_library(lib);
@@ -559,20 +573,67 @@ namespace jank::jit
         {
           load_dynamic_library(lib);
         }
+
+        continue;
       }
-      else if(lib.starts_with(':'))
+
+      /* Try finding the lib literally, in case it contains a file name or a relative path.
+       * Example: -llibfoo.a or -l./libfoo.so */
       {
+        auto const result{ processor::find_lib(lib) };
+        if(result.is_some())
+        {
+          auto const &found_lib{ result.unwrap() };
+          if(is_static_lib(found_lib))
+          {
+            load_static_library(found_lib);
+          }
+          else
+          {
+            load_dynamic_library(found_lib);
+          }
+          continue;
+        }
+      }
+
+      /* If the lib starts with a colon, force static lib loading, by still try to find it
+       * normally.
+       * Example: -l:foo or -l:libfoo.a or -l:./libfoo.so */
+      if(lib.starts_with(':'))
+      {
+        auto result{ processor::find_lib(lib.substr(1)) };
+        if(result.is_some())
+        {
+          if(!is_static_lib(result.unwrap()))
+          {
+            return err(
+              util::format("Failed to find static library '{}'. This library is not static.",
+                           lib.substr(1)));
+          }
+          load_static_library(result.unwrap());
+          continue;
+        }
+
         auto const &stat{ static_lib_name(lib.substr(1)) };
-        auto const result{ processor::find_lib(stat) };
+        result = processor::find_lib(stat);
         if(result.is_none())
         {
-          return err(util::format("Failed to load static library '{}'.", lib.substr(1)));
+          return err(util::format("Failed to find static library '{}'.", lib.substr(1)));
         }
         else
         {
+          if(!is_static_lib(result.unwrap()))
+          {
+            return err(
+              util::format("Failed to find static library '{}'. This library is not static.",
+                           lib.substr(1)));
+          }
           load_static_library(result.unwrap());
         }
       }
+      /* Otherwise, just try to find the lib as first a dynamic lib and then a static lib.
+       * This is the typical case.
+       * Example: -lfoo */
       else
       {
         auto const &shared{ shared_lib_name(lib) };
@@ -583,7 +644,7 @@ namespace jank::jit
           result = processor::find_lib(stat);
           if(result.is_none())
           {
-            return err(util::format("Failed to load library '{}'.", lib));
+            return err(util::format("Failed to find library '{}'.", lib));
           }
           else
           {
--- a/compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
@@ -530,6 +530,18 @@ namespace jank::jit
 
   jtl::option<jtl::immutable_string> processor::find_lib(jtl::immutable_string const &lib) const
   {
+    std::filesystem::path const lib_path{ lib.c_str() };
+    if(lib_path.is_absolute())
+    {
+      if(!std::filesystem::is_regular_file(lib_path))
+      {
+        return none;
+      }
+      return lib_path.string();
+
+      return none;
+    }
+
     if(lib.starts_with("./"))
     {
       if(std::filesystem::is_regular_file(lib.c_str()))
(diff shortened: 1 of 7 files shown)


jeaye:
@kylc This should give us all we need to default to static linking in lein-jank.

compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
```diff
@@ -51,6 +51,16 @@ namespace jank::jit
...
 #endif
 
+  static jtl::immutable_string static_lib_name(jtl::immutable_string const &lib)
+  {
+    return util::format("lib{}.a", lib);
+  }
+
+  static bool is_static_lib(jtl::immutable_string const &lib)
+  {
+    return lib.ends_with(".a");
```
kylc, line 67: Windows may be `.lib`

compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
```diff
@@ -537,20 +545,50 @@ namespace jank::jit
...
+          load_static_library(lib);
+        }
+        else
+        {
+          load_dynamic_library(lib);
+        }
+      }
+      else if(lib.starts_with(':'))
+      {
+        auto const &stat{ static_lib_name(lib.substr(1)) };
```
kylc, line 608: I think `-l:` behavior is more nuanced than this. `-lfoo` would be more like `-l:libfoo.a`.

GCC manual:
> -lnamespec
--library=namespec
    Add the archive or object file specified by namespec to the list of files to link. This option may be used any number of times. If namespec is of the form :filename, ld will search the library path for a file called filename, otherise it will search the library path for a file called libnamespec.a. 

I'm sure this will be a niche feature though, so not sure if you care to support it.

kylc: I think it's sufficient to change this to `auto const &stat{ lib.substr(1) };` (remove the `static_lib_name` call), since the given name after the `:` prefix _is_ the static lib name.

jeaye: I prefer the existing behavior. :man_shrugging:  Works like normal `-l`, but only looks for static libs. Since Clang doesn't support it anyway, we have flexibility with how things work. I'm not a fan of `-l:libfoo.a`.

We could use a different flag instead, but I don't think it matters much.

compiler+runtime/src/cpp/jank/jit/processor_dynamic.cpp
jeaye: Ah, right.

kylc:
Very cool, on this branch I was able to compile my `minitui` example and statically link `ftxui`. 

```
$ ldd target/release/minitui 
	linux-vdso.so.1 (0x000070a5215fc000)
	libm.so.6 => /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libm.so.6 (0x000070a520908000)
	libz.so.1 => /nix/store/61a1nwx3w6rqyaisj5rn1sal1981apm7-zlib-1.3.2/lib/libz.so.1 (0x000070a5208e9000)
	libzstd.so.1 => /nix/store/gd74vdpqzq0qmmikvmrf3r3kghhhnf50-zstd-1.5.7/lib/libzstd.so.1 (0x000070a520814000)
	libstdc++.so.6 => /nix/store/chqq8mpmpyfi9kgsngya71akv5xicn03-gcc-15.2.0-lib/lib/libstdc++.so.6 (0x000070a520400000)
	libgcc_s.so.1 => /nix/store/chqq8mpmpyfi9kgsngya71akv5xicn03-gcc-15.2.0-lib/lib/libgcc_s.so.1 (0x000070a5207e7000)
	libc.so.6 => /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6 (0x000070a520000000)
	/nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/ld-linux-x86-64.so.2 => /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib64/ld-linux-x86-64.so.2 (0x000070a5215fe000)
$ du -hc target/release/minitui 
14M	target/release/minitui
14M	total
```

jeaye:
> Very cool, on this branch I was able to compile my `minitui` example and statically link `ftxui`.
> 
> ```
> $ ldd target/release/minitui 
> 	linux-vdso.so.1 (0x000070a5215fc000)
> 	libm.so.6 => /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libm.so.6 (0x000070a520908000)
> 	libz.so.1 => /nix/store/61a1nwx3w6rqyaisj5rn1sal1981apm7-zlib-1.3.2/lib/libz.so.1 (0x000070a5208e9000)
> 	libzstd.so.1 => /nix/store/gd74vdpqzq0qmmikvmrf3r3kghhhnf50-zstd-1.5.7/lib/libzstd.so.1 (0x000070a520814000)
> 	libstdc++.so.6 => /nix/store/chqq8mpmpyfi9kgsngya71akv5xicn03-gcc-15.2.0-lib/lib/libstdc++.so.6 (0x000070a520400000)
> 	libgcc_s.so.1 => /nix/store/chqq8mpmpyfi9kgsngya71akv5xicn03-gcc-15.2.0-lib/lib/libgcc_s.so.1 (0x000070a5207e7000)
> 	libc.so.6 => /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6 (0x000070a520000000)
> 	/nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/ld-linux-x86-64.so.2 => /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib64/ld-linux-x86-64.so.2 (0x000070a5215fe000)
> $ du -hc target/release/minitui 
> 14M	target/release/minitui
> 14M	total
> ```

Oh nice. So that means it circumvents the ftxui collisions that were happening before?

Also, that 14M should `strip` down to 10M or so. We don't yet have a flag to strip the binary in jank or lein-jank. Not sure if we want one.

kylc:
> Oh nice. So that means it circumvents the ftxui collisions that were happening before?

Maybe... I saw some symbol collisions in Debug mode, but not in Release. I guess the linker is gc'ing the colliding symbols when optimizations are enabled. I didn't look into it much. But that's an ftxui-specific issue, so I ran some other tests:

I tried static linking a slint program and an imgui one, and I stripped both executables.

```
$  ls -lh target/release/jank-slint-example
-rwxr-xr-x 1 kyle users 53M Jul  9 11:47 target/release/jank-slint-example
$ ls -lh target/release/imgui-hello
-rwxr-xr-x 1 kyle users 11M Jul  9 11:50 target/release/imgui-hello
```

Compressing with `upx --best` is also possible for smaller executables (for the jank marketing poster :laughing:):

```
$ upx --best target/release/jank-slint-example && ls -lh target/release/jank-slint-example
-rwxr-xr-x 1 kyle users 18M Jul  9 11:47 target/release/jank-slint-example
$ upx --best target/release/imgui-hello && ls -lh target/release/imgui-hello
-rwxr-xr-x 1 kyle users 3.1M Jul  9 11:50 target/release/imgui-hello
```

Presumably LTO will also be possible for AOT executables at some point?

jeaye:
> > Oh nice. So that means it circumvents the ftxui collisions that were happening before?
> 
> Maybe... I saw some symbol collisions in Debug mode, but not in Release. I guess the linker is gc'ing the colliding symbols when optimizations are enabled. I didn't look into it much. 

Ok, gotcha. The best thing we can do is limit the libs that jank uses, since they'll always affect JIT symbol resolution, even if they're not in the PCH. I'll be stripping out ftxui in a couple of weeks, when I get to working on error messages again. I'll be trying to get more Clang diagnostic info so we can do things like show overload candidates on overload resolution failures.

> 
> I tried static linking a slint program and an imgui one, and I stripped both executables.
> 
> ```
> $  ls -lh target/release/jank-slint-example
> -rwxr-xr-x 1 kyle users 53M Jul  9 11:47 target/release/jank-slint-example
> $ ls -lh target/release/imgui-hello
> -rwxr-xr-x 1 kyle users 11M Jul  9 11:50 target/release/imgui-hello
> ```
> 
> Compressing with `upx --best` is also possible for smaller executables (for the jank marketing poster 😆):
> 
> ```
> $ upx --best target/release/jank-slint-example && ls -lh target/release/jank-slint-example
> -rwxr-xr-x 1 kyle users 18M Jul  9 11:47 target/release/jank-slint-example
> $ upx --best target/release/imgui-hello && ls -lh target/release/imgui-hello
> -rwxr-xr-x 1 kyle users 3.1M Jul  9 11:50 target/release/imgui-hello
> ```

Oh, wow. I've not heard of upx prior to this, but I just looked it up after seeing these results. That's impressive! Seems like we could add upx support into lein-jank? Depending on how fast it is, we could do it by default for release builds, though we'd want to `strip` first. Unless there's a good reason to not to it, we might as well opt for the smallest binaries by default. As you said, it makes for an even more impressive marketing campaign.

> Presumably LTO will also be possible for AOT executables at some point?

Yes, but there are two caveats which have prevented me from supporting it just yet.

1. We need to have a dedicated build of libjank-standalone with LTO enabled, since LTO needs to be passed at compilation time AND link time. This is because the compiled "object" files for LTO are actually not object files; they're LLVM IR bitcode.
2. We need to manually do tree shaking to remove unused vars, but that requires tracing all var usages, which either requires storing var usage metadata in compiled artifacts OR compiling all jank sources together into one translation unit.

Compiling all sources together into one translation unit also present big win opportunities for direct linking (or direct calls, as jank calls it), monomorphization, and escape analysis, which will aid in optimization even further. So I intend on starting with this approach, by adding an `-Omega` flag, to enable a "mega" build (aka unity build). With a mega build enabled, we'll then do an optimization pass at the end to trim out unused vars. Without this, since every var references its corresponding fns, LTO isn't going to help much.

If you have some feedback on this approach, let me know. The big downside of `-Omega` is that every build will be a fresh build and a build of the entire application from source.

If we go the route of storing compiler data about vars/fns as an artifact, we could store things like:

1. Which vars are used (and by which vars)
2. Escape analysis info (which fns are pure, which fns capture which params)

We can likely achieve the same benefits this way, while not impacting build times as much, but it's going to be much harder to manage.

kylc:
> Seems like we could add upx support into lein-jank?

I'm not so familiar with the tradeoffs, but I'm sure there are some. `upx --best` is very slow, but you can choose the compression level to run faster. I think it's used in the Go world.

I also think that GraalVM does some sort of compression when you compile with `-Os`, because I remember seeing that `libz.so` showed up in the linked executable only in this configuration. Maybe worth keeping in mind when comparing binary sizes, so it's apples to apples.

> If you have some feedback on this approach, let me know. The big downside of -Omega is that every build will be a fresh build and a build of the entire application from source.

Long build times for the "final final" executable doesn't seem like a big issue. In Rust it's common to have debug/release profiles (like we typically do in jank) and then a further release-lto profile which enables LTO, disables compiler parallelization (because it splits up the translation units), disables incremental compilation, strips, etc. Since you only run it when releasing binaries, the build times don't matter.

jeaye:
> I also think that GraalVM does some sort of compression when you compile with -Os, because I remember seeing that libz.so showed up in the linked executable only in this configuration. Maybe worth keeping in mind when comparing binary sizes, so it's apples to apples.

Oh, interesting. jank's binaries are already nearly half the size of Graal native images. If they're using compression, we can be even more competitive with an apples:apples.

> Long build times for the "final final" executable doesn't seem like a big issue. In Rust it's common to have debug/release profiles (like we typically do in jank) and then a further release-lto profile which enables LTO, disables compiler parallelization (because it splits up the translation units), disables incremental compilation, strips, etc. Since you only run it when releasing binaries, the build times don't matter.

Agreed. Sounds like we should add a third profile for this. Might as well call it `:release-lto` as well. That will default to `-Omega -Olto -Odirect-call` and so on. Much longer build, but everything will be as baked down as it can be.

jeaye:
Okay, I've followed up with a test suite and some more robust support. This should cover the GCC behavior as well. Covered cases are (with negative cases separated):

```
; Cases
;   Absolute
;     -l/foo/bar/libfoo.so
;     -l/foo/bar/libfoo.a
;
;     -l/foo/bar/libmissing.a
;     -l/foo/bar
;   Relative
;     -l./libfoo.so
;     -l./libfoo.a
;
;     -l./libmissing.so
;     -l./
;   File
;     -llibfoo.so
;     -llibfoo.a
;
;     -llibmissing.so
;   Static
;     -l:foo (with static lib)
;     -l:./libfoo.a
;     -l:libfoo.a
;     -l:/foo/bar/libfoo.a
;
;     -l:foo (with dynamic lib)
;     -l:./libfoo.so
;     -l:libfoo.so
;     -l:libmissing.a
;   General
;     -lfoo (with static lib)
;     -lfoo (with dynamic lib)
;     -lfoo (with both libs)
;
;     -lmissing
```

PR #894, opened 2026-07-10
Support :managed-dependencies in lein-jank.

kylc:
This is just restoring some of the upstream lein features we lost when we bypassed lein's dependency resolution, since it doesn't behave correctly in the presence of native deps (removes duplicate versions from the tree silently).

I think this will be an especially important feature when writing an application, and including native dependencies which may not agree on their own downstream dependency versions, which would normally lead to compiling/linking errors.

### DIFF
--- a/lein-jank/src/leiningen/jank/resolve.clj
+++ b/lein-jank/src/leiningen/jank/resolve.clj
@@ -1,5 +1,5 @@
 (ns leiningen.jank.resolve
-  (:require [leiningen.core.classpath :refer [default-aether-args]]
+  (:require [leiningen.core.classpath :refer [default-aether-args normalize-dep-vector]]
             [cemerick.pomegranate.aether :as aether]))
 
 (defn dependency-hierarchy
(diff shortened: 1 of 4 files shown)


lein-jank/src/leiningen/jank/resolve.clj
```diff
@@ -8,15 +8,23 @@
...
+        coordinates   (aether/merge-versions-from-managed-coords [dep] managed-deps)
+        opts          (merge (default-aether-args project) {:coordinates         coordinates
+                                                            :managed-coordinates managed-deps})
         full-tree     (aether/resolve-dependencies opts)
         dep-with-meta (first (filter #{dep} (keys full-tree)))]
     (when dep-with-meta
       {dep-with-meta
        (->> (get full-tree dep)
-            (mapv (fn [d] [d (-> (dependency-hierarchy opts d) vals first)]))
+            (mapv (fn [d] [d (-> (dependency-hierarchy project managed-deps d) vals first)]))
```
kylc, line 27: RE: `opts -> project` I'm amazed anything worked with this typo 🤷

jeaye: 🤔 🤷‍♂️

jeaye (approved):
Thanks for keeping on this, Kyle.

PR #898, opened 2026-07-14
Fix lexing of multi-byte UTF-8 characters in comments

mudPi314:
## Problem

Any comment containing non-ASCII text fails to lex:

```clojure
; comment with ▀ in it
(println "ok")
```

```
─ lex/invalid-unicode ────────────
error: Unfinished character.
```

## Cause

Both comment-scanning loops — line comments (`case ';'`) and shebang comments (`#!`) — advance by a single byte (`++pos`) after `peek()` decodes a full codepoint. For a multi-byte UTF-8 character, the next `peek()` lands mid-sequence, returns an error, and silently breaks out of the comment loop; the orphaned continuation bytes are then lexed as a new token, surfacing as a spurious `lex/invalid-unicode` error.

## Fix

Advance by the decoded codepoint's byte length (`pos += oc.expect_ok().len;`), exactly as symbol lexing already does a few lines above.

## Tests

Added lexer subcases covering 2-, 3-, and 4-byte codepoints (`é`, `▀`, `🍜`):
- `;` comment with multi-byte content followed by more code
- `;` comment with multi-byte content at EOF
- `#!` comment with multi-byte content

Verified locally against a build of the current lexer: without the fix, exactly these three new subcases fail (the `;` cases via the reported error, matching the user-visible bug); with the fix, the full lexer suite passes (20 test cases, 220 assertions).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/compiler+runtime/src/cpp/jank/read/lex.cpp
+++ b/compiler+runtime/src/cpp/jank/read/lex.cpp
@@ -653,7 +653,7 @@ namespace jank::read::lex
               hit_non_semi = true;
             }
 
-            ++pos;
+            pos += oc.expect_ok().len;
           }
           if(pos == token_start)
           {
@@ -1396,7 +1396,7 @@ namespace jank::read::lex
                     break;
                   }
 
-                  ++pos;
+                  pos += oc.expect_ok().len;
                 }
 
                 ++pos;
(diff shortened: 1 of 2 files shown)


jeaye:
Thanks for the PR. The actual changes look good, but jank's contribution policy currently disallows AI-generated contributions from external contributors. I expect that our policy will change as things get figured out, but I am not ready to change it yet because I am not prepared to handle the influx of AI-generated PR, to gauge which are worthy of review, and to deal with the social and legal repercussions.

So I'll close this now and apply a similar change myself. Despite not merging this PR, due to the current policy, I do appreciate this PR, since it points out the bug and the quick fix.

PR #900, opened 2026-07-16
Fix `lein test` test var resolution

kylc:
`-workaround-get-test` was removed in 308a8570c. Matches [upstream implementation](https://github.com/technomancy/leiningen/blob/main/src/leiningen/test.clj#L64C22-L64C42) now.

Tested on a hello world clojure.test project.


jeaye:
Nice! I had a local TODO for this. Thank you!

PR #901, opened 2026-07-16
Standardize `lein test` CLI and native build handling

kylc:
- Native build step now runs before tests like other lein-jank commands
- `lein help test` returns a nice string like the other lein-jank commands
- Standard CLI options like -v and --disable-sandbox now apply
- Passing a file to `lein test` works now


jeaye (approved):
Nice! Thanks, Kyle.

PR #908, opened 2026-07-19
Extract jank-build into its own lein-free namespace

kylc:
Let's keep jank-build small and not depending on leiningen. It now uses only babashka-compatible dependencies and Java classes.

There is still some implicit leiningen dependency on the `project.clj` structure, which we pass around in jank-build. We can tackle this later.

- Moved any lein-specific behavior (like dependency resolution) up above jank-build
- Added a unit test to load `jank-build.core` via bb to ensure we don't break compatibility
- Little bit of cleanup here and there

### DIFF
--- a/lein-jank/src/jank_build/change_detection.clj
+++ b/lein-jank/src/jank_build/change_detection.clj
@@ -1,7 +1,8 @@
-(ns leiningen.jank.changed
-  "For tracking changes in the environment and in build source files."
+(ns jank-build.change-detection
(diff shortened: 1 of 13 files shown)


jeaye (approved):
Ah, nice. This will make things easier. :)

jeaye:
I've published a new lein-jank, with all of the recent changes you've merged as well. I don't yet have automation set up for that, like in jank commons, but that's hopefully coming soon.

PR #923, opened 2026-07-25
Speed up lein-jank by lazy loading

kylc:
Another approach to speeding up lein-jank.

There are a few namespaces which each add 100-200ms of load time: clojure.tools.namespace (this one is really bad), babashka.fs, and babashka.process. If we can avoid loading these in commands that don't use them, or replace them entirely (babashka.process -> clojure.java.process), then everything gets faster.

Here are the before/after comparison (I installed the lein-jank from this PR during the sleep):

```
$ hyperfine --setup 'sleep 5' -r 5 'lein run --help' 'lein run --help'
Benchmark 1: lein run --help
  Time (mean ± σ):      1.202 s ±  0.028 s    [User: 1.649 s, System: 0.174 s]
  Range (min … max):    1.167 s …  1.244 s    5 runs
 
Benchmark 2: lein run --help
  Time (mean ± σ):     689.2 ms ±  16.8 ms    [User: 1024.2 ms, System: 125.5 ms]
  Range (min … max):   673.8 ms … 716.3 ms    5 runs
 
Summary
  lein run --help ran
    1.74 ± 0.06 times faster than lein run --help
```
```
$ hyperfine --setup 'sleep 5' -r 5 'lein run -v' 'lein run -v'
Benchmark 1: lein run -v
  Time (mean ± σ):      1.319 s ±  0.035 s    [User: 1.751 s, System: 0.199 s]
  Range (min … max):    1.275 s …  1.359 s    5 runs
 
Benchmark 2: lein run -v
  Time (mean ± σ):     984.0 ms ±  29.4 ms    [User: 1386.5 ms, System: 188.9 ms]
  Range (min … max):   951.5 ms … 1015.6 ms    5 runs
 
Summary
  lein run -v ran
    1.34 ± 0.05 times faster than lein run -v
```

There is probably more speedup if we can replace `babashka.fs` in jank-build, but we use it pretty heavily so I haven't done it here.

Also improved the message when the `jank` binary cannot be found/run.

### DIFF
--- a/lein-jank/project.clj
+++ b/lein-jank/project.clj
@@ -4,9 +4,9 @@
   :license {:name "MPL 2.0"
             :url "https://www.mozilla.org/en-US/MPL/2.0/"}
   :dependencies [[babashka/fs "0.5.33"]
-                 [babashka/process "0.6.25"]
                  [leiningen-core "2.12.0"]
                  [org.clojure/tools.cli "1.4.256"]
                  [org.clojure/tools.namespace "1.5.1"]]
-  :profiles {:dev {:dependencies [[nubank/matcher-combinators "3.10.0"]]}}
+  :profiles {:dev {:dependencies [[babashka/process "0.6.25"]
+                                  [nubank/matcher-combinators "3.10.0"]]}}
   :eval-in-leiningen true)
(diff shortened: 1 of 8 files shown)


jeaye (approved):
Nice! I'm open to this. Hopefully, with the new Clojure CLI (ETA ~September/October), our tooling perf issues go away entirely. This will still help us be efficient, even then.

Thanks, Kyle!

jeaye:
Is this one ready for merging?

kylc:
Yep this is ready to go.

PR #924, opened 2026-07-25
Allow :static? flag in project.clj

kylc:
Some flags don't result in jank CLI flags, but are used in jank-build. We shouldn't print a warning when these are in the project.clj `:jank` section.


jeaye (approved):
Yep, works for me.

PR #928, opened 2026-07-26
Enable `clojure.string-test.capitalize`

djblue:


compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -143,26 +144,29 @@
   [s]
   (when (nil? s)
     (throw "upper-case does not support nil"))
-  (cpp/clojure.string_native.upper_case (.to_string s)))
+  (-> (cpp/cast jtl.immutable_string (.to_string s))
+      (.c_str)
```
jeaye: Hm, the `cpp/cast` should not be needed here. `.to_string` already returns a `jtl::immutable_string`.

compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -1,6 +1,7 @@
 (ns clojure.string
   (:refer-clojure :exclude [replace reverse])
-  (:include "clojure/string_native.hpp"))
+  (:include "jank/util/string.hpp"
```
jeaye, line 3: These fns take `std::string`, but it looks like nobody is even using them. We could just update them to use `jtl::immutable_string` so we can remove the `c_str` call below. If you're interested in that, please tackle it. Otherwise, please add a TODO in `util/string.hpp` for it.

compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -143,26 +144,29 @@
...
     (throw "capitalize does not support nil"))
-  (let [s (str s)]
-    (if (< (count s) 2)
-      (upper-case s)
-      (str (upper-case (subs s 0 1))
-           (lower-case (subs s 1))))))
+  (-> (cpp/cast jtl.immutable_string (.to_string s))
+      (.c_str)
+      (cpp/jank.util.to_lowercase)
+      (doto (cpp/jank.util.capitalize))))
```
jeaye: Fascinating usage of `doto` to handle the mutation. I wonder if this will become an idiom in jank.

jeaye (changes requested):
Thanks, Chris!

jeaye (approved):
Works for me!

PR #929, opened 2026-07-26
Enable `clojure.core-test.char`

djblue:
Depends on https://github.com/jank-lang/clojure-test-suite/pull/946 for tests to pass.

```clojure
% jank repl
user=> (seq "𐅧")
(\𐅧)
user=> (first (seq "𐅧"))
\𐅧
user=> (seq "𐅧 𐅧 𐅧")
(\𐅧 \space \𐅧 \space \𐅧)
user=> (count (seq "𐅧 𐅧 𐅧"))
5
user=> (map str (seq "𐅧 𐅧 𐅧"))
("𐅧" " " "𐅧" " " "𐅧")
user=> (map (comp count str) (seq "𐅧 𐅧 𐅧"))
(4 1 4 1 4)
user=> (count "𐅧 𐅧 𐅧")
14
user=> (seq "👋 😃")
(\👋 \space \😃)
user=> (nth (seq "👋 😃") 0)
\👋
user=> (nth (seq "👋 😃") 2)
\😃
user=> (apply str (seq "👋 😃"))
"👋 😃"
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
@@ -105,7 +105,7 @@ namespace jank::runtime::obj
 
   persistent_string_sequence_ref persistent_string_sequence::next() const
   {
-    auto n(index + jtl::next_char_size(str->data, index));
+    auto const n(index + jtl::next_char_size(str->data, index));
 
     if(n == str->data.size())
     {
--- a/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
+++ b/compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
@@ -1,5 +1,8 @@
+#include <jtl/utf8.hpp>
+
 #include <jank/runtime/obj/persistent_string_sequence.hpp>
 #include <jank/runtime/obj/persistent_string.hpp>
+#include <jank/runtime/obj/character.hpp>
 #include <jank/runtime/core/seq_ext.hpp>
 #include <jank/runtime/core/make_box.hpp>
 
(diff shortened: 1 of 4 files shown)


compiler+runtime/src/cpp/jank/runtime/obj/persistent_string_sequence.cpp
```diff
@@ -59,7 +76,13 @@ namespace jank::runtime::obj
   /* behavior::countable */
   usize persistent_string_sequence::count() const
   {
-    return str->data.size() - index;
+    usize count{};
+    auto const size(str->data.size());
+    for(auto i(index); i < size; i += jtl::next_char_size(str->data, i))
```
djblue, line 82: Technically this implementation is no longer `O(1)`, is that ok? It is still better than going through [`sequence_length/behavior::seqable<T>`](https://github.com/jank-lang/jank/blob/main/compiler%2Bruntime/src/cpp/jank/runtime/core/seq.cpp#L960) as it will avoid all the unnecessary allocations.

djblue: Also, should we cache the length once computed? We could propagate it via next as well.

jeaye: I think it's ok as is, until it becomes an issue. We can leave a TODO noting that this could be cached.

compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
```diff
@@ -62,7 +62,7 @@
     clojure.core-test.double-qmark
     clojure.core-test.drop
     ; clojure.core-test.drop-last ; "TODO: port int-array"
-    clojure.core-test.drop-while
+    ; clojure.core-test.drop-while ; "TODO: port int-array" /  "TODO: port sequence"
```
jeaye, line 65: I'm working on adding support for these arrays. Hopefully we can enable all of these soon!

jeaye (dismissed):
Looks good! Waiting to merge for the repoint.

jeaye (approved):
Thanks!

PR #933, opened 2026-07-26
Enable `clojure.core-test.plus-squote`

djblue:


compiler+runtime/src/cpp/jank/read/parse.cpp
```diff
@@ -848,26 +848,26 @@ namespace jank::read::parse
...
     }
     else if(sym->name == "-Inf")
     {
-      n = -std::numeric_limits<f64>::infinity();
+      n = jank_neg_inf;
     }
     else if(sym->name == "NaN")
     {
-      n = std::numeric_limits<f64>::quiet_NaN();
+      n = jank_nan;
```
djblue, line 862: Any idea why I'm getting the following:
```clojure
% jank repl
user=> (identical? ##NaN ##NaN)
false
```
I did confirm with clj:
```clojure
% clj
Clojure 1.12.5
user=> (identical? ##NaN ##NaN)
true
```

djblue: This works however:
```clojure
% jank repl
user=> (identical? (identity ##NaN) ##NaN)
true
user=> (identical? ##NaN (identity ##NaN)) ;; but not this
false
user=> (cpp/== cpp/jank.runtime.jank_nil cpp/jank.runtime.jank_nil)
true
```

djblue: Either way, this PR does work, so we could defer this research until later.

jeaye: The best way to inspect something like this is to look at the generated C++.

```cpp
❯ JANK_PRINT_CODEGEN=1 jank repl
nREPL server started on port 42003 on host 127.0.0.1 - nrepl://127.0.0.1:42003
user=> (let [r (identical? ##NaN ##NaN)])
...
extern "C" jank::runtime::object_ref user_let_2_0(jank::runtime::object_ref const let)
{
#line 3494 "/home/jeaye/projects/jank/compiler+runtime/src/jank/clojure/core.jank"
  {
    auto v2(NAN);
    auto v3(NAN);
    auto &&v4(v2 == v3);
    auto const v5(jank::runtime::jank_nil);
    return v5;

#line 3494 "/home/jeaye/projects/jank/compiler+runtime/src/jank/clojure/core.jank"
  }
}
```

So we can see that `NAN` is unboxed and you're just comparing two values at that point. Clojure doesn't require that numbers are boxed, as far as I know, and it doesn't require that two literals are identical. It worked when you introduced `identity`, since you forced boxing.

compiler+runtime/src/cpp/jank/runtime/core/math.cpp
```diff
@@ -50,10 +50,33 @@ namespace jank::runtime
       [](auto const typed_l, auto const r) -> object_ref {
         using LT = typename decltype(typed_l)::value_type;
 
+        if constexpr(std::same_as<LT, obj::real>)
```
jeaye: Anywhere we handle `obj::real`, we also need to handle `obj::small_real`.

```cpp
jtl::is_any_same<LT, obj::real, obj::small_real>
```

compiler+runtime/src/cpp/jank/runtime/obj/number.cpp
```diff
@@ -488,8 +488,32 @@ namespace jank::runtime
     return &r;
   }
 
+  static obj::real_ref nan_const()
```
jeaye: I think we should just make these all `small_real_ref` instead. `real_ref` should only be used when we're doing heap allocations.

jeaye (changes requested):
Thanks, Chris!

compiler+runtime/src/cpp/jank/runtime/obj/number.cpp
djblue: Can we force `##NaN` into a `small_real`?

```clojure
% jank repl
user=> (type 1.0)
"small_real"
user=> (type ##NaN)
"real"
```

jeaye: Yeah, just parse it into one. :) Right now, you're parsing it into a `real_ref` in this PR.

djblue: So switching `jank_nan` to a `small_real` was a bit more work than I expected. It exposed some difference in how the `real` vs `small_real` version of `NAN` were being treated.

jeaye: Ah, thanks for the iteration. That sounds like an altogether good thing, since we need them to work the same.

djblue: It appears the new tests I added were failing because `cpp/std.isnan`/`cpp/std.isinf` weren't resolving on macos or windows so I just reverted them https://github.com/jank-lang/jank/pull/933/commits/7a11c52eabc68249f2abef13961d066c56135ec1. They did pass on linux which I think confirms the unboxed handling of NaN/Inf/-Inf was working correctly.

jeaye (approved):
Thanks, Chris!

PR #934, opened 2026-07-27
Enable `clojure.string-test.split-lines`

djblue:
Depends on https://github.com/jank-lang/clojure-test-suite/pull/948 for tests to pass.

```clojure
% ./pass-test 
warning: 'sleep' already referred to #'clojure.core/sleep in namespace 'clojure.core-test.portability' but has been replaced by #'clojure.core-test.portability/sleep

Testing clojure.string-test.split-lines

Ran 1 tests containing 15 assertions.
0 failures, 0 errors.
:clojure-test-suite-successful
```

### DIFF
--- a/compiler+runtime/src/cpp/clojure/string_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/string_native.cpp
@@ -400,17 +400,21 @@ namespace clojure::string_native
       std::sregex_token_iterator iter(search_str.begin(), search_str.end(), regex, -1);
       std::sregex_token_iterator const end;
 
-      for(auto peek(iter); iter != end; ++iter)
+      for(; iter != end; ++iter)
       {
-        auto const substr(iter->str());
+        vec.push_back(make_box<obj::persistent_string>(iter->str()));
+      }
 
-        if(++peek == end && substr.empty())
+      /* Discard all trailing empty strings to match java.lang.String/split behavior. */
+      auto n(vec.size());
+      for(; n > 0; --n)
+      {
+        if(!expect_object<obj::persistent_string>(vec[n - 1])->data.empty())
         {
           break;
         }
-
-        vec.push_back(make_box<obj::persistent_string>(substr));
       }
+      vec.take(n);
     }
 
     return make_box<obj::persistent_vector>(vec.persistent());
--- a/compiler+runtime/src/cpp/clojure/string_native.cpp
+++ b/compiler+runtime/src/cpp/clojure/string_native.cpp
@@ -395,17 +395,32 @@ namespace clojure::string_native
 
   obj::persistent_vector_ref split(jtl::immutable_string const &s, obj::re_pattern_ref const re)
   {
-    auto const regex(re->regex);
-
-    std::string const search_str{ s.c_str() };
-
     detail::native_transient_vector vec;
-    std::sregex_token_iterator iter(search_str.begin(), search_str.end(), regex, -1);
-    std::sregex_token_iterator const end;
 
-    for(; iter != end; ++iter)
+    if(s.empty())
     {
-      vec.push_back(make_box<obj::persistent_string>(iter->str()));
+      vec.push_back(make_box<obj::persistent_string>(s));
+    }
+    else
+    {
+      auto const regex(re->regex);
+
+      std::string const search_str{ s.c_str() };
+
+      std::sregex_token_iterator iter(search_str.begin(), search_str.end(), regex, -1);
+      std::sregex_token_iterator const end;
+
+      for(auto peek(iter); iter != end; ++iter)
+      {
+        auto const substr(iter->str());
+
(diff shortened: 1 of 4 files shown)


compiler+runtime/src/jank/clojure/string.jank
```diff
@@ -175,7 +175,7 @@
 (defn split-lines
   "Splits s on \\n or \\r\\n. Trailing empty lines are not returned."
   [s]
-  (cpp/clojure.string_native.split s #"\r?\n"))
+  (cpp/clojure.string_native.split s (cpp/jank.runtime.re_pattern "\\r?\\n")))
```
djblue: I know I shouldn't need to do this, but I'm not sure how to fix the following error:

```
Testing clojure.string-test.split-lines
Uncaught exception: invalid object type (expected re_pattern found small_real)
```

djblue: I think it might have to do with how the `clojure.string` library is build as the following works fine from the REPL:

```clojure
% jank repl
user=> (cpp/raw "#include <clojure/string_native.hpp>")
nil
user=> (cpp/clojure.string_native.split "foo\nbar" #"\r?\n")
["foo" "bar"]
```

djblue: AOT:
```cpp
jank::runtime::object_ref const_193{ _jank_null{} }; /* I think this is the problem. */
extern "C" jank::runtime::object_ref
clojure_string_split_lines_138_1(jank::runtime::object_ref const split_lines,
                                 jank::runtime::object_ref s)
{
  auto const v1(jank::runtime::convert<jtl::immutable_string>::from_object(s));
  auto const v2(const_193);
  auto const v3(
    jank::runtime::convert<jank::runtime::oref<jank::runtime::obj::re_pattern>>::from_object(v2));
  auto &&v4(clojure::string_native::split(v1, v3));
  return v4;
}

extern "C" void jank_load_clojure_string()
{
  new(&clojure::string::const_193) auto(
    jank::runtime::make_box<jank::runtime::obj::re_pattern>("\\r?\\n"));
}
```

REPL:
```cpp
% JANK_PRINT_CODEGEN=1 jank repl
user=> (cpp/clojure.string_native.split "foo\nbar" #"\r?\n")

extern "C" jank::runtime::object_ref user_cpp_call_1_0(jank::runtime::object_ref const cpp_call)
{
  auto v1("foo\nbar");
  jtl::immutable_string v2(static_cast<jtl::immutable_string::const_pointer_type>(v1));
  auto const v3(jank::runtime::object_ref{ (void *)0xfffa7fbef07d7720 });
  auto const v4(
    jank::runtime::convert<jank::runtime::oref<jank::runtime::obj::re_pattern>>::from_object(v3));
  auto &&v5(clojure::string_native::split(v2, v4));
  return v5;
}

["foo" "bar"]
```

djblue: https://github.com/jank-lang/jank/pull/934/changes/99673e5cde0ffde2a9f827e82ba777fb23f4d7ae appears to be the fix, please let me know otherwise.

jeaye: Yep, that looks good. Nice work!

compiler+runtime/src/cpp/clojure/string_native.cpp
```diff
@@ -395,17 +395,32 @@ namespace clojure::string_native
...
+      vec.push_back(make_box<obj::persistent_string>(s));
+    }
+    else
     {
-      vec.push_back(make_box<obj::persistent_string>(iter->str()));
+      auto const regex(re->regex);
+
+      std::string const search_str{ s.c_str() };
+
+      std::sregex_token_iterator iter(search_str.begin(), search_str.end(), regex, -1);
```
jeaye, line 400: Looks like this works with any bidirection iterator. Why do we need `search_str` to be a `std::string` instead of a `jtl::immutable_string`?

jeaye (dismissed):
One question, but if we can't do anything about it then I'm cool with this. Thanks, Chris!

compiler+runtime/src/cpp/clojure/string_native.cpp
djblue: Not sure, but I get the following error when I try:
```cpp
/home/chris/repos/jank/compiler+runtime/src/cpp/clojure/string_native.cpp:408:34: error: no matching constructor for initialization of 'std::sregex_token_iterator' (aka 'regex_token_iterator<__gnu_cxx::__normal_iterator<const char *, basic_string<char, std::char_traits<char>, std::allocator<char>>>>')
  408 |       std::sregex_token_iterator iter(s.begin(), s.end(), regex, -1);
      |                                  ^    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

jeaye (approved):
Nice work!

PR #935, opened 2026-07-29
Enable `clojure.core-test.int`

djblue:
Resolves the following codegen error:

```cpp
/home/chris/repos/jank/compiler+runtime/src/jank/clojure/core.jank:3495:16: error: call to constructor of 'jank::runtime::obj::small_integer_ref' (aka 'oref<small_integer>') is ambiguous
 3495 | auto const v89(jank::runtime::obj::small_integer_ref{ -2147483648 });
      |                ^                                    ~~~~~~~~~~~~~~~
/home/chris/repos/jank/compiler+runtime/include/cpp/jank/runtime/oref.hpp:1557:5: note: candidate constructor
 1557 |     oref(i32 const data) noexcept
      |     ^
/home/chris/repos/jank/compiler+runtime/include/cpp/jank/runtime/oref.hpp:1562:5: note: candidate constructor
 1562 |     oref(i64 const data) noexcept
      |     ^
```

### DIFF
--- a/compiler+runtime/src/cpp/jank/codegen/cpp_processor.cpp
+++ b/compiler+runtime/src/cpp/jank/codegen/cpp_processor.cpp
@@ -176,7 +176,7 @@ namespace jank::codegen
         }
         else if(runtime::detail::is_tagged_small_int(ptr))
         {
-          fmt_str = util::format("{}{ {} }",
+          fmt_str = util::format("{}{ static_cast<jtl::i32>({}) }",
                                  get_qualified_type_name(type),
                                  runtime::detail::as_integer(ptr));
         }
--- a/compiler+runtime/src/cpp/jank/codegen/cpp_processor.cpp
+++ b/compiler+runtime/src/cpp/jank/codegen/cpp_processor.cpp
@@ -176,9 +176,16 @@ namespace jank::codegen
         }
         else if(runtime::detail::is_tagged_small_int(ptr))
         {
-          fmt_str = util::format("{}{ static_cast<jtl::i32>({}) }",
-                                 get_qualified_type_name(type),
-                                 runtime::detail::as_integer(ptr));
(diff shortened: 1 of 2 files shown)


compiler+runtime/src/cpp/jank/codegen/cpp_processor.cpp
```diff
@@ -176,7 +176,7 @@ namespace jank::codegen
         }
         else if(runtime::detail::is_tagged_small_int(ptr))
         {
-          fmt_str = util::format("{}{ {} }",
+          fmt_str = util::format("{}{ static_cast<jtl::i32>({}) }",
```
djblue: I assume this is always a safe cast since the value satisfies `runtime::detail::is_tagged_small_int`. Do we want to use `int` instead of `jtl::i32`?

jeaye: No, it's helpful to always be clear about the int size we want.

jeaye: So, this cast is only needed for one value out of over four billion. I would rather not pay the code size cost of the cast for every int value, since this will seriously affect generated code sizes.

Instead, let's only insert the cast when the value is equal to `std::numeric_limits<i32>::min()`.

The reason this is a problem is that C++ doesn't have negative integer literals. Combine that with the range of `i32` to be `-2147483648` to `2147483647`. When a C++ compiler sees `-2147483648`, it sees the `2147483648` on its own, which is too large to fit in an `i32`, so it promotes it to an `i64` (`long`). Then it applies the `-` unary operator. So we end up with a larger int than we need because C++ is dumb. Literally no other values are affected, since they would all be valid `i32`.

jeaye (approved):
Thanks!

PR #938, opened 2026-07-30
WIP POC arrays

djblue:
I played around with some ideas for how we could get the best of both worlds for arrays in jank:

The following is currently working:
```clojure
user=> (get (byte-array 10 2) 0) ;; support clojure functions
2
user=> (bytes? (byte-array 10 2)) ;; supports element predicates
true
user=> (bytes? (double-array 10 2.0)) ;; supports element predicates
false
user=> (let [x (cpp/jank.runtime.obj.boolean_array 10)] ;; we can update boolean-array to support inlining to avoid doing the interop
[(cpp/aget (cpp/* x) 0) ;; support raw unboxed read
 (aset (cpp/* x) 0 true) ;; support raw unboxed write
 (cpp/aget (cpp/* x) 0) 
 (count x)  ;; converts to boxed object
 x
 (get x 0)])
[false true true 10 #object [array_box 0x7f6f5b20ed20] true]
```
We can also support unboxing the array via `booleans/bytes/chars/shorts/floats/ints/doubles/longs` so any object can get cast into a `array<T>`.

### DIFF
--- a/compiler+runtime/CMakeLists.txt
+++ b/compiler+runtime/CMakeLists.txt
@@ -580,6 +580,7 @@ add_library(
   src/cpp/jank/runtime/obj/uuid.cpp
   src/cpp/jank/runtime/obj/inst.cpp
   src/cpp/jank/runtime/obj/opaque_box.cpp
+  src/cpp/jank/runtime/obj/array.cpp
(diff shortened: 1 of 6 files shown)


djblue:
Closing in favor of https://github.com/jank-lang/jank/pull/946.

PR #939, opened 2026-07-31
fix windows dir lock race in temp directory cleanup

ikappaki:
Hi,

can you please consider this patch to fix the likely root cause of the issue that is causing the `ahead-of-time` pass test to intermittently fail on windows, as reported in #846.

Basically, windows does not immediately release the file handle on the working directory where a process executes, preventing an immediate delete of the temp directory where the ahead-of-time test CLI command is run. The workaround is to retry removing the directory after sleeping for some ms.

Thanks


jeaye:
Thanks for the fix! :)

PR #940, opened 2026-07-31
Add a print-cflags command for external build compatibility

kylc:
Tested by `clang++ hello.cpp $(jank print-cflags)`, where hello.cpp is from `jank compile-module --output-target cpp` with a placeholder main method (since `jank compile` won't output a C++ file).

Towards `zig cc` AOT compilation.

### DIFF
--- a/compiler+runtime/src/cpp/main.cpp
+++ b/compiler+runtime/src/cpp/main.cpp
@@ -412,7 +412,7 @@ int main(int const argc, char const **argv)
       else if(util::cli::opts.command == util::cli::command::print_cflags)
       {
         auto const compiler_args_res{ aot::build_compiler_args() };
-        std::vector<char const *> const compiler_args{ jtl::move(compiler_args_res.expect_ok()) };
+        auto const &compiler_args{ compiler_args_res.expect_ok() };
         for(auto const arg : compiler_args)
         {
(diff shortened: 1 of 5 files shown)


compiler+runtime/src/cpp/main.cpp
```diff
@@ -404,11 +404,26 @@ int main(int const argc, char const **argv)
...
+        auto const compiler_args_res{ aot::build_compiler_args() };
+        std::vector<char const *> compiler_args{ jtl::move(compiler_args_res.expect_ok()) };
+        for(auto const arg : compiler_args)
+        {
+          util::print("{} ", arg);
+        }
+        for(auto const arg : aot::build_linker_args())
+        {
+          util::print("{} ", arg);
+        }
```
kylc, line 423: We leak the `char*` flags here, but it hardly seems worth cleaning up since we are just exiting.

jeaye: Yeah, all of those `strdup` usages can be removed, eventually. They're not needed and the pattern was started by someone who didn't know that. :joy: Not a concern for now.

jeaye (dismissed):
Looks good to me! I'll do a bit of style cleanup once CI is green and then merge.

Thanks, Kyle!

PR #941, opened 2026-07-31
Remove ftxui

jeaye:


jeaye:
@kylc Heads up that ftxui has been removed now.

PR #942, opened 2026-08-01
Add `include-header` macro

djblue:
Before digging into making this a special form, I wanted to get something working to test our theories. The code in this PR seems to be working for me, however going through `eval_cpp_string` appears to be causing issues:
```diff
   void load_header(jtl::immutable_string const &header)
   {
-    __rt_ctx->eval(cpp_raw_include(header));
+    auto const cpp(header.starts_with("./") ? util::format("#include \"{}\"", header)
+                                            : util::format("#include <{}>", header));
+    __rt_ctx->eval_cpp_string(cpp).expect_ok();
   }
```
Will cause the following error on compile:
```
─ internal/failure ─────────────────────────────────────────────────────────────
error: Failed to find symbol: 'jank_load_clojure_core'                          
                                                                                

#0  in jank::error::internal_failure(jtl::immutable_string const&) at error.cpp:507
#1  in jank::error::throw_internal_failure(jtl::immutable_string const&) at error.cpp:513
#2  in void jtl::detail::panic<jtl::result<void*, jtl::immutable_string> >(jtl::result<void*, jtl::immutable_string> const&) at result.hpp:60
#3  in jtl::result<void*, jtl::immutable_string>::assert_ok() const at result.hpp:155
#4  in jtl::result<void*, jtl::immutable_string>::expect_ok() at result.hpp:160
#5  in jank::runtime::module::loader::load_o(jtl::immutable_string const&, jank::runtime::module::file_entry const&) const at loader_dynamic.cpp:807
#6  in jank::runtime::module::loader::load(jtl::immutable_string const&, jank::runtime::module::origin) at loader_dynamic.cpp:755
#7  in jank::runtime::context::load_module(jtl::immutable_string const&, jank::runtime::module::origin) at context.cpp:341
#8  in jank::runtime::context::compile_module(jtl::immutable_string const&) at context_dynamic.cpp:231
#9  in jank::compile_module() at main.cpp:184
#10 in main::$_0::operator()(int, char const**) const at main.cpp:487
#11 in main::$_0::__invoke(int, char const**) at main.cpp:388
#12 in jank_init_dynamic at c_api_dynamic.cpp:43
#13 in main at main.cpp:382
#14 at libc.so.6
#15 in __libc_start_main at libc.so.6
```


jeaye (changes requested):
Thanks for your continued help, Chris. :slightly_smiling_face: 

From what I can tell, this PR works around the `clojure.core` issue, but it doesn't provide a solution for every other user's includes. We don't want everyone to need to follow this same pattern of bootstrapping. Also, we need this to work for C++ output targets `--output-target cpp`, which is why I believe the special form is needed.

It may be helpful to clarify that the `_static.cpp` files are only ever linked into an AOT binary with a static runtime. So, for header inclusion, there's nothing for them to do at all. That part is right for `load_header`, but `include_header` shouldn't need to do anything.

PR #945, opened 2026-08-11
Add MSYS2 pacman packaging for MS-Windows

ikappaki:
Hi,

Could you please review the patch to enable support for creating the jank MSYS2 package? It addresses #797.

The package is built from the build tree using an inline `PKGBUILD` template in `package.clj`, via the standard `create-package!` dispatch mechanism for Windows.

The win CI build creates the package, which is then published using the same publishing mechanism as the Linux CI workflow to `root@cache.jank-lang.org:/var/www/cache.jank-lang.org/html/msys2`.

@jeaye, I couldn't test the full workflow end to end, since this requires publishing to `cache.jank-lang.org`, which I can't do without running the workflow on the official jank repo.

I was able to test everything up to generating the packages on CI, putting both LLVM and jank in the staging folder, and creating the jank repo database locally on CI.

To enable end to end testing without merging to `main` yet, I've made some temporary changes to the CI workflow:

1. Disabled the non windows builds.
2. Switched build for packaging from `Debug` to `Release`. Do we need a complementary non packaging build for `Debug` to match the workflow of the other archs?
3. Commented out the `if: github.ref == 'refs/heads/main'` guards so the workflow can run on a PR.
4. Created the repo database locally on CI using `add-repo`, instead of calling a new script on the VPS.
5. Chosen a provisional location on the VPS for the packages and database: `cache.jank-lang.org:/var/www/cache.jank-lang.org/html/msys2`, which I believe maps to `https://ppa.jank-lang.org/msys2` (as used in the installation docs).
6. Refrained from using `rsync --delete` for now, since I'm not confident enough about the VPS path and don't want to risk accidentally deleting unrelated files 😅. I think we should eventually use it, though, since each LLVM/jank build produces a package filename containing the source commit hash, e.g. `mingw-w64-clang-x86_64-jank-0.1-jank_3f918f1-any.pkg.tar.zst`. The old package would then be removed when a new build is published.
7. I haven't attempted to sign the package with the GPG key. Do we need this at this stage?

My idea for testing is:

* Enable the CI workflow on this PR.
* Let it publish the packages to the VPS.
* Try installing them locally from the VPS using `pacman`, following the installation instructions.
* If that works, restore the `if == main` guards and address the points above before merging to `main`.

What do you think? I may have got the VPS side wrong since I couldn't test it, but I've followed the publishing pattern used by the ubuntu/macos builds as closely as I could.

Thanks

An example CI run of the same PR on my fork, up to the point where it fails to connect to the VPS, is available at https://github.com/ikappaki/jank-win/actions/runs/31532089050/job/93917375570?pr=45

Example of the generated files (LLVM/jank packages + database) that need to be pushed to he VPS:

```drwxr-xr-x 1 runneradmin None        0 Aug 11 20:28 .
drwxr-xr-x 1 runneradmin None        0 Aug 11 20:28 ..
-rw-r--r-- 1 runneradmin None     2191 Aug 11 20:28 jank.db
-rw-r--r-- 1 runneradmin None     2191 Aug 11 20:28 jank.db.tar.gz
-rw-r--r-- 1 runneradmin None    55972 Aug 11 20:28 jank.files
-rw-r--r-- 1 runneradmin None    55972 Aug 11 20:28 jank.files.tar.gz
-rw-r--r-- 1 runneradmin None 49337211 Aug 11 20:27 mingw-w64-clang-x86_64-clang-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None   143723 Aug 11 20:27 mingw-w64-clang-x86_64-clang-analyzer-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 32547284 Aug 11 20:27 mingw-w64-clang-x86_64-clang-libs-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 26001393 Aug 11 20:27 mingw-w64-clang-x86_64-clang-tools-extra-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None  1330386 Aug 11 20:27 mingw-w64-clang-x86_64-compiler-rt-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None    63918 Aug 11 20:27 mingw-w64-clang-x86_64-gcc-compat-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 18374440 Aug 11 20:27 mingw-w64-clang-x86_64-jank-0.1-jank_3f918f1-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 15889371 Aug 11 20:27 mingw-w64-clang-x86_64-lld-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 92115254 Aug 11 20:27 mingw-w64-clang-x86_64-llvm-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 45680037 Aug 11 20:28 mingw-w64-clang-x86_64-llvm-libs-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 37433815 Aug 11 20:28 mingw-w64-clang-x86_64-llvm-tools-23.0.0git-1.jank_106644f6c-any.pkg.tar.zst
```


book/src/getting-started/01-installation.md
```diff
@@ -64,6 +64,29 @@ yay -S jank-bin
...
 If you'd like to install from source on Arch, you can install `jank-git` instead.
 
+## Windows (MSYS2)
+We have binary jank packages in our own pacman repo, so installation is quick and easy.
+
+First, install [MSYS2](https://www.msys2.org/) and open a **CLANG64** shell.
+
+Add our package repo to `/etc/pacman.conf` (e.g. with `nano /etc/pacman.conf`).
+Insert the following **before** the `[clang64]` section so that our bundled LLVM
+packages take priority over those in `[clang64]`:
```
jeaye, line 74: For every other OS, jank bundles its LLVM into the jank package so that it doesn't compete with the system LLVM. This is very intentional, since jank is often using an LLVM version which is not yet released. Also, LLVM breaks its APIs with every new release (every 6 months), which can then leave jank applications in a broken state on someone's system. The only way around this is to vendor our own LLVM.

Is there any practical reason that we can't follow this same practice for MSYS2? The jank CMake install setup is already prepared to bundle LLVM when `-Djank_local_clang=on` is set.

jeaye (commented):
Nice work, @ikappaki! This looks great. I have one concern, which I left as a comment.

book/src/getting-started/01-installation.md
ikappaki: Hi @jeaye,

When I was porting jank to MSYS2, I did try to use the same approach, but I couldn't get it to link with jank due to compilation errors, which I think were related to system libraries that I couldn't resolve at the time. I therefore switched to using the standard LLVM package build provided by MSYS2.

I've gained a lot more experience with MSYS2/LLVM since then, so it's possible I could now work through those compilation issues and get the same setup working, but it would require some effort to investigate.

That said, I don't think the current approach is particularly problematic. An MSYS2 installation is an isolated environment, and you can have multiple installations on the same computer. They don't interfere with each other or affect the system as a whole. A good analogy is a python virtual environment: installing LLVM in an MSYS2 environment only affects that environment, so it's unlikely that the jank LLVM installation would affect the system as a whole.

Furthermore, the installation instructions tell users to put the jank VPS repo first in their list of providers, so it will always prefer our LLVM package in that env. There shouldn't be any conflict with another LLVM installation.

I could perhaps make this clearer in the installation instructions by explicitly stating that our LLVM package will be preferred, or suggesting that users create a dedicated MSYS2 environment for compiling jank? These are cheap and easy to create, much like python virtual environments as I mentioned earlier.

Given that, I'm not sure it's currently worth going back and trying to make the LLVM installation work the same way as on linux/macos. I think my time would be better spent pushing through the LLVM structured exception changes with Lang?

Thanks

jeaye: Ok, gotcha. Thanks for the explanation. 

> Given that, I'm not sure it's currently worth going back and trying to make the LLVM installation work the same way as on linux/macos. I think my time would be better spent pushing through the LLVM structured exception changes with Lang?

Yes, this does seem like the highest priority for MSYS2.

book/src/getting-started/01-installation.md
```diff
@@ -64,6 +64,29 @@ yay -S jank-bin
...
+
+```conf
+[jank]
+SigLevel = Optional TrustAll
+Server = https://ppa.jank-lang.org/msys2
```
jeaye: We want to use `cache.jank-lang.org` for this. It resolves to the same VPS as `ppa.jank-lang.org`, but it's more generally used for caching artifacts. PPA is a Debian term.

jeaye (commented):
I left one comment, to replace `ppa` with `cache`. Aside from that, let's try to get this running on main. We can't test the publishing from a PR, I think, since the SSH key secret won't be available. We need to remove the commented out bits that run this on any branch before merging.

Thank you for the help!

book/src/getting-started/01-installation.md
ikappaki: switched to `cache.jank-lang.org`

ikappaki:
> I left one comment, to replace `ppa` with `cache`. Aside from that, let's try to get this running on main. We can't test the publishing from a PR, I think, since the SSH key secret won't be available. We need to remove the commented out bits that run this on any branch before merging.
> 
> Thank you for the help!

Hi @jeaye,

I've changed `ppa` to `cache` and enabled the `if` guards so that the jank package is created and published on the `main` branch.

There are two nuances I'd like to point out:

1. Package signing: the package is currently not signed, and the installation guide uses `SigLevel = Optional TrustAll`, which accepts unsigned packages from our `jank` repo. To enable package signing, we'd need a private PGP key (I assume you already have one that could be added as a workflow secret) and we'd need to publish the corresponding public key somewhere (on the jank-lang VPS?). Do we want package signing for alpha? I think it would be a good idea.

2. Package accumulation: both the LLVM and jank packages include a commit timestamp and hash in their filenames, so every build produces a new filename. This means the packages will accumulate on the VPS over time. The timestamp in the filename ensures that pacman picks up the latest version, which is what we want. I also like having the timestamp and hash in the filename: it makes the installed version immediately identifiable and makes the packages immutable.

The downside is that old jank packages will accumulate and consume disk space. This isn't much of a concern for LLVM, since new versions should be relatively infrequent.

A few options for housekeeping:

* (a) Have a housekeeping job on the VPS that keeps only the N newest jank packages.
* (b) Add `--delete` to the `rsync` command so the server contains only what we produced locally. I don't like this approach because a small configuration mistake could accidentally delete unrelated files from the server.
* (c) Remove the commit timestamp and hash from the jank package filename, and update the installation instructions to use a force reinstall when users want to update jank.

For reference, this is what the package directory currently looks like:

```
-rw-r--r-- 1 runneradmin None     2205 Sep  7 05:29 jank.db
-rw-r--r-- 1 runneradmin None     2205 Sep  7 05:29 jank.db.tar.gz
-rw-r--r-- 1 runneradmin None    56075 Sep  7 05:29 jank.files
-rw-r--r-- 1 runneradmin None    56075 Sep  7 05:29 jank.files.tar.gz
-rw-r--r-- 1 runneradmin None 49353432 Sep  7 05:29 mingw-w64-clang-x86_64-clang-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None   143719 Sep  7 05:29 mingw-w64-clang-x86_64-clang-analyzer-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 32550446 Sep  7 05:29 mingw-w64-clang-x86_64-clang-libs-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 26001776 Sep  7 05:29 mingw-w64-clang-x86_64-clang-tools-extra-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None  1329764 Sep  7 05:29 mingw-w64-clang-x86_64-compiler-rt-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None    63235 Sep  7 05:29 mingw-w64-clang-x86_64-gcc-compat-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 17199551 Sep  7 05:29 mingw-w64-clang-x86_64-jank-0.1-1788752223.1323629-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 15890426 Sep  7 05:29 mingw-w64-clang-x86_64-lld-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 92095068 Sep  7 05:29 mingw-w64-clang-x86_64-llvm-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 45665106 Sep  7 05:29 mingw-w64-clang-x86_64-llvm-libs-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
-rw-r--r-- 1 runneradmin None 37396353 Sep  7 05:29 mingw-w64-clang-x86_64-llvm-tools-23.0.0git-1.jank_1782523676.106644f6c-any.pkg.tar.zst
```

Let me know what you think.

thanks

jeaye:
Hi @ikappaki! I have not yet looked at this, but I have every intention of getting it merged. I just need to get some time to set aside to sort out the two nuances you set up, since I want to ensure these are signed and that we don't run out of space on the VPS. However, this month I'm swamped with preparation for the Conj, so I don't know when I'll be getting to this. Worst case, I'll get it merged after the Conj.

I'm sorry for taking this long to review and I appreciate your patience.

ikappaki:
No worries at all! Thanks for the update, and good luck with the Conj!

PR #946, opened 2026-08-12
Initial array implementation

djblue:
Not everything is complete, but lots of things are working. I did implement visit for arrays to support some of the Clojure functions that still use it.

One of the main things that hasn't been completed is how arrays should be treated as seqs. Should we have another wrapper type for when we want to treat an array only as a seq, or do we use the same array type? If so, we would also need to track the head index as well, curious about your thoughts on this.

Also, what should the canonical_type for the first class arrays be? I took a guess (bool/f32/i32/...), but not sure if this is correct.

I also threw in a few smaller fixes to enable more tests.

And, I think because of inlining, primitive arrays still work with the new versions of `aget`/`aset`:

```clojure
% jank repl
user=> (let [a (#cpp (:array bool 256))] (aset a 1 true) [(aget a 0) (aget a 1)])
[false true]
```


compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
```diff
@@ -4,15 +4,15 @@
...
+    clojure.core-test.aclone
     ; clojure.core-test.add-watch ; TODO: port sync
     ; clojure.core-test.ancestors ; analyze/unresolved-symbol error: Unable to resolve symbol 'Object'.
     clojure.core-test.and
     clojure.core-test.any-qmark
     clojure.core-test.apply
     clojure.core-test.assoc
     clojure.core-test.assoc-bang
-    ; clojure.core-test.associative-qmark ; TODO: port to-array
+    ; clojure.core-test.associative-qmark ; FIXME: There is no 'seq' member function within 'const char[3]'.
```
djblue, line 15: I think this is a new bug due to primitive literals and inling the method call for `.seq`.

compiler+runtime/include/cpp/jank/runtime/core/array.hpp
```diff
@@ -0,0 +1,136 @@
...
+      };
+    }
+
+    return a->data[i];
+  }
+
+  template <typename T, size_t N>
+  T aget(T const (&a)[N], i64 const i)
+  {
+    return a[i];
```
jeaye, line 82: Do we want a bounds check here?

compiler+runtime/include/cpp/jank/runtime/core/array.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+    a->data[i] = val;
+
+    return val;
+  }
+
+  template <typename T, size_t N>
+  T aset(T (&a)[N], i64 const i, T const val)
+  {
+    a[i] = val;
```
jeaye, line 144: Do we want a bounds check here?

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,177 @@
...
+#include <jank/runtime/obj/opaque_box.hpp>
+
+namespace jank::runtime::obj
+{
+  template <typename T>
+  struct array : object
+  {
+    static constexpr object_type obj_type{ object_type::array };
+    static constexpr object_behavior obj_behaviors{ object_behavior::get | object_behavior::seqable
+                                                    | object_behavior::sequence_like
```
jeaye: This should be seqable, but not sequence_like. We need a separate array sequence type. We can probably repurpose the `native_array_sequence` object for this.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,177 @@
...
+      , size{ size }
+      , data{ new(UseGC) T[size] }
+    {
+      for(usize i{}; i < size; ++i)
+      {
+        data[i] = init_value;
+      }
+    }
+
+    T &operator[](std::size_t const i)
```
jeaye: Please use `usize` instead of `std::size_t`.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,177 @@
...
+      {
+        throw std::runtime_error{
+          util::format("out of bounds index {}; array has a size of {}", i, size)
+        };
+      }
+
+      return data[i];
+    }
+
+    T at(std::size_t i) const
```
jeaye: Normally `at` does the throwing and `op[]` just does blind access. If we have `op[]` throwing, we really don't need an `at`.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,177 @@
...
+    object_ref nth(object_ref const index) const override
+    {
+      auto const i(static_cast<usize>(index.to_integer()));
+      if(i >= size)
+      {
+        throw std::runtime_error{
+          util::format("out of bounds index {}; array has a size of {}", i, size)
+        };
+      }
+      return make_box(data[i]);
```
jeaye: I think we want `convert<T>::into_object(data[i])` here and in the other `make_box` uses below.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /* behavior::countable */
+    usize count() const
+    {
+      return size;
+    }
+
+    jtl::immutable_string canonical_type{};
+    usize size;
```
jeaye: Please default initialize this with `{}`.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,177 @@
...
+    static constexpr object_behavior obj_behaviors{ object_behavior::get | object_behavior::seqable
+                                                    | object_behavior::sequence_like
+                                                    | object_behavior::indexable };
+    static constexpr bool pointer_free{ false };
+
+    array(jtl::immutable_string const &canonical_type, usize const size)
+      : object{ obj_type, obj_behaviors }
+      , canonical_type{ canonical_type }
+      , size{ size }
+      , data{ new(UseGC) T[size] }
```
jeaye: Please default initialize this with `T[size]{}`.

compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
```diff
@@ -699,6 +700,64 @@ namespace jank::analyze::cpp_util
         }
       case jank::runtime::object_type::var:
         return var_type();
+      case jank::runtime::object_type::array:
+        {
+          auto const type(get_element_type(o));
```
jeaye: Arrays can hold any type. We cannot use a hard-coded list here. We need to construct the actual array type here.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
+#include <jank/runtime/core/array.hpp>
+
+#include <jank/util/fmt.hpp>
+
+#include <jank/runtime/sequence_range.hpp>
```
jeaye: These should be logically grouped together, with util at the end.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+#include <jank/runtime/core/seq.hpp>
+
+namespace jank::runtime
+{
+  jtl::immutable_string get_element_type(object_ref const o)
+  {
+    return try_object<obj::array<int>>(o)->canonical_type;
```
jeaye: This is undefined behavior, since you're just assuming that this would be an array of ints. Thus, you're aliasing some arbitrary array type over some other arbitrary array type. We need a virtual fn for this instead.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  obj::bool_array_ref boolean_array(u64 const size, bool const init_value)
+  {
+    return make_box<obj::array<bool>>("bool", size, init_value);
+  }
+
+  obj::bool_array_ref boolean_array(object_ref const o)
+  {
+    if(o.get_type() == object_type::small_integer || o.get_type() == object_type::integer)
```
jeaye: We have an `is_integer` predicate for this.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+    return a;
+  }
+
+  obj::f32_array_ref float_array(u64 const size)
+  {
+    return make_box<obj::array<f32>>("f32", size);
+  }
+
+  obj::f32_array_ref float_array(u64 const size, f32 const init)
```
jeaye: All of these `u64` uses should be `usize`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+  {
+    return make_box<obj::array<f32>>("f32", size, init);
+  }
+
+  obj::f32_array_ref float_array(object_ref const o)
+  {
+    if(o.get_type() == object_type::small_integer || o.get_type() == object_type::integer)
```
jeaye: `is_integer`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  obj::f64_array_ref double_array(u64 const size, f64 const init)
+  {
+    return make_box<obj::array<f64>>("f64", size, init);
+  }
+
+  obj::f64_array_ref double_array(object_ref const o)
+  {
+    if(o.get_type() == object_type::small_integer || o.get_type() == object_type::integer)
```
jeaye: `is_integer`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  obj::i16_array_ref short_array(u64 const size, i16 const init)
+  {
+    return make_box<obj::array<i16>>("i16", size, init);
+  }
+
+  obj::i16_array_ref short_array(object_ref const o)
+  {
+    if(o.get_type() == object_type::small_integer || o.get_type() == object_type::integer)
```
jeaye: `is_integer`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  obj::i32_array_ref int_array(u64 const size, i32 const init)
+  {
+    return make_box<obj::array<i32>>("i32", size, init);
+  }
+
+  obj::i32_array_ref int_array(object_ref const o)
+  {
+    if(o.get_type() == object_type::small_integer || o.get_type() == object_type::integer)
```
jeaye: `is_integer`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  obj::i64_array_ref long_array(u64 const size, i64 const init)
+  {
+    return make_box<obj::array<i64>>("i64", size, init);
+  }
+
+  obj::i64_array_ref long_array(object_ref const o)
+  {
+    if(o.get_type() == object_type::small_integer || o.get_type() == object_type::integer)
```
jeaye: `is_integer`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  obj::object_array_ref object_array(u64 const size)
+  {
+    return make_box<obj::array<object_ref>>("object", size);
+  }
+
+  obj::object_array_ref object_array(object_ref const o)
+  {
+    if(o.get_type() == object_type::small_integer || o.get_type() == object_type::integer)
```
jeaye: `is_integer`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  obj::u8_array_ref byte_array(u64 const size, u8 const init)
+  {
+    return make_box<obj::array<u8>>("u8", size, init);
+  }
+
+  obj::u8_array_ref byte_array(object_ref const o)
+  {
+    if(o.get_type() == object_type::small_integer || o.get_type() == object_type::integer)
```
jeaye: `is_integer`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,478 @@
...
+      ++i;
+    }
+
+    return a;
+  }
+
+  object_ref aget(object_ref const a, i64 const i)
+  {
+    auto const type(get_element_type(a));
+    if(type == "bool")
```
jeaye: We can't just have a hard-coded list of supported types. Clojure arrays support any element type. jank's should too.

jeaye (changes requested):
Thanks, Chris! The big thing is that I believe the array object needs to be able to support any type. Right now, we have a hard-coded list of assumed types instead.

Perhaps we can discuss this more so I can understand how this can by viable. For example, do you imagine that anyone who wants an array of some custom type (like `my_game::vec3`) needs to use a native array instead of an array object?

compiler+runtime/include/cpp/jank/runtime/core/array.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+    if(a->canonical_type != canonical_type)
+    {
+      throw std::runtime_error{ util::format(
+        "Array cast failed, 'canonical_type' mismatch: expected '{}', received '{}'.",
+        canonical_type,
+        a->canonical_type) };
+    }
+
+    return expect_object<obj::array<T>>(o);
```
jeaye: You can just `return a;` here, right?

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,177 @@
...
+      return {};
+    }
+
+    /* behavior::countable */
+    usize count() const
+    {
+      return size;
+    }
+
+    jtl::immutable_string canonical_type{};
```
jeaye: If arrays can only hold a hard-coded set of types, this should be an enum, not a string. We're in between two worlds here.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -6621,139 +6625,166 @@
...
   "Return true if x is a byte array"
   [x]
-  ;; (if (nil? x)
-  ;;   false
-  ;;   (-> x class .getComponentType (= Byte/TYPE)))
-  (throw "TODO: port bytes?"))
+  (if (nil? x)
+    false
+    (and (cpp/== (.get_type x)
+                 cpp/jank.runtime.object_type.array))))
```
jeaye: This is just checking the the object is an array, not that it's an array of bytes.

compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
jeaye: Yeah, we can't inline to member fn calls. I've made that mistake a couple of times.

djblue:
> Thanks, Chris! The big thing is that I believe the array object needs to be able to support any type. Right now, we have a hard-coded list of assumed types instead.
> 
> Perhaps we can discuss this more so I can understand how this can by viable. For example, do you imagine that anyone who wants an array of some custom type (like `my_game::vec3`) needs to use a native array instead of an array object?

So I think the main issue here, as far as I know, is that the JVM currently doesn't support storing structs inline for arrays. Arrays store primitives or object references. And since the JVM doesn't support this, neither does Clojure. There are plans to enable default support for "Value Objects" which would allow for inline storage in arrays, but that is currently a preview https://openjdk.org/jeps/401. Also they have the following non-goal: `It is not a goal to guarantee any particular optimization strategy or memory layout.`

I do think there is room here for jank to have better array support for arbitrary types, but I don't know what that should look like, or if it would be provided via `clojure.core`.

jeaye:
> Arrays store primitives or object references. And since the JVM doesn't support this, neither does Clojure.

Oh, I hadn't realized this. Thanks for spelling it out for me.

Ok, so maybe our array object can be simpler like this. I think we'll want to put a `requires` constraint on the array template to ensure that it's either constructed with a primitive or `object_ref` or `object_like<oref<T>::value_type>`.

jeaye:
Ok, following up on this after a bit more research. Looks like Clojure JVM does indeed support arrays of arbitrary **class types**. For example:

```clojure
user=> (def ar (make-array Thread 3))
#'user/ar

user=> (pprint ar)
[nil, nil, nil]

user=> (type ar)
[Ljava.lang.Thread;
```

So that's not just an array of `Object`. It's an array of specifically typed objects. Since every class is an object, that means you can have an array of effectively any type.

djblue:
> Ok, following up on this after a bit more research. Looks like Clojure JVM does indeed support arrays of arbitrary **class types**. For example:
> 
> ```clojure
> user=> (def ar (make-array Thread 3))
> #'user/ar
> 
> user=> (pprint ar)
> [nil, nil, nil]
> 
> user=> (type ar)
> [Ljava.lang.Thread;
> ```
> 
> So that's not just an array of `Object`. It's an array of specifically typed objects. Since every class is an object, that means you can have an array of effectively any type.

For jank, that would mean `array<object_ref>` or `array<obj::thread_ref>` then, no? So arrays would be limited in the same way object_ref is limited.

djblue:
This PR is ready for another round of review. I still need to update error messages to fit our new style guide, but I will address those after the PR is confirmed to be technically sound.

compiler+runtime/include/cpp/jank/runtime/core/array.hpp
```diff
@@ -0,0 +1,258 @@
...
+  obj::array_ref<T> try_array(object_ref const o)
+  {
+    auto const a(try_object<obj::array<T>>(o));
+    auto const expected_element_type(obj::get_array_element_type<T>());
+    auto const actual_element_type(o.get_element_type());
+
+    if(actual_element_type != expected_element_type)
+    {
+      throw std::runtime_error{ util::format(
+        "Array cast failed, 'element_type' mismatch: expected '{}', received '{}'.",
```
jeaye: I know you said the error messages need another pass, so I'll just leave this one comment about all of them.
 
1. Use complete sentences.
2. Use back ticks instead of single quotes.

Try to write a sentence like you're describing the problem to someone using spoken English.

compiler+runtime/include/cpp/jank/runtime/core/array.hpp
```diff
@@ -0,0 +1,258 @@
...
+  requires(!jtl::is_same<T, char> && obj::is_array_element_type<T>)
+  struct convert<T[N]>
+  {
+    static obj::array_ref<T> into_object(T (&a)[N])
+    {
+      auto const clone(make_box<obj::array<T>>(N));
+      std::copy(a, a + N, clone->data.data);
+      return clone;
+    }
+  };
```
jeaye, line 206: This belongs in `builtin.hpp`.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+    {
+      return array_element_type::f64;
+    }
+    else if constexpr(jtl::is_same<T, object_ref>)
+    {
+      return array_element_type::object;
+    }
+  }
+
+  template <typename T>
```
jeaye, line 78: We should have a `requires is_array_element_type<T>` on here.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+  {
+    static constexpr object_type obj_type{ object_type::array };
+    static constexpr object_behavior obj_behaviors{ object_behavior::get | object_behavior::seqable
+                                                    | object_behavior::indexable
+                                                    | object_behavior::array_like };
+    static constexpr bool pointer_free{ false };
+
+    array(usize const size)
+      : object{ obj_type, obj_behaviors }
+      , element_type{ get_array_element_type<T>() }
```
jeaye: These members which always get the same value should be initialized in their definitions, not duplicated into each ctor.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+    jtl::immutable_string to_code_string() const override
+    {
+      jtl::string_builder buff;
+      util::format_to(buff, "#array [{} {} {}]", element_type_str(element_type), size, this);
+      return buff.release();
+    }
+
+    object_ref aset(i64 const index, object_ref const val) override
+    {
+      auto const i(static_cast<usize>(index));
```
jeaye: This is a problem, since we're not checking if `index` is negative before casting to an unsigned type. Just have `op[]` use `i64` so we don't need to worry about this.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+    {
+      if(i >= size)
+      {
+        throw std::runtime_error{
+          util::format("out of bounds index {}; array has a size of {}", i, size)
+        };
+      }
+
+      return data[i];
+    }
```
jeaye: This shouldn't be needed.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+    /* behavior::indexable */
+    object_ref nth(object_ref const index) const override
+    {
+      auto const i(static_cast<usize>(index.to_integer()));
```
jeaye: Not a safe cast, since the integer could be negative.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+        throw std::runtime_error{
+          util::format("out of bounds index {}; array has a size of {}", i, size)
+        };
+      }
+      return convert<T>::into_object(data[i]);
+    }
+
+    object_ref nth(object_ref const index, object_ref const fallback) const override
+    {
+      auto const i(static_cast<usize>(index.to_integer()));
```
jeaye: Not a safe cast.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+      }
+      return convert<T>::into_object(data[i]);
+    }
+
+    /* behavior::get */
+    using object::get;
+
+    object_ref get(object_ref const key) const override
+    {
+      auto const i(static_cast<usize>(key.to_integer()));
```
jeaye: Not a safe cast.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+      if(i >= size)
+      {
+        return {};
+      }
+      return convert<T>::into_object(data[i]);
+    }
+
+    object_ref get(object_ref const key, object_ref const fallback) const override
+    {
+      auto const i(static_cast<usize>(key.to_integer()));
```
jeaye: Not a safe cast.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,269 @@
...
+      if(i >= size)
+      {
+        return fallback;
+      }
+      return convert<T>::into_object(data[i]);
+    }
+
+    bool contains(object_ref const key) const override
+    {
+      auto const i(static_cast<usize>(key.to_integer()));
```
jeaye: Not a safe cast.

compiler+runtime/include/cpp/jank/runtime/oref.hpp
```diff
@@ -399,6 +399,21 @@ namespace jank::runtime
...
+
+    object_ref aset(i64 const index, object_ref const val)
+    {
+      return ptr()->aset(index, val);
+    }
+
+    object_ref aclone() const
+    {
+      return ptr()->aclone();
+    }
```
jeaye: 1. This is not added in the correct place.
2. It needs a `behavior::array_like` comment section above it.
3. Be sure to follow the existing patterns in the other behavior facade fns so we can handle small ints and reals properly.
4. These facades need to be added to ALL oref specializations, which is four total. We're missing half of them and the two that we do have are not correct.

compiler+runtime/include/cpp/jank/runtime/visit.hpp
```diff
@@ -244,6 +246,45 @@ namespace jank::runtime
         return fn(expect_object<obj::inst>(erased), std::forward<Args>(args)...);
       case object_type::opaque_box:
         return fn(expect_object<obj::opaque_box>(erased), std::forward<Args>(args)...);
+      case object_type::array:
+        /* TODO: Remove arrays from visit. It is still required for certain
+         * clojure functions such as nth. */
```
jeaye: There is no TODO here anymore. visiting arrays is totally fine, since they have a closed set of types. Arrays are no different than any other object type now, in terms of visiting.

compiler+runtime/include/cpp/jank/runtime/visit.hpp
```diff
@@ -244,6 +246,45 @@ namespace jank::runtime
...
+            case array_element_type::f32:
+              return fn(expect_object<obj::array<f32>>(erased), std::forward<Args>(args)...);
+            case array_element_type::f64:
+              return fn(expect_object<obj::array<f64>>(erased), std::forward<Args>(args)...);
+            case array_element_type::object:
+              return fn(expect_object<obj::array<object_ref>>(erased), std::forward<Args>(args)...);
+            default:
+              throw std::runtime_error{ util::format(
+                "visit not supported for 'array' with element type: '{}'",
+                element_type_str(type)) };
```
jeaye: This needs to be an internal error, since this should never happen.

compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
```diff
@@ -704,6 +705,91 @@ namespace jank::analyze::cpp_util
         }
       case jank::runtime::object_type::var:
         return var_type();
+      case jank::runtime::object_type::array:
+        {
+          auto const type(o.get_element_type());
+          switch(type)
```
jeaye, line 738: We need to handle the `default` case with an internal error.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  obj::bool_array_ref boolean_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return boolean_array(o.to_integer());
+    }
+
+    auto a(boolean_array(sequence_length(o)));
```
jeaye: This can be `const`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+  obj::char_array_ref char_array(usize const size, char const init)
+  {
+    return make_box<obj::array<char>>(size, init);
+  }
+
+  obj::char_array_ref char_array(jtl::immutable_string const &s)
+  {
+    auto const size(s.size());
+    auto a(char_array(size));
```
jeaye: This can be `const`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+    for(usize i{}; i < size; ++i)
```
jeaye: Prefer `std::copy`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  obj::f32_array_ref float_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return float_array(o.to_integer());
+    }
+
+    auto a(float_array(sequence_length(o)));
```
jeaye: This can be `const`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  obj::f64_array_ref double_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return double_array(o.to_integer());
+    }
+
+    auto a(double_array(sequence_length(o)));
```
jeaye: This can be `const`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  obj::i16_array_ref short_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return short_array(o.to_integer());
+    }
+
+    auto a(short_array(sequence_length(o)));
```
jeaye: This can be `const`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  obj::i32_array_ref int_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return int_array(o.to_integer());
+    }
+
+    auto a(int_array(sequence_length(o)));
```
jeaye: This can be `const`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  obj::i64_array_ref long_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return long_array(o.to_integer());
+    }
+
+    auto a(long_array(sequence_length(o)));
```
jeaye: This can be `const`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  obj::object_array_ref object_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return object_array(o.to_integer());
+    }
+
+    auto a(object_array(sequence_length(o)));
```
jeaye: This can be `const`.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  obj::u8_array_ref byte_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return byte_array(o.to_integer());
+    }
+
+    auto a(byte_array(sequence_length(o)));
```
jeaye: This can be `const`.

compiler+runtime/src/jank/clojure/core.jank
```diff
@@ -6018,37 +6018,41 @@
 (defn alength
```
jeaye, line 6018: We should be able to do `amap` and `areduce` now, too.

compiler+runtime/test/bash/clojure-test-suite/src/jank_test/run_clojure_test_suite.cljc
jeaye: I have since fixed the `.seq` inlining.

jeaye (changes requested):
Nice job, Chris! Overall, I see no issues with this approach. We just need to clean it up, fix the error messages, and fill in any remaining bits.

compiler+runtime/include/cpp/jank/runtime/core/array.hpp
djblue: Moving this into `builtin.hpp` causes a circular dependency that I'm an unsure of how to resolve.

jeaye: Ok, please leave a TODO above it and I will handle it.

compiler+runtime/include/cpp/jank/runtime/visit.hpp
djblue: How do I signal an internal error, `jtl::panic`?

jeaye: No, `panic` is for unrecoverable errors which need the program to abort.

For this, throw a `error::runtime_internal_failure(util::format("..."))`.

compiler+runtime/include/cpp/jank/runtime/core/array.hpp
```diff
@@ -0,0 +1,245 @@
...
+  obj::array_ref<T> try_array(object_ref const o)
+  {
+    auto const a(try_object<obj::array<T>>(o));
+    auto const expected_element_type(obj::get_array_element_type<T>());
+    auto const actual_element_type(o.get_element_type());
+
+    if(actual_element_type != expected_element_type)
+    {
+      throw std::runtime_error{ util::format(
+        "Array with element type of `{}` cannot be cast to array of element type `{}`.",
```
jeaye, line 24: This is not valid English grammar. Suggestion:

```cpp
"An array of `{}` was needed here, but an array of `{}` was provided."
```

compiler+runtime/include/cpp/jank/runtime/core/array.hpp
```diff
@@ -0,0 +1,245 @@
...
+  extern template usize alength(obj::i8_array_ref const a);
+  extern template usize alength(obj::u16_array_ref const a);
+  extern template usize alength(obj::i16_array_ref const a);
+  extern template usize alength(obj::u32_array_ref const a);
+  extern template usize alength(obj::i32_array_ref const a);
+  extern template usize alength(obj::u64_array_ref const a);
+  extern template usize alength(obj::i64_array_ref const a);
+  extern template usize alength(obj::f32_array_ref const a);
+  extern template usize alength(obj::f64_array_ref const a);
+  extern template usize alength(obj::object_array_ref const a);
```
jeaye, line 159: If we're just going to specify an overload for every single type, the template here doesn't buy us anything. Let's just remove the template and have a bunch of overloads defined in the `cpp` file.

As you can see in this godbolt, having this template doesn't actually allow any inlining, since all of the definitions are externed anyway. https://godbolt.org/z/b6sPvY8Ez

Might as well just use simple functions. The same can be said for `aget`, `aset`, `alength`, `aclone`. We cannot do this for `try_array`, since it doesn't take a differently typed parameter.

To avoid the duplication of written code in `aclone`, we should just use a macro. For example, https://godbolt.org/z/GdePWEsoh

jeaye: The reason why all of this matters is that we need to keep our headers as small as possible. This is adding a lot of cruft and we're not even getting inlining (perf) benefits from it.

compiler+runtime/include/cpp/jank/runtime/obj/array.hpp
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    object_ref aclone() const override
+    {
+      auto const clone(make_box<array<T>>(size));
+      std::copy(data.data, data.data + size, clone->data.data);
+      return clone;
+    }
+
+    T &operator[](i64 const i) const
```
jeaye, line 131: Having a `const` member return a non-const reference to its data is poor etiquette. This member should be non-const.

compiler+runtime/include/cpp/jank/runtime/visit.hpp
```diff
@@ -244,6 +247,43 @@ namespace jank::runtime
...
+            case array_element_type::f32:
+              return fn(expect_object<obj::array<f32>>(erased), std::forward<Args>(args)...);
+            case array_element_type::f64:
+              return fn(expect_object<obj::array<f64>>(erased), std::forward<Args>(args)...);
+            case array_element_type::object:
+              return fn(expect_object<obj::array<object_ref>>(erased), std::forward<Args>(args)...);
+            default:
+              throw error::runtime_internal_failure(
+                util::format("Visit is not supported for `array` with element type of `{}`.",
+                             element_type_str(type)));
```
jeaye, line 284: We know this is not a supported type, which is why this is an internal error. The `element_type_str` is always going to give `unknown`, so it's not helpful here. We should just render the actual value. Also, the error message should say that this element type is unexpected. It's not that it's a valid array element type, we just can't visit it. This is clearly a bogus array element type.

compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
```diff
@@ -732,6 +732,95 @@ namespace jank::analyze::cpp_util
...
+            case runtime::array_element_type::object:
+              {
+                static auto const type{ Cpp::GetTypeFromScope(
+                  resolve_scope("jank.runtime.obj.object_array_ref").expect_ok()) };
+                return type;
+              }
+            default:
+              throw error::codegen_internal_failure(util::format(
+                "Unable to resolve literal_type for `array` with element type of `{}`.",
+                element_type_str(type)));
```
jeaye, line 821: Same thing here. Let's render the actual value, since `element_type_str` isn't helpful when the value is unexpected.

compiler+runtime/src/cpp/jank/runtime/core/array.cpp
```diff
@@ -0,0 +1,433 @@
...
+  obj::bool_array_ref boolean_array(usize const size, bool const init_value)
+  {
+    return make_box<obj::array<bool>>(size, init_value);
+  }
+
+  obj::bool_array_ref boolean_array(object_ref const o)
+  {
+    if(is_integer(o))
+    {
+      return boolean_array(o.to_integer());
```
jeaye, line 38: This pattern is used a lot, in this file. We're converting from a `i64` into a `usize`, which is going to explode for negative values. We should just have the `array` ctor take an `i64` and do a sign check on it in that one spot, rather than having a dozen different footguns in this file.

jeaye (changes requested):
Thank you, Chris! I think we have some signed/unsigned and header bloat concerns. Aside from that, this is looking great.

PR #955, opened 2026-08-20
Add type hinting support

jeaye:


jeaye:
The macOS failures in this PR are from an LLVM bug. I have been chasing it down today and I have a working patch, which is now up as a PR: https://github.com/llvm/llvm-project/pull/217781

I'm not sure if it's any good, so we'll need to hold off and see what the LLVM folks say. I'm going to leave this PR open for now, until I have a clearer idea of the timeline.

PR #964, opened 2026-08-29
Update folly, remove ::close workarounds, handle excpt.h

ikappaki:
Hi, could you please consider this patch to prevent Folly's `unistd` compatibility namespace from being promoted to the global namespace? It resolves #962.

MSYS2 provides a `unistd.h` API that conflicts with Folly's compatibility API when transitively included. The fix allows the MSYS2 API to be used directly, avoiding these conflicts.

The patch also improves the existing `exception_info` macro workaround by explicitly including MSYS2's `excpt.h` before undefining the macro, making it independent of transitive include order.

The Folly submodule currently points to my fork until https://github.com/jank-lang/folly/pull/3 is merged upstream, after which we'll switch to the upstream commit here before merging.

Thanks!

### DIFF
--- a/compiler+runtime/include/cpp/jank/runtime/obj/exception_info.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/exception_info.hpp
@@ -1,8 +1,13 @@
 #pragma once
 
-/* Windows defines this and it becomes an issue in unity builds. */
-#ifdef exception_info
-  #undef exception_info
+/* MSYS2's excpt.h is included transitively and defines exception_info
+ * as a macro, which conflicts with our class. Undefine it here to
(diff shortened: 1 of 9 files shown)


jeaye (dismissed):
This one just needs a repoint of folly and then we're good to merge! Thank you, @ikappaki! Your solo effort is keeping the MSYS2 build alive and I appreciate your support.

ikappaki:
> This one just needs a repoint of folly and then we're good to merge! Thank you, @ikappaki! Your solo effort is keeping the MSYS2 build alive and I appreciate your support.

Switched over to folly's main branch. I'm glad I can help! I wish I had more time to spend on the project.

jeaye (approved):

PR #965, opened 2026-08-29
Sync with Clojure test suite

shantanu-sardesai:
> [!important]
> Requires [a re-point](https://github.com/jank-lang/clojure-test-suite/pull/954).

Brings the count to **189 tests**.


jeaye (approved):
Nice!

PR #968, opened 2026-09-01
Bump LLVM commit to latest main

jeaye:
This includes two jank-related fixes:

- https://github.com/llvm/llvm-project/pull/217781
- https://github.com/llvm/llvm-project/pull/217147


jeaye:
Holding off on this while I try to get these two fixes into LLVM 23, rather than from main.

PR #977, opened 2026-09-05
Enable strip & linker garbage collection for AOT

kylc:
I don't see any downside to this, and it gives a modest size reduction for statically linked binaries. Sort of like a poor man's LTO. Normally you would combine with `-fdata-sections -ffunction-sections` but I didn't see any change with those flags.

Here are the results for a jank hello world program:

|       | -O2  | -O2 -s | -O2 -s & upx --best |
|-------|------|--------|--------------|
| main  | 8.4M | 6.6M   | 1.9M         |
| w/ PR | 7.4M | 5.9M   | 1.4M         |

(`-s` was added via `JANK_EXTRA_FLAGS`)

### DIFF
--- a/compiler+runtime/src/cpp/jank/aot/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/aot/processor.cpp
@@ -299,6 +299,8 @@ int main(int argc, const char** argv)
       {
         linker_args.push_back(strdup(lib));
       }
+
+      linker_args.push_back(strdup("-Wl,--gc-sections"));
     }
     else
     {
--- a/compiler+runtime/src/cpp/jank/aot/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/aot/processor.cpp
@@ -300,7 +300,14 @@ int main(int argc, const char** argv)
         linker_args.push_back(strdup(lib));
       }
 
-      linker_args.push_back(strdup("-Wl,--gc-sections"));
+      if constexpr(jtl::current_platform == jtl::platform::macos_like)
+      {
+        linker_args.push_back(strdup("-Wl,-dead_strip"));
+      }
+      else
+      {
+        linker_args.push_back(strdup("-Wl,--gc-sections"));
+      }
     }
     else
     {
--- a/compiler+runtime/src/cpp/jank/aot/processor.cpp
+++ b/compiler+runtime/src/cpp/jank/aot/processor.cpp
@@ -248,8 +248,15 @@ int main(int argc, const char** argv)
       compiler_args.emplace_back(strdup(framework.c_str()));
     }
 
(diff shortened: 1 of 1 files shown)


jeaye:
Works for me! What do you think about adding `-s` if `util::cli::opts.debug` is false?

kylc:
Maybe, I'm not sure. Do you think it's OK for stack traces to be basically useless if debug is false? After stripping they won't contain any information.

Since `--no-debug` is opt-in (I think?) that might be alright.

jeaye:
> Maybe, I'm not sure. Do you think it's OK for stack traces to be basically useless if debug is false? After stripping they won't contain any information.
> 
> Since `--no-debug` is opt-in (I think?) that might be alright.

Yeah, if we leave it opt-in, I think this doesn't harm anything. If we cleverly decide to put `--no-debug` on release builds... we'd better make sure we're shipping that debug info in a separate file so people can later symbolicate their stack traces.

kylc:
Added stripping in the latest commit. `-Wl,--strip-all` is just the long-form `-s` from what I can tell.

```
$ jank compile -O2 --module-path . hello --name hello && du -hc target/hello
7.4M	target/hello
7.4M	total
$ ./target/hello
Hello from jank
error: a
#0  in hello./main (arity 0) at hello.jank:5
#1  in jank::runtime::apply_to at hello
#2  (inlined)          in main::$_0::operator() at jank-main-8GZXpE:44
#3  in main::$_0::__invoke at jank-main-8GZXpE:33
#4  in jank_init_static at hello

$ jank compile -O2 --module-path . hello --name hello --no-debug && du -hc target/hello
5.9M	target/hello
5.9M	total
$ ./target/hello
Hello from jank
error: a
#0 at hello
#1 at hello
#2 at hello
#3 at hello
#6 at hello
```

jeaye (approved):
Looks great!

jeaye:
This looks very odd:

```
#0  in hello./main (arity 0) at hello.jank:5
```

I wonder why there's a `.` after `hello` there.

kylc:
Here is the source file if you want to debug:

```
(ns hello)

(defn -main []
  (println "Hello from jank")
  (throw (cpp/std.runtime_error "a")))
```

Maybe because it's a single-part namespace (not `hello.a.b.c`)?

jeaye:
> Here is the source file if you want to debug:
> 
> ```
> (ns hello)
> 
> (defn -main []
>   (println "Hello from jank")
>   (throw (cpp/std.runtime_error "a")))
> ```
> 
> Maybe because it's a single-part namespace (not `hello.a.b.c`)?

Thanks! I've made a note with this source so I can dig into it.

If you're interested in adding upx to lein-jank, I'd be happy to merge that as well! Seems like something we could keep on by default for a `release-lto` profile.

PR #978, opened 2026-09-05
Add a `lein debug` helper alias

kylc:
This helper drops you into gdb with all the proper jank arguments for your project.

I'm not sure what the implications are for Windows/MacOS, but at least for now it will just say that gdb is not available. Maybe since we ship with LLVM we can use a bundled lldb? Not sure.

Alternatively this could be an argument to `lein run` like `lein run --debug`, but this seems more discoverable.

### DIFF
--- a/lein-jank/src/leiningen/jank.clj
+++ b/lein-jank/src/leiningen/jank.clj
@@ -29,8 +29,10 @@
   (find-deprecated-flags project)
   (when (:verbose opts)
     (reset! ljc/verbose? true))
+  (when (:debug opts)
+    (ljc/verify-executable! ljc/debug-tool))
   (let [project (ljc/native-build project opts)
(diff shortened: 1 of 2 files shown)


jeaye:
Thanks, Kyle! I like having `lein debug` versus having a flag.

On macOS, we can default to `lldb --` instead of `gdb --args` and the rest should just work. Is that something you're open to adding to this, or shall we start with just gdb?

kylc:
> Thanks, Kyle! I like having lein debug versus having a flag.

Well now we can do both, which also lets us reuse some code :P

> On macOS, we can default to lldb -- instead of gdb --args and the rest should just work. Is that something you're open to adding to this, or shall we start with just gdb?

I added the `lldb` variant, and I tested it on Linux. Presumably it should work fine on MacOS too.

jeaye (approved):
Thank you, Kyle!

PR #988, opened 2026-09-08
Display system exit codes in lein-jank

kylc:
If you run into a segfault etc. with `jank run` you will get a SIGSEGV message printed by your shell. The same code with `lein run` will still quit with the right exit code, but will swallow the message making it appear that the program ran fine.

Adds behavior like Cargo to detect system exit codes and print the relevant string.

```
$ lein run
terminated by signal SIGSEGV: invalid memory reference
```

Left a TODO for Windows support.


jeaye (approved):
Works for me! Thanks for including the link to Cargo.

PR #989, opened 2026-09-10
Send exception_info via nrepl

kylc:
Minimal implementation to get stack traces in CIDER. Sends the nREPL `:ex` field when the evaluation throws, which causes CIDER to eval `(clojure.stacktrace/print-cause-trace *e)`, also implemented minimally here.

I made an effort to always put an `exception_info` in `*e` for consistency, even if we don't have a stack trace or any additional metadata (e.g. if the thrown value was a C++ exception or some jank value).

It would be good for someone using a different nREPL client to test this out. Other clients may use some different `clojure.stacktrace` functionality. I tested by evaluating these:
```
(throw :some-value)
(throw (cpp/std.runtime_error "A runtime error"))
(throw (ex-info "some error" {:hello :world}))
(/ 1 nil)
```

Note that even in CIDER we don't get the "full experience", since that requires cider-nrepl. We just get a text buffer with the pretty-printed `exception_info`.

### DIFF
--- a/compiler+runtime/include/cpp/jank/error/report.hpp
+++ b/compiler+runtime/include/cpp/jank/error/report.hpp
@@ -1,9 +1,10 @@
 #pragma once
 
 #include <jank/error.hpp>
+#include <jtl/immutable_string.hpp>
 
 namespace jank::error
 {
-  void report(error_ref e);
-  void warn(jtl::immutable_string const &);
+  jtl::immutable_string report(error_ref e);
+  jtl::immutable_string warn(jtl::immutable_string const &);
 }
--- a/compiler+runtime/include/cpp/jank/evaluate.hpp
+++ b/compiler+runtime/include/cpp/jank/evaluate.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <jank/runtime/object.hpp>
+#include <jank/read/source.hpp>
 
 namespace jank::runtime::obj
 {
@@ -102,4 +103,6 @@ namespace jank::evaluate
   runtime::object_ref eval(analyze::expr::cpp_unbox_ref);
   runtime::object_ref eval(analyze::expr::cpp_new_ref);
   runtime::object_ref eval(analyze::expr::cpp_delete_ref);
+
+  bool safe_eval(jtl::immutable_string const &code, read::source_position const &p);
 }
--- a/compiler+runtime/include/cpp/jank/runtime/obj/exception_info.hpp
+++ b/compiler+runtime/include/cpp/jank/runtime/obj/exception_info.hpp
@@ -12,7 +12,9 @@
 
 #include <memory>
 
+#include <jank/error.hpp>
 #include <jank/runtime/object.hpp>
+#include <jtl/option.hpp>
 
 /* NOLINTNEXTLINE(modernize-concat-nested-namespaces): Not doable, due to the inline. */
 namespace cpptrace
@@ -35,8 +37,10 @@ namespace jank::runtime::obj
     static constexpr bool pointer_free{ false };
 
     exception_info(jtl::immutable_string const &message, object_ref const data);
+    exception_info(error_ref error);
 
     /* behavior::object_like */
+    jtl::immutable_string to_string() const override;
     jtl::immutable_string to_code_string() const override;
 
     void resolve();
@@ -48,5 +52,6 @@ namespace jank::runtime::obj
     exception_info_ref cause;
     std::unique_ptr<cpptrace::raw_trace> raw_trace;
     std::unique_ptr<cpptrace::stacktrace> resolved_trace;
+    jtl::option<error_ref> error;
   };
 }
--- a/compiler+runtime/src/cpp/jank/error/report.cpp
+++ b/compiler+runtime/src/cpp/jank/error/report.cpp
@@ -5,12 +5,11 @@
 
 namespace jank::error
 {
-  void warn(jtl::immutable_string const &msg)
+  jtl::immutable_string warn(jtl::immutable_string const &msg)
   {
-    util::println(stderr,
-                  "{}warning:{} {}",
-                  jtl::terminal::text_style::yellow,
-                  jtl::terminal::text_style::reset,
-                  msg);
+    return util::format("{}warning:{} {}\n",
+                        jtl::terminal::text_style::yellow,
+                        jtl::terminal::text_style::reset,
+                        msg);
   }
 }
--- a/compiler+runtime/src/cpp/jank/error/report_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/error/report_dynamic.cpp
@@ -3,6 +3,7 @@
 
 #include <boost/algorithm/string/replace.hpp>
 
+#include <jtl/string_builder.hpp>
 #include <jtl/terminal.hpp>
 #include <jtl/utf8.hpp>
 
@@ -764,19 +765,20 @@ namespace jank::error
     return ui::highlight_cpp(source);
   }
 
-  void report(error_ref const e)
+  jtl::immutable_string report(error_ref const e)
   {
+    jtl::string_builder sb;
     plan const p{ e };
 
     auto const terminal_width{ jtl::terminal::get_size().width };
     auto const max_width{ std::min(terminal_width, 100ull) };
 
-    util::println(stderr, "{}", header(kind_str(e->kind), max_width));
-    util::println(stderr,
-                  "{}error:{} {}\n",
-                  text_style::bold | text_style::red,
-                  text_style::reset,
-                  e->message);
+    util::format_to(sb, "{}\n", header(kind_str(e->kind), max_width));
+    util::format_to(sb,
+                    "{}error:{} {}\n\n",
+                    text_style::bold | text_style::red,
+                    text_style::reset,
+                    e->message);
 
     /* TODO: Context. */
     //auto context{ vbox(
@@ -797,47 +799,47 @@ namespace jank::error
 
     for(usize i{}; i < p.snippets.size(); ++i)
     {
-      util::println(stderr,
-                    "{}",
-                    code_snippet(p.snippets[i], max_width, i == p.snippets.size() - 1));
+      util::format_to(sb,
+                      "{}\n",
+                      code_snippet(p.snippets[i], max_width, i == p.snippets.size() - 1));
     }
     if(p.snippets.empty())
     {
-      util::println(stderr,
-                    "https://book.jank-lang.org/reference/error/{}.html",
-                    kind_str(e->kind));
+      util::format_to(sb,
+                      "https://book.jank-lang.org/reference/error/{}.html\n",
+                      kind_str(e->kind));
     }
     else
     {
-      util::println(stderr, "{}", documentation_box(e, max_width));
+      util::format_to(sb, "{}\n", documentation_box(e, max_width));
     }
 
     if(!e->candidates.empty())
     {
       auto const show_count{ std::min(e->candidates.size(),
                                       static_cast<size_t>(util::cli::opts.max_error_candidates)) };
-      util::println(stderr,
-                    "  Candidates considered (showing {} of {}):\n",
-                    show_count,
-                    e->candidates.size());
+      util::format_to(sb,
+                      "  Candidates considered (showing {} of {}):\n\n",
+                      show_count,
+                      e->candidates.size());
       for(size_t i{}; i < show_count; ++i)
       {
         auto const &c{ e->candidates[i] };
 
         if(c.viable)
         {
-          util::print(stderr, "  {}✓{}", text_style::green, text_style::reset);
+          util::format_to(sb, "  {}✓{}", text_style::green, text_style::reset);
         }
         else
         {
-          util::print(stderr, "  {}✗{}", text_style::red, text_style::reset);
+          util::format_to(sb, "  {}✗{}", text_style::red, text_style::reset);
         }
-        util::print(stderr, " {}", format_and_highlight_cpp(c.signature, "    "));
-        util::println(stderr,
-                      "{}╰─ declared at {}{}",
-                      text_style::bright_black,
-                      shorten_path(c.source),
-                      text_style::reset);
+        util::format_to(sb, " {}", format_and_highlight_cpp(c.signature, "    "));
+        util::format_to(sb,
+                        "{}╰─ declared at {}{}\n",
+                        text_style::bright_black,
+                        shorten_path(c.source),
+                        text_style::reset);
 
         native_vector<native_vector<jtl::immutable_string>> rows;
         for(auto const &a : c.arguments)
@@ -870,28 +872,30 @@ namespace jank::error
         }
 
         auto const column_padding{ determine_column_padding(rows, 1) };
-        util::print(stderr, "{}", render_columns("    ", rows, column_padding));
+        util::format_to(sb, "{}", render_columns("    ", rows, column_padding));
 
-        util::println(stderr,
-                      "    {}──────────────────────────────────────────────────────────{}",
-                      text_style::bright_black,
-                      text_style::reset);
-        util::println(stderr, "    {}{}", (c.viable ? "Viable: " : "Not viable: "), c.reason);
+        util::format_to(sb,
+                        "    {}──────────────────────────────────────────────────────────{}\n",
+                        text_style::bright_black,
+                        text_style::reset);
+        util::format_to(sb, "    {}{}\n", (c.viable ? "Viable: " : "Not viable: "), c.reason);
         if(!c.clang_reason.empty())
         {
-          util::println(stderr,
-                        "    {}╰─ {}{}",
-                        text_style::bright_black,
-                        c.clang_reason,
-                        text_style::reset);
+          util::format_to(sb,
+                          "    {}╰─ {}{}\n",
+                          text_style::bright_black,
+                          c.clang_reason,
+                          text_style::reset);
         }
-        util::println(stderr, "");
+        util::format_to(sb, "\n");
       }
     }
 
     if(e->cause)
     {
-      report(e->cause.as_ref());
+      sb.push_back(report(e->cause.as_ref()));
     }
+
+    return sb.release();
   }
 }
--- a/compiler+runtime/src/cpp/jank/error/report_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/error/report_dynamic.cpp
@@ -771,8 +771,9 @@ namespace jank::error
     auto const terminal_width{ jtl::terminal::get_size().width };
     auto const max_width{ std::min(terminal_width, 100ull) };
 
-    util::println("{}", header(kind_str(e->kind), max_width));
-    util::println("{}error:{} {}\n",
+    util::println(stderr, "{}", header(kind_str(e->kind), max_width));
+    util::println(stderr,
+                  "{}error:{} {}\n",
                   text_style::bold | text_style::red,
                   text_style::reset,
                   e->message);
@@ -796,22 +797,27 @@ namespace jank::error
 
     for(usize i{}; i < p.snippets.size(); ++i)
     {
-      util::println("{}", code_snippet(p.snippets[i], max_width, i == p.snippets.size() - 1));
+      util::println(stderr,
+                    "{}",
+                    code_snippet(p.snippets[i], max_width, i == p.snippets.size() - 1));
     }
     if(p.snippets.empty())
     {
-      util::println("https://book.jank-lang.org/reference/error/{}.html", kind_str(e->kind));
+      util::println(stderr,
+                    "https://book.jank-lang.org/reference/error/{}.html",
+                    kind_str(e->kind));
     }
     else
     {
-      util::println("{}", documentation_box(e, max_width));
+      util::println(stderr, "{}", documentation_box(e, max_width));
     }
 
     if(!e->candidates.empty())
     {
       auto const show_count{ std::min(e->candidates.size(),
                                       static_cast<size_t>(util::cli::opts.max_error_candidates)) };
-      util::println("  Candidates considered (showing {} of {}):\n",
+      util::println(stderr,
+                    "  Candidates considered (showing {} of {}):\n",
                     show_count,
                     e->candidates.size());
       for(size_t i{}; i < show_count; ++i)
@@ -820,14 +826,15 @@ namespace jank::error
 
         if(c.viable)
         {
-          util::print("  {}✓{}", text_style::green, text_style::reset);
+          util::print(stderr, "  {}✓{}", text_style::green, text_style::reset);
         }
         else
         {
-          util::print("  {}✗{}", text_style::red, text_style::reset);
+          util::print(stderr, "  {}✗{}", text_style::red, text_style::reset);
         }
-        util::print(" {}", format_and_highlight_cpp(c.signature, "    "));
-        util::println("{}╰─ declared at {}{}",
+        util::print(stderr, " {}", format_and_highlight_cpp(c.signature, "    "));
+        util::println(stderr,
+                      "{}╰─ declared at {}{}",
                       text_style::bright_black,
                       shorten_path(c.source),
                       text_style::reset);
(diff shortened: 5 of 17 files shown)


jeaye:
I've tried this out locally and didn't make it past `(throw :some-value)`. I end up with `std::terminate` being called, since the exception is not caught.

```
#0  0x00007fffe769fdcc in __pthread_kill_implementation () from /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6
#1  0x00007fffe764265e in raise () from /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6
#2  0x00007fffe7629350 in abort () from /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6
#3  0x00007fffe7ab2ce4 in __gnu_cxx::__verbose_terminate_handler() [clone .cold] ()
   from /nix/store/chqq8mpmpyfi9kgsngya71akv5xicn03-gcc-15.2.0-lib/lib/libstdc++.so.6
#4  0x00007fffe7ac539a in __cxxabiv1::__terminate(void (*)()) () from /nix/store/chqq8mpmpyfi9kgsngya71akv5xicn03-gcc-15.2.0-lib/lib/libstdc++.so.6
#5  0x00007fffe7ab286e in std::terminate() () from /nix/store/chqq8mpmpyfi9kgsngya71akv5xicn03-gcc-15.2.0-lib/lib/libstdc++.so.6
#6  0x00007fffe7ac5637 in __cxa_throw () from /nix/store/chqq8mpmpyfi9kgsngya71akv5xicn03-gcc-15.2.0-lib/lib/libstdc++.so.6
#7  0x0000555556622831 in jank::evaluate::eval (expr=...) at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/evaluate.cpp:570
#8  0x000055555666ef56 in jank::evaluate::eval(jtl::ref<jank::analyze::expression>)::$_0::operator()<jtl::ref<jank::analyze::expr::throw_> >(jtl::ref<jank::analyze::expr::throw_>) const (this=0x7fffcae6d0c8, typed_ex=...) at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/evaluate.cpp:267
#9  0x000055555661e75b in jank::analyze::visit_expr<jank::evaluate::eval(jtl::ref<jank::analyze::expression>)::$_0>(jank::evaluate::eval(jtl::ref<jank::analyze::expression>)::$_0 const&, jtl::ref<jank::analyze::expression>) (f=..., e=...)
    at /home/jeaye/projects/jank/compiler+runtime/include/cpp/jank/analyze/visit.hpp:85
#10 0x000055555661e177 in jank::evaluate::eval (ex=...) at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/evaluate.cpp:267
#11 0x0000555556758e81 in jank::runtime::context::eval_string (this=0x7fffe7985a80, code=..., p=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/runtime/context_dynamic.cpp:72
#12 0x00007fffcd7a1075 in jank::nrepl::server::eval::safe_eval (code=..., p=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/jank/jank/nrepl/server/eval.jank:18
#13 0x00007fffcd7a0052 in jank_nrepl_server_eval_safe_eval_8_3 (safe_eval=..., code=..., line=..., col=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/jank/jank/nrepl/server/eval.jank:63
#14 0x0000555555e4df3f in jank::runtime::obj::jit_function::call (this=0x7fffcca9f2c0, a1=..., a2=..., a3=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/runtime/obj/jit_function.cpp:93
#15 0x00007fffcd7a7189 in jank_nrepl_server_handler_eval_fn_128_0 (jank_nrepl_server_handler_eval_fn_128=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/jank/jank/nrepl/server/handler/eval.jank:31
#16 0x0000555555e5bb8b in jank::runtime::obj::jit_closure::call (this=0x7fffca484b40)
    at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/runtime/obj/jit_closure.cpp:67
#17 0x00007fffcd798675 in jank::runtime::oref<jank::runtime::object>::call (this=<optimized out>)
    at /home/jeaye/projects/jank/compiler+runtime/include/cpp/jank/runtime/oref.hpp:542
#18 jank_nrepl_server_capture_with_capture_118_1 (with_capture=..., f=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/jank/jank/nrepl/server/capture.jank:70
#19 0x0000555555e4dbe1 in jank::runtime::obj::jit_function::call (this=0x7fffcc727840, a1=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/runtime/obj/jit_function.cpp:75
#20 0x00007fffcd7a76a8 in jank_nrepl_server_handler_eval_do_eval_117_6 (do_eval=..., ns=..., code=..., file_path=..., line=..., col=..., inspect_QMARK_=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/jank/jank/nrepl/server/handler/eval.jank:26
#21 0x0000555555e4e550 in jank::runtime::obj::jit_function::call (this=0x7fffcc7d7d10, a1=..., a2=..., a3=..., a4=..., a5=..., a6=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/runtime/obj/jit_function.cpp:132
#22 0x00007fffcd7a8fd4 in jank_nrepl_server_handler_eval_fn_137_1 (jank_nrepl_server_handler_eval_fn_137=..., msg=...)
    at /home/jeaye/projects/jank/compiler+runtime/src/jank/jank/nrepl/server/handler/eval.jank:59
```

The exception is thrown here, which is an `object_ref`.

```
#1  0x0000555556622871 in jank::evaluate::eval (expr=...) at /home/jeaye/projects/jank/compiler+runtime/src/cpp/jank/evaluate.cpp:570
570	      throw eval(expr->value.unwrap());
```

But we're not catching it. I verified the behavior is also broken on `main` for me. LLVM historically had some issues with this (catching exceptions across JIT compiled frames), so I wonder if the LLVM build in the Nix flake is too old, but I figure that's exactly what you must be using? Since it's broken on `main` and also your branch, it means I can't test your actual changes very well.

------

Ok, I also tested in an Ubuntu distrobox, with jank's local Clang 23. Same problem. The stack trace is largely the same:

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=6, no_tid=0) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (threadid=<optimized out>, signo=6) at ./nptl/pthread_kill.c:89
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:100
#3  0x00007fffee584b7e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007fffee5678ec in __GI_abort () at ./stdlib/abort.c:77
#5  0x00007fffee80c275 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007fffee82364a in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007fffee80bc6c in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007fffee823901 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#9  0x000055555644576b in jank::evaluate::eval (expr=...) at /home/jeaye/projects/jank-ai/compiler+runtime/src/cpp/jank/evaluate.cpp:570
#10 0x0000555556480d79 in jank::evaluate::eval(jtl::ref<jank::analyze::expression>)::$_0::operator()<jtl::ref<jank::analyze::expr::throw_> >(jtl::ref<jank::analyze::expr::throw_>) const (this=0x7fffdeff3170, typed_ex=...) at /home/jeaye/projects/jank-ai/compiler+runtime/src/cpp/jank/evaluate.cpp:267
#11 0x0000555556441fc2 in jank::analyze::visit_expr<jank::evaluate::eval(jtl::ref<jank::analyze::expression>)::$_0>(jank::evaluate::eval(jtl::ref<jank::analyze::expression>)::$_0 const&, jtl::ref<jank::analyze::expression>) (f=..., e=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/include/cpp/jank/analyze/visit.hpp:85
#12 0x0000555556441ab0 in jank::evaluate::eval (ex=...) at /home/jeaye/projects/jank-ai/compiler+runtime/src/cpp/jank/evaluate.cpp:267
#13 0x000055555654337c in jank::runtime::context::eval_string (this=0x7fffee1d7a80, code=..., p=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/cpp/jank/runtime/context_dynamic.cpp:72
#14 0x00007fffe0970075 in jank::nrepl::server::eval::safe_eval (code=..., p=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/jank/jank/nrepl/server/eval.jank:18
#15 0x00007fffe096f052 in jank_nrepl_server_eval_safe_eval_8_3 (safe_eval=..., code=..., line=..., col=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/jank/jank/nrepl/server/eval.jank:63
#16 0x0000555555d41073 in jank::runtime::obj::jit_function::call (this=0x7fffe07d5d10, a1=..., a2=..., a3=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/cpp/jank/runtime/obj/jit_function.cpp:93
#17 0x00007fffe0976189 in jank_nrepl_server_handler_eval_fn_128_0 (jank_nrepl_server_handler_eval_fn_128=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/jank/jank/nrepl/server/handler/eval.jank:31
#18 0x0000555555d4c372 in jank::runtime::obj::jit_closure::call (this=0x7fffde5e69c0)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/cpp/jank/runtime/obj/jit_closure.cpp:67
#19 0x00007fffe0967675 in jank::runtime::oref<jank::runtime::object>::call (this=<optimized out>)
    at /home/jeaye/projects/jank-ai/compiler+runtime/include/cpp/jank/runtime/oref.hpp:542
#20 jank_nrepl_server_capture_with_capture_118_1 (with_capture=..., f=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/jank/jank/nrepl/server/capture.jank:70
#21 0x0000555555d40e08 in jank::runtime::obj::jit_function::call (this=0x7fffe0671e70, a1=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/cpp/jank/runtime/obj/jit_function.cpp:75
#22 0x00007fffe09766a8 in jank_nrepl_server_handler_eval_do_eval_117_6 (do_eval=..., ns=..., code=..., file_path=..., line=..., col=..., inspect_QMARK_=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/jank/jank/nrepl/server/handler/eval.jank:26
#23 0x0000555555d414f2 in jank::runtime::obj::jit_function::call (this=0x7fffe074cc60, a1=..., a2=..., a3=..., a4=..., a5=..., a6=...)
    at /home/jeaye/projects/jank-ai/compiler+runtime/src/cpp/jank/runtime/obj/jit_function.cpp:132
#24 0x00007fffe0977fd4 in jank_nrepl_server_handler_eval_fn_137_1 (jank_nrepl_server_handler_eval_fn_137=..., msg=...)
```

-----

What're your setup details, Kyle? I wonder why this isn't affecting you.

I was just doing a `lein repl` in an imgui example, then doing a `:ConjureConnect <port>` and then evaluating the whole buffer, and then evaluating the `throw` in a rich comment.

kylc:
> What're your setup details, Kyle? I wonder why this isn't affecting you.

I'm using a clean checkout of this branch with the Nix flake. Health check below, including the clang version.

I'm just starting up a blank repl with `jank repl`, connecting to it with `cider-connect`, and then evaluating `(throw :some-value)` from CIDER. 

I also tested in gdb using `break jank::nrepl::server::eval::safe_eval` and `catch catch` and everything seems OK to me (the jank line numbers might be a little off for some reason though).

I'm not sure what could be different with my setup from yours. Let me know if anything looks different in the health check.

```
─ system ───────────────────────────────────────────────────────────────────────────────────────────
─ ✅ operating system: linux
─ ✅ default triple: x86_64-unknown-linux-gnu

─ jank install ─────────────────────────────────────────────────────────────────────────────────────
─ ✅ jank version: jank-0.1-alpha
─ ✅ jank cmake build type: RelWithDebInfo
─ ✅ jank resource dir: ../lib/jank/0.1 
─ ✅ jank resolved resource dir: /home/kyle/code/jank/compiler+runtime/build/../lib/jank/0.1 (ignored for dev build)
─ ✅ jank user cache dir: /home/kyle/.cache/jank/x86_64-unknown-linux-gnu-8a08ddd4115e1ae1c754f1ff111a9e64992938217ea472c2921b0af3c32051a0 (not found)

─ clang install ────────────────────────────────────────────────────────────────────────────────────
─ ✅ configured clang path: /nix/store/yjvbvb8rm3pbws3hap8csf62jf9cgvfc-clang-wrapper-23.0.0-unstable-2026-06-21/bin/clang++ (found)
─ ✅ configured clang resource dir: /nix/store/yjvbvb8rm3pbws3hap8csf62jf9cgvfc-clang-wrapper-23.0.0-unstable-2026-06-21/resource-root (found)

─ jank runtime ─────────────────────────────────────────────────────────────────────────────────────
─ ✅ jank runtime initialized
─ ✅ jank pch path: /home/kyle/.cache/jank/x86_64-unknown-linux-gnu-8a08ddd4115e1ae1c754f1ff111a9e64992938217ea472c2921b0af3c32051a0 (found)
─ ✅ jank can jit compile c++
─ ✅ jank can aot compile working binaries
```

jeaye:
Thanks for the details. I did a clean build. My health check show that you and I are using the same Clang version.

```
❯ jank check-health
─ system ───────────────────────────────────────────────────────────────────────────────────────────
─ ✅ operating system: linux
─ ✅ default triple: x86_64-unknown-linux-gnu

─ jank install ─────────────────────────────────────────────────────────────────────────────────────
─ ✅ jank version: jank-0.1-alpha
─ ✅ jank cmake build type: Debug
─ ⚠️  jank assertions are enabled; performance will be impacted
─ ✅ jank resource dir: ../lib/jank/0.1
─ ✅ jank resolved resource dir: /home/jeaye/projects/jank/compiler+runtime/build/../lib/jank/0.1 (ignored for dev build)
─ ✅ jank user cache dir: /home/jeaye/.cache/jank/x86_64-unknown-linux-gnu-34a11d455c721963dc566ce2da6a406e8eda532455372c26d4f17eb2df6a0dc1 (not found)

─ clang install ────────────────────────────────────────────────────────────────────────────────────
─ ✅ configured clang path: /nix/store/yjvbvb8rm3pbws3hap8csf62jf9cgvfc-clang-wrapper-23.0.0-unstable-2026-06-21/bin/clang++ (found)
─ ✅ configured clang resource dir: /nix/store/yjvbvb8rm3pbws3hap8csf62jf9cgvfc-clang-wrapper-23.0.0-unstable-2026-06-21/resource-root (found)

─ jank runtime ─────────────────────────────────────────────────────────────────────────────────────
─ ✅ jank runtime initialized
─ ✅ jank pch path: /home/jeaye/.cache/jank/x86_64-unknown-linux-gnu-34a11d455c721963dc566ce2da6a406e8eda532455372c26d4f17eb2df6a0dc1 (found)
─ ✅ jank can jit compile c++
─ ✅ jank can aot compile working binaries

─ support ──────────────────────────────────────────────────────────────────────────────────────────
If you're having issues with jank, please either ask the jank community on the Clojurians Slack or report the issue on Github.

─ Slack: https://clojurians.slack.com/archives/C03SRH97FDK
─ Github: https://github.com/jank-lang/jank
```

Now here's the fun bit. If I build for `Debug`, I get the crashes described above. Exceptions thrown in REPL-evaluated code result in `std::terminate`. If I build for `RelWithDebInfo`, as you have, that problem goes away and instead I get responses back from the REPL server, but Conjure doesn't show me any error messages. Here's an example of a `RelWithDebInfo` build, doing a `(/ 1 nil)`.

```
<- {"id" "7ecf7a21-44a1-4e92-9…", "line" 4, "column" 1, "file" "/home/jeaye/projects…", "nrepl.middleware.pri…" "conjure.internal/ppr…", "code" "(/ 1 nil)", "op" "eval", "session" "b5174495-0b06-4e1e-b…"}
-> {:out "", "id" "7ecf7a21-44a1-4e92-9…", "session" "b5174495-0b06-4e1e-b…"}
-> {:err "", "id" "7ecf7a21-44a1-4e92-9…", "session" "b5174495-0b06-4e1e-b…"}
-> {:ex "#error {\n :cause \"Ca…", :status [:done :eval-error], "id" "7ecf7a21-44a1-4e92-9…", "session" "b5174495-0b06-4e1e-b…"}
```

In Conjure, this just renders as empty `(out)` and empty `(err)`, but no exception.

However, on `main`, if I build for `RelWithDebInfo`, then everything works nicely. I also get error messages from Conjure. So, let me summarize this with a table.

|      | Debug     | RelWithDebInfo    |
|------|-----------|-------------------|
| main | terminate | works             |
| pr   | terminate | no error messages |

-----

Kyle, would you mind trying a `Debug` build on your end?

kylc:
I tried a `Debug` build and jank terminates. So do we merge something that works only with optimized builds and file away an LLVM investigation for later?

It's hard to say what Conjure expects. https://spec.nrepl.org seems to indicate that you can just populate the "ex" field like we do here, but I guess every client has its own behavior. What we can do is output the full error on the stderr stream as well, as a catch-all. This might result in some duplication but it looks OK in CIDER, and I assume this makes it show up somewhere in Conjure (probably the REPL buffer). If that works, I will leave any further investigation into specialized Conjure behavior to a vim user.

Accidentally force pushed BTW.

jeaye:
> I tried a `Debug` build and jank terminates. So do we merge something that works only with optimized builds and file away an LLVM investigation for later?

Thanks for giving this a look. I am relieved that it's failing for you with `Debug` as well. Yep, if it's failing on `main` already anyway, it's not something we need to worry about right now. It won't (seemingly) affect production jank users either.

> It's hard to say what Conjure expects. https://spec.nrepl.org seems to indicate that you can just populate the "ex" field like we do here, but I guess every client has its own behavior. What we can do is output the full error on the stderr stream as well, as a catch-all. This might result in some duplication but it looks OK in CIDER, and I assume this makes it show up somewhere in Conjure (probably the REPL buffer). If that works, I will leave any further investigation into specialized Conjure behavior to a vim user.

Yep, very reasonable. I have just checked this locally and it works fine.

I also added the `error_ref` handling back in, with full reporting (including C++ candidate info). I patched the reporting to always output to `stderr`. I noticed you had originally removed the `error_ref` catch. Does this still work for you on the CIDER side?

The only annoying thing about it is that the `exception_info` we show doesn't actually help. It's just noise at the bottom of the output. I'm not sure if there's a way we can get the "correct" output for everyone involved. To me, "correct" means the full error report, without duplication, when available, and the `exception_info` otherwise.

```
; eval (current-form): (cpp/rand 1 2)
; (out) 
; (err) ─ analyze/invalid-cpp-call ─────────────────────────────────────────────────────
; (err) error: No matching overload of the `rand` function was found for this call.
; (err) 
; (err) ─────┬──────────────────────────────────────────────────────────────────────────
; (err)      │ test.jank
; (err) ─────┼──────────────────────────────────────────────────────────────────────────
; (err)   3  │ (throw (ex-info "some error" {:hello :world}))
; (err)   4  │ (/ 1 nil)
; (err)   5  │  
; (err)      │  ^^^^^^^^ Found here.
; (err) ─────┼──────────────────────────────────────────────────────────────────────────
; (err)   🗎  │ https://book.jank-lang.org/reference/error/analyze/invalid-cpp-call.html
; (err) ─────┴──────────────────────────────────────────────────────────────────────────
; (err) 
; (err)   Candidates considered (showing 1 of 1):
; (err) 
; (err)   ✗ extern int rand() noexcept(true)
; (err)     ╰─ declared at /nix/store/15h9askp4k1lx44d9871wid23j2a8ijp-glibc-2.42-61-dev/include/stdlib.h:573
; (err)     ──────────────────────────────────────────────────────────
; (err)     Not viable: This function requires 0 arguments, but 2 were provided.
; (err) 
; (err) #error {
; (err)  :cause "No matching overload of the `rand` function was found for this call."
; (err)  :data nil
; (err)  :via
; (err)  [{:message "No matching overload of the `rand` function was found for this call.", :data nil}]
```

kylc:
Your change works for me.

I'm not sure of a good solution to avoid the duplication. Whatever we put in `*e`, `(clojure.stacktrace/print-cause-trace *e)` must at least work or else CIDER will break.

Perhaps we could put either a `string` or an `exception_info` into `*e`, and then in `clojure.stacktrace/print-cause-trace` we ignore non-`exception_info` inputs, or just return string inputs as-is? It's an unfortunate JVM Clojure-ism because in this case, if we want to show the `error_ref` report in the CIDER error buffer, we need to return it from `print-cause-trace` which is a misnomer.

jeaye:
> Perhaps we could put either a `string` or an `exception_info` into `*e`, and then in `clojure.stacktrace/print-cause-trace` we ignore non-`exception_info` inputs, or just return string inputs as-is? It's an unfortunate JVM Clojure-ism because in this case, if we want to show the `error_ref` report in the CIDER error buffer, we need to return it from `print-cause-trace` which is a misnomer.

If we add an `error_ref` member to `exception_info`, we could check in `clojure.stacktrace/print-cause-trace` whether the `error_ref` member is set. If it is, we could just do `jank::error::report` on it and ignore the rest of the `exception_info`. Otherwise, we do the `prn` we're currently doing. But that wouldn't address Conjure, or potentially other clients, which don't call `print-cause-trace`.

kylc:
I've implemented your suggestion of adding the optional `error_ref` to `exception_info`. Now in my `*cider-error*` buffer, when I throw I get:
1) for a regular object, the `exception_info` code string without a stack trace
2) for an ex-info, the `exception_info` code string with a stack trace
3) for a compiler error, the error report

This behavior is mirrored on the nREPL `:err` stream, so I think you will see the same in Conjure.

Some of the code change feels like a shoehorn to me but you know the context better, so let me know if you want more changes.

https://github.com/user-attachments/assets/1f97b2e1-aed3-4660-acff-0b9c59d67c88

jeaye:
> I've implemented your suggestion of adding the optional `error_ref` to `exception_info`. Now in my `*cider-error*` buffer, when I throw I get:
> 
>     1. for a regular object, the `exception_info` code string without a stack trace
> 
>     2. for an ex-info, the `exception_info` code string with a stack trace
> 
>     3. for a compiler error, the error report
> 
> 
> This behavior is mirrored on the nREPL `:err` stream, so I think you will see the same in Conjure.

Nice! I've just pushed some changes which add stack traces to all of the cases. I moved the C++ fn into jank's C++ code as well, since `cpptrace::try_catch` does some weird RTTI shit which the JIT runtime doesn't like. This is a bit noisier, since every exception has a trace now, but I think that's better in the long term. Let me know what you think.

> Some of the code change feels like a shoehorn to me but you know the context better, so let me know if you want more changes.

Yeah, it's not perfect. Perhaps we can fix this down the road by replacing `error_ref` (used for errors produced by the jank compiler/runtime) with a type which inherits from `exception_info`. Then we can just rely on virtual dispatch for everything. Not something we need to worry about right now, though.

Thank you for iterating on the UX of this. If you're cool with my latest changes, I'm cool with getting this merged. This PR is already a _huge_ win in terms of jank REPL usage.

kylc:
Pushed a small fix so `safe_eval` returns the status of the eval. CIDER looks OK with this commit.

> Nice! I've just pushed some changes which add stack traces to all of the cases.

That looks fine to me.

Should we be using the same formatting for stacktraces/exception_info that the compiler uses, instead of showing the code string?

jeaye:
> Should we be using the same formatting for stacktraces/exception_info that the compiler uses, instead of showing the code string?

That's a great idea, I think. It's much easier to read that way. The code string is ok, but it's machine output (EDN) formatted for humans.

Are there any benefits of showing the code string? Based on my REPL workflow, I'd say no, but past conversations with folks have indicated that Emacs users have very different REPL workflows than me.

jeaye:
> Pushed a small fix so safe_eval returns the status of the eval. CIDER looks OK with this commit.

:man_facepalming: Thanks for the fix!

kylc:
> Are there any benefits of showing the code string? Based on my REPL workflow, I'd say no, but past conversations with folks have indicated that Emacs users have very different REPL workflows than me.

They might be referring to the `cider-nrepl` plugin, which runs in the Clojure process (next to the nREPL server) and provides additional nREPL functionality for CIDER to get more information/interactivity. It has its own pathway for extracting the exception stack trace etc. But we don't have `cider-nrepl` for jank, and that's orthogonal to the way we print errors when responding to `clojure.stacktrace/print-cause-trace`, so I don't think we need to worry about it.

Otherwise as a CIDER user I think it would be nice to have a human-readable exception printout in the error buffer.

I will leave this work for later unless you want to take it on. It seems like it will be some work to convert the `print_exception` machinery to output a string so that we can use it in `exception_info::to_string()`.

jeaye:
> Otherwise as a CIDER user I think it would be nice to have a human-readable exception printout in the error buffer.
> 
> I will leave this work for later unless you want to take it on. It seems like it will be some work to convert the `print_exception` machinery to output a string so that we can use it in `exception_info::to_string()`.

Sounds good! Let's not let perfect be the enemy of good.

PR #990, opened 2026-09-10
WIP: Add "cargo check"-type workflow

kylc:
This is just an exploratory PR for high-level feedback. We can ignore the implementation details for now.

This PR is towards a `cargo check` workflow with the jank compiler, which can be run by the user's editor to show jank compiler warnings/errors.

What's done here:
- Add a `compilation_target::none` to avoid codegen, since we only care about getting the jank analysis errors quickly.
- Use an easier-to-parse error format `file:line:col` like standard GCC errors. This also makes the errors display nicer in emacs `compilation-mode` since it can easily pick out the start of the next error. They look a little ugly though so maybe we find a better way.

What needs to be figured out:
- What does the workflow look like with `lein`? `lein check` would be equivalent to `cargo check` (adding all the compiler flags for the project and calling the jank compiler without codegen). But `lein` is too slow to start for a CLI linting tool. Maybe we just ignore that for now until we get a faster CLI?
- Can `jank compile` output more than 1 compiler error? It seems to short-circuit for now.

---

I've copy-pasted https://github.com/turbo-cafe/flymake-kondor/ into my emacs config and modified it to invoke jank and parse the jank errors. We can publish this as an official `flymake-jank` plugin later. jank errors now show up when I save a `jank-mode` buffer, just like any other linter/compiler plugin.

It should be possible to do the same with vim/quickfix.

image

image

### DIFF
--- a/compiler+runtime/include/cpp/jank/util/cli.hpp
+++ b/compiler+runtime/include/cpp/jank/util/cli.hpp
@@ -23,7 +23,8 @@ namespace jank::util::cli
      * If that's not possible, we'll error out. */
     unspecified,
     cpp,
-    object
+    object,
+    none,
   };
 
   constexpr char const *compilation_target_str(compilation_target const target)
@@ -36,6 +37,8 @@ namespace jank::util::cli
         return "cpp";
       case compilation_target::object:
         return "object";
+      case compilation_target::none:
+        return "none";
       default:
         return "unknown";
     }
@@ -51,6 +54,8 @@ namespace jank::util::cli
         return "cpp";
       case compilation_target::object:
         return "o";
+      case compilation_target::none:
+        return "";
       default:
         return "unknown";
     }
--- a/compiler+runtime/src/cpp/jank/error/report_dynamic.cpp
+++ b/compiler+runtime/src/cpp/jank/error/report_dynamic.cpp
@@ -770,10 +770,15 @@ namespace jank::error
 
     auto const terminal_width{ jtl::terminal::get_size().width };
     auto const max_width{ std::min(terminal_width, 100ull) };
+    auto const file_name{ e->source.file.substr(e->source.file.rfind("/") + 1) };
 
     util::println("{}", header(kind_str(e->kind), max_width));
-    util::println("{}error:{} {}\n",
+
+    util::println("{}{}:{}:{}: error:{} {}\n",
                   text_style::bold | text_style::red,
+                  file_name,
+                  e->source.start.line,
+                  e->source.start.col,
                   text_style::reset,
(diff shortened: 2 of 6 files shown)


kylc:
Fairly usable from a lein project with `lein check`, but there is some latency from `lein` startup like I mentioned.

https://github.com/user-attachments/assets/0706d536-51e4-4858-a9ba-5f1ee2099632

jeaye:
Fascinating! I love that you're exploring this, Kyle. Some initial thoughts:

1. I am not opposed to the addition of a `lein check` for jank; I agree that `lein` is too heavy for this, but we could add it in preparation for the port to the new Clojure CLI (latest update from Alex yesterday says the new CLI will be ready after the Conj)
2. I wonder who would use the "check" workflow versus the LSP workflow; in my mind, reporting compiler errors in-editor is very much an LSP job these days. Of course, jank doesn't yet have an LSP server and getting one is going to be a large effort which will need to balance both Clojure LSP and Clang LSP together. So this may be a good temporary solution, even if LSP is preferred. I know that I've used `lein check` for Clojure projects, since it's a default verb, just to know that everything compiles nicely.
3. I had originally imagined jank would have different output modes, via `--output-mode`, which might be values like `rich` (default), `minimal` (GCC style), and `json`. Perhaps that's a good long-term solution to this `file:line:col` parsing problem, since it doesn't fit into the `rich` output well.
4. Yes, jank will only every give one error, most likely for the long-term. I think this makes sense for Clojure, since each expression can be an effect which mutates the global environment in a way which makes the next line valid to analyze (like defining a var). If we tried to continue analyzing after an error, we'd just be doing guess work, so I don't have confidence that the quality of errors would be worthwhile.

kylc:
> I wonder who would use the "check" workflow versus the LSP workflow

I guess everyone has their own workflow, but the former is what I do for Clojure. I know that `clojure-lsp` exists, but I just use `clj-kondo` via flymake in emacs, plus CIDER for REPL. So my full jank setup would probably be CIDER + `lein check` + `clj-kondo`.

Unfortunately from what I understand, an LSP server (even one which doesn't do both jank and C++ code) is a huge undertaking and the protocol is pretty complex. I think, for users who eventually want an LSP, `lein check` is at least a stepping stone.

> I had originally imagined jank would have different output modes

That would be nice. I looked at a `cargo check` flymake plugin and Cargo/rustc is able to output errors as JSON for that purpose.

> Yes, jank will only every give one error, most likely for the long-term.

I guess that's reasonable enough when combining `lein check` with `clj-kondo` or some other static analyzer, since the latter can catch the simpler Clojure errors en masse.

PR #991, opened 2026-09-11
Don't close temp-fd twice in nrepl server

kylc:
It's already closed on L32. This was causing stdout/stderr to not be redirected back to their original streams in the terminal REPL after doing evals via nREPL.


jeaye:
Thank you!

PR #993, opened 2026-09-11
Don't send empty :out/:err messages over nrepl

kylc:
These show up as empty `(out)` and `(err)` line spam in Conjure.


jeaye:
This is a nice UX improvement! I did a quick merge conflict resolution and now I need to run afk for the night. CI should be green when I get back and I'll merge it in.

Thank you for all of the help, Kyle!

PR #994, opened 2026-09-12
Guard against concurrent access to clang::Interpreter

### LINKED ISSUE #986: Exitting with code `139` when calling functions in a thread
I was testing some async code and found some weird issue: when I have multiple functions and I call them in a `future`, the process exists with status `139`:

```clojure
(ns bugs.main)

(defn work-0 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-1 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-2 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-3 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-4 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-5 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-6 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-7 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-8 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-9 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))
(defn work-10 [n] (loop [i n acc 0] (if (zero? i) acc (recur (dec i) (+ acc i)))))

(def all [work-0 work-1 work-2 work-3 work-4 work-5 work-6 work-7 work-8 work-9 work-10])

(defn -main [& _]
  (let [fs (doall (map (fn [f] (future (f 10000))) all))]
    (println "sum:" (reduce + (map deref fs)))))
```

This doesn't happen if I first run `lein compile`, but it does if I run with `lein run`.

kylc:
`clang::Interpreter` isn't thread-safe so we need to guard against concurrent access to it in the `processor` (for example by `deferred_cpp_function`s being realized on separate threads).

Here I've blasted locks all over the `processor` member functions, but we can also try to be more fine-grained about it. Alternatively we can wrap it in a `folly::Synchronized` if that's still en vogue in the jank codebase.

Fixes https://github.com/jank-lang/jank/issues/986

### DIFF
--- a/compiler+runtime/include/cpp/jank/jit/processor.hpp
+++ b/compiler+runtime/include/cpp/jank/jit/processor.hpp
@@ -92,7 +92,6 @@ namespace jank::jit
     std::map<char const *, std::string_view> vfs;
 
     /*** XXX: Everything here is thread-safe. ***/
-    folly::Synchronized<jtl::ptr<CppInternal::Interpreter>, std::recursive_mutex>
-      interpreter;
+    folly::Synchronized<jtl::ptr<CppInternal::Interpreter>, std::recursive_mutex> interpreter;
   };
 }
--- a/compiler+runtime/include/cpp/jank/jit/processor.hpp
+++ b/compiler+runtime/include/cpp/jank/jit/processor.hpp
@@ -92,7 +92,7 @@ namespace jank::jit
     std::map<char const *, std::string_view> vfs;
 
     /*** XXX: Everything here is thread-safe. ***/
-    mutable folly::Synchronized<jtl::ptr<CppInternal::Interpreter>, std::recursive_mutex>
+    folly::Synchronized<jtl::ptr<CppInternal::Interpreter>, std::recursive_mutex>
       interpreter;
   };
 }
--- a/compiler+runtime/include/cpp/jank/jit/processor.hpp
+++ b/compiler+runtime/include/cpp/jank/jit/processor.hpp
@@ -2,7 +2,9 @@
 
 #include <filesystem>
 #include <map>
+#include <mutex>
 
+#include <folly/Synchronized.h>
 #include <jtl/result.hpp>
 #include <jtl/string_builder.hpp>
 
@@ -82,7 +84,6 @@ namespace jank::jit
              jtl::immutable_string const &lib);
 
     /*** XXX: Everything here is immutable after initialization. ***/
-    /*** XXX: Calls through the interpreter and LLVM JIT runtime are thread-safe. ***/
     native_vector<std::filesystem::path> library_dirs;
 
     /* The files within this map will get added into Clang's VFS prior to the creation of
@@ -91,6 +92,7 @@ namespace jank::jit
     std::map<char const *, std::string_view> vfs;
 
     /*** XXX: Everything here is thread-safe. ***/
-    jtl::ptr<CppInternal::Interpreter> interpreter;
+    mutable folly::Synchronized<jtl::ptr<CppInternal::Interpreter>, std::recursive_mutex>
+      interpreter;
   };
 }
--- a/compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
+++ b/compiler+runtime/src/cpp/jank/analyze/cpp_util.cpp
@@ -31,7 +31,8 @@ namespace jank::analyze::cpp_util
    * After that failure, Clang gets back into a good state. */
   static void reset_sfinae_state()
   {
-    static_cast<void>(runtime::__rt_ctx->jit_prc.interpreter->Parse("1"));
+    auto locked_interpreter{ runtime::__rt_ctx->jit_prc.interpreter.lock() };
+    static_cast<void>((*locked_interpreter)->Parse("1"));
   }
 
   jtl::string_result<void> instantiate_if_needed(jtl::ptr<void> scope)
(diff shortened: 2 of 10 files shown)


compiler+runtime/include/cpp/jank/jit/processor.hpp
```diff
@@ -82,7 +83,6 @@ namespace jank::jit
              jtl::immutable_string const &lib);
 
     /*** XXX: Everything here is immutable after initialization. ***/
-    /*** XXX: Calls through the interpreter and LLVM JIT runtime are thread-safe. ***/
```
kylc, line 85: This doesn't seem true, from looking at clang's `Interpreter.h/cpp`. I don't see any synchronization in there.

kylc:
This PR fixes the code in the linked issue, but I've found another case which still fails on main and with this PR:

```
(defn a-deferred-func [])

(defn -main []
  (future (a-deferred-func))
  ;; fails without this sleep
  ;; (sleep 100)
  )

```

This code segfaults without the sleep. I believe this is because the `processor` dtor is deleting the `clang::Interpreter` because the process is exiting, and the deferred function is still compiling on a different thread. That might require some more thought to fix.

jeaye:
Thanks, Kyle. I've taken a look at this issue when that ticket was reported and I believe it's more involved than this. Based on my local reproductions of the issue, I've boiled it down to this:

1. I've been told that `clang::Interpreter` is thread-safe, but I think you're right to question this -- I have just followed up with the author of it to verify, since it's been a while. I'll let you know when I hear back.
2. I think the main issue here is the usage of CppInterOp, not the usage of `clang::Interpreter`. The crashes I've seen locally are generally within `Cpp::InstantiateTemplate` and similar fns. This is because we're using the `clang::Sema` and `clang::ASTContext` directly, which is 100% not thread-safe.

So I think that the first thing we want to fix is the CppInterOp lib usage, which is a large API surface area with currently no synchronization. We access it from various places in jank, but almost entirely from `analyze/processor.cpp` (the semantic analyzer) and `analyze/cpp_util.cpp` (semantic analysis helpers). We're able to change whatever we want in CppInterOp, since we're already so far diverged from upstream.

As side note, `folly::Synchronized` is still the way.

jeaye:
> This code segfaults without the sleep. I believe this is because the `processor` dtor is deleting the `clang::Interpreter` because the process is exiting, and the deferred function is still compiling on a different thread. That might require some more thought to fix.

This is an interesting case. jank doesn't currently have a `(shutdown-agents)` style fn which tracks the running threads, so having the main thread exit means everything gets torn down. I think that'll be a nice-to-have, but isn't crucial for the short term, since we can just have people keep the main thread alive.

jeaye:
Ok, I dug into this a little more, since you said that this PR fixes the issue. Obviously that means that I'm confused. :upside_down_face: 

Turns out I was conflating the issue you're fixing here with this one: https://github.com/jank-lang/jank/discussions/957

Both are thread safety issues. This one seems very clearly related to `clang::Interpreter` and the other one is very much related to CppInterOp. We don't need to try to solve both at the same time.

So yeah, let's get this guy merged, with `folly::Synchronized` usage. That'll solve one issue. Apologies for the confusion.

kylc:
Ok great :smiley: 

I've updated this PR to use `folly::Synchronized`. This ends up being noisy because:
1. There were a lot of unguarded uses of the interpreter sprinkled around. It's great that we are able to catch these by using the wrapped folly type.
2. I couldn't use `folly::SynchronizedPtr` because it doesn't seem to support `std` mutex types (it doesn't have `lock()` like `folly::Synchronized` does), so we end up with this ugly double-deref pattern `(*locked_interpreter)->do_thing()`.
    - I had to use `std::recursive_mutex` since there are so many uses of the interpreter that recursive locking is inevitable. The default `folly::SharedMutex` isn't recursive, and I don't think folly has any recursive mutex type.


I think that's all an OK tradeoff. I double checked and the linked issue is still fixed.

jeaye:
> Ok great 😃
> 
> I've updated this PR to use `folly::Synchronized`. This ends up being noisy because:
> 
>     1. There were a lot of unguarded uses of the interpreter sprinkled around. It's great that we are able to catch these by using the wrapped folly type.
> 
>     2. I couldn't use `folly::SynchronizedPtr` because it doesn't seem to support `std` mutex types (it doesn't have `lock()` like `folly::Synchronized` does), so we end up with this ugly double-deref pattern `(*locked_interpreter)->do_thing()`.
>        
>        * I had to use `std::recursive_mutex` since there are so many uses of the interpreter that recursive locking is inevitable. The default `folly::SharedMutex` isn't recursive, and I don't think folly has any recursive mutex type.
> 
> 
> I think that's all an OK tradeoff. I double checked and the linked issue is still fixed.

Yep, this all looks as I had imagined. I think we could clean this API up in the future and we may be able to get rid of locking around accessors, but this should get rid of the UB, which is most important.

Thank you!

PR #998, opened 2026-09-14
Various nREPL improvements

kylc:
- Fix argslist display in CIDER
- Add completion for special forms (cpp/new, etc.)
- Use all mapped ns vars for completion, not just the ones defined in that namespace
    - Now in your 'user namespace you will get completions from 'clojure.core, :refer-ed symbols, etc.
- Refactor to use type hints


jeaye (approved):
It's so nice to see the code get cleaned up as the language evolves.

